PR #1, opened 2026-04-22
[Draft] Add CI

atmnp:


leofang:
/ok to test [cb94b6e](https://github.com/NVIDIA/cuda-lang-numba-compat/pull/1/commits/cb94b6e7e5bcba9f01bca54e5305e98dd3b1203a)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,320 @@
...
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
+      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd
+        with:
+          extra_args: --all-files --show-diff-on-failure
+
+  compute-matrix:
```
leofang: Drive-by comment (still catching up): this step is probably not necessary. FWIW, I am killing (gradually) the "old" CI in numba-cuda, to make it align better with cuda-python: https://github.com/NVIDIA/numba-cuda/pull/872

.github/workflows/conda-python-tests.yaml
leofang, line 1: Q: Does the conda CI work now? If not, I suggest to just remove it for now.

.github/workflows/wheels-test.yaml
leofang, line 1: Ditto, as part of the housekeeping I will remove this file from numba-cuda.

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,336 @@
...
+      matrix:
+        python-version:
+          # TODO: Re-enable 3.10/3.11 after fixing PEP 695 type parameter syntax
+          # (def func[T](...)) to be compatible with Python < 3.12.
+          # - "3.10"
+          # - "3.11"
+          - "3.12"
+          - "3.13"
+          - "3.14"
+          # - "3.14t"
```
leofang: btw if we're blocked by numba-cuda's lack of freethreading support, let me know

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,336 @@
...
+    name: py${{ matrix.python-version }}
+    # Windows builds disabled:
+    # (inputs.host-platform == 'win-64' && 'windows-2022')
+    runs-on: ${{ (inputs.host-platform == 'linux-64' && 'linux-amd64-cpu8') ||
+                 (inputs.host-platform == 'linux-aarch64' && 'linux-arm64-cpu8') }}
+    container:
+      image: ${{ (inputs.host-platform == 'linux-aarch64' && 'quay.io/pypa/manylinux_2_28_aarch64') ||
+                 (inputs.host-platform == 'linux-64' && 'quay.io/pypa/manylinux_2_28_x86_64') || '' }}
```
leofang: cibuildwheel would also launch a manylinux container, so in other repos we just let cibuildwheel kick in. 

The (slight) complexity here is we need to download the LLVM builds. But I think they can be mounted from the host VM to the cibuildwheel-launched container ([example](https://github.com/NVIDIA/cuda-python/blob/3fc7c80adb6dad4a70081a2f943eacc7f2d42a62/.github/workflows/build-wheel.yml#L171-L173)).

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,336 @@
...
+        with:
+          name: llvm-modern-install-${{ env.LLVM_PY_TAG }}-${{ inputs.host-platform }}
+          path: llvm-modern-install
+
+      - name: Download LLVM 7 artifacts
+        if: ${{ startsWith(inputs.host-platform, 'linux') }}
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
+        with:
+          name: llvm7-install-${{ inputs.host-platform }}
+          path: llvm7-install
```
leofang, line 111: 👍

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,336 @@
...
+          ls -lahR dist/
+
+      - name: Upload cuda-lang-numba-compat build artifacts
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f  # v7.0.0
+        with:
+          name: ${{ env.CL_NUMBA_COMPAT_CUDA_ARTIFACT_NAME }}
+          path: dist/*.whl
+          if-no-files-found: error
+
+  build-tests:
```
leofang, line 167: Q: Do we have this step on Gitlab?! I thought this is unique to numba-cuda (that it has some test artifacts that requires nvcc to build).

atmnp: We do! It came in with the numba-cuda test suites, this step will likely be removed from the CI anyways after one of the open MRs gets merged.

leofang: OH! Good catch, thanks, I missed that...

leofang (approved):
Approve for now to jump start the development. We can follow up as needed later.

.github/CODEOWNERS
leofang, line 1: FYI this file has many errors, we should fix them in the immediate follow-up

atmnp: Yep, I had a feeling that one would require some discussion 😂

PR #3, opened 2026-04-30
Add CodeQL and Bandit code scanning workflows

leofang:
## Summary

Migrate the static analysis pipelines from cuda-python to keep code scanning infrastructure consistent across our repos:

- **CodeQL** (`codeql.yml`): Python (build-mode: `none`) + C/C++ (build-mode: `autobuild`) with `security-extended` queries.
- **Bandit** (`bandit.yml`): Ruff S-rules scan with SARIF upload to GitHub Security tab. Adapted from cuda-python to read ignore codes from `pyproject.toml` (instead of `ruff.toml`) and to handle the empty-ignores case gracefully.
- **Ruff config** (`pyproject.toml`): Enable flake8-bandit (`S`) rules via `extend-select`, with project-appropriate suppressions (S101 assert, S102 exec, S110 try-except-pass, S301 pickle, S603/S607 subprocess, S108/S307/S314 TODO for follow-up). Set `target-version = "py312"` so ruff can parse PEP 695 type parameter syntax used throughout the codebase.
- **Symlink fix** (`src/cusimt/cuda/libdevice.pyi`): Fix pre-existing broken symlink that pointed to `cusimt/cuda/libdevice.py` instead of `libdevice.py` (same directory). This was causing ruff to exit non-zero with a cache-key error.

Both workflows trigger on pushes to `pull-request/[0-9]+` and `main`, matching the existing `ci.yaml`.

## CI status

All scanning workflows pass:
- ✅ CodeQL Analyze (python)
- ✅ CodeQL Analyze (c-cpp)
- ✅ Bandit (analyze)
- ✅ pre-commit

## Notes

- Action versions and pin SHAs are kept in sync with cuda-python.
- The Bandit workflow dynamically reads S-prefixed ignore codes from `pyproject.toml` via `toml2json | jq`, so adding/removing suppressions in the ruff config is automatically reflected in CI without touching the workflow file.
- TODO items for follow-up: remove S108/S307/S314 suppressions after fixing the 10 underlying violations; update `requires-python` to `>=3.12`.

-- Leo's bot

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,8 @@ include-package-data = true
 
 [tool.ruff]
 line-length = 100
+# TODO: update requires-python to >=3.12 (code uses PEP 695 type params)
+target-version = "py312"
 
 [tool.ruff.format]
 docstring-code-format = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,9 +103,16 @@ ignore = [
     "B904",
     "B905",
     "B019",
+    # flake8-bandit (S) rules suppressed project-wide
+    "S101",   # assert usage (fundamental to test suite)
+    "S102",   # exec() used intentionally in compiler/codegen
+    "S110",   # try-except-pass for optional feature detection
+    "S301",   # pickle used for compiler serialization
+    "S603",   # subprocess calls are internal tooling
+    "S607",   # partial executable paths in subprocess (same as S603)
 ]
 select = ["B"]
-extend-select = ["PERF"]
+extend-select = ["PERF", "S"]
 fixable = ["ALL"]
 
 exclude = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,10 @@ ignore = [
     "S301",   # pickle used for compiler serialization
     "S603",   # subprocess calls are internal tooling
     "S607",   # partial executable paths in subprocess (same as S603)
+    # TODO: fix violations and remove these suppressions
+    "S108",   # insecure /tmp paths (4 hits in scripts/ci/)
+    "S307",   # eval() usage (2 hits in compiler)
+    "S314",   # xml.parse without defusedxml (4 hits in scripts/ci/)
 ]
 select = ["B"]
 extend-select = ["PERF", "S"]
--- a/src/cusimt/cuda/libdevice.pyi
+++ b/src/cusimt/cuda/libdevice.pyi
@@ -1,1 +1,1 @@
-cusimt/cuda/libdevice.py
+libdevice.py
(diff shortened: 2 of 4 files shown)


leofang:
/ok to test [e214791](https://github.com/NVIDIA/numba-cuda-mlir/pull/3/commits/e214791085946fd8ee83d03698e80217773fde03)

leofang:
/ok to test [2194079](https://github.com/NVIDIA/numba-cuda-mlir/pull/3/commits/2194079436cdac1281b36fe6c9594e3c409a3671)

leofang:
/ok to test [1331361](https://github.com/NVIDIA/numba-cuda-mlir/pull/3/commits/1331361806a751910ada8efbe3fecbaf92da674c)

leofang:
/ok to test [d2d5b21](https://github.com/NVIDIA/numba-cuda-mlir/pull/3/commits/d2d5b215d95daf721d77df09cbb817836ec7b0b8)

src/cusimt/cuda/libdevice.pyi
leofang, line 1: FYI @atmnp this seems like a bug coming from Gitlab. It chokes the code scanners.

leofang (commented):
OK, all code scanning are now blocked by GHAS being disabled org-wide (for private repos). I have reached out to the admins on Slack. Calling it a day.

cc @kkraus14 for vis

leofang:
/ok to test [d2d5b21](https://github.com/NVIDIA/numba-cuda-mlir/pull/3/commits/d2d5b215d95daf721d77df09cbb817836ec7b0b8)

leofang:
OK, Neil unlocked GHAS and the code scanning works now. PTAL.

I applied a few more rules than what we have in cuda-python. My focus was to get the scanning up and running, and follow up with the crafted set latter. If there is any immediate concern, let me know!

pyproject.toml
```diff
@@ -82,6 +82,8 @@ include-package-data = true
 
 [tool.ruff]
 line-length = 100
+# TODO: update requires-python to >=3.12 (code uses PEP 695 type params)
+target-version = "py312"
```
kkraus14, line 86: Does this comment indicate that we aren't compatible with someone using Python < 3.12 in some way?

leofang: My bot told me his finding when testing with ruff locally:
> ruff 0.15 now reports 16 invalid-syntax errors from PEP 695 type parameter lists (def func[T](...)), because ruff defaults to target-version = "py310" (derived from requires-python = ">=3.10").

So the answer is yes, we aren't compatible today. But we will be. I think once we settle the CI infra and move the team to GitHub, we can sort out the Python support matrix. The goal is to align with cuda-python and push the minimum down to 3.10.

leofang: (FWIW, the Gitlab CI is PY312-only.)

kkraus14 (approved):

PR #5, opened 2026-05-01
Enable extending linkage test

isVoid:
## Summary
- Add public `MLIRLower.link_external_item()` API and route internal linking through it
- Convert `TestExtendingLinkage.test_extension_adds_linkable_code` to the cuSIMT lowering callback signature and enable it

## Testing
- `LD_PRELOAD=/usr/local/cuda/lib64/libnvJitLink.so.13 .venv/bin/python -X faulthandler -m pytest -p no:rerunfailures -o addopts= tests/numba_cuda_tests/cudapy/test_extending.py::TestExtendingLinkage -vv`
- `LD_PRELOAD=/usr/local/cuda/lib64/libnvJitLink.so.13 .venv/bin/python -X faulthandler -m pytest -p no:rerunfailures -o addopts= tests/numba_cuda_tests/cudapy/test_extending.py -vv`

### DIFF
--- a/tests/numba_cuda_tests/cudapy/test_extending.py
+++ b/tests/numba_cuda_tests/cudapy/test_extending.py
@@ -7,6 +7,9 @@ import numpy as np
 import numba_cuda_mlir
 from numba_cuda_mlir import cuda
 
+from numba_cuda_mlir._mlir import ir as mlir_ir
+from numba_cuda_mlir._mlir.dialects import func
+from numba_cuda_mlir.lowering_utilities import convert, get_or_insert_function
 from numba_cuda_mlir.numba_cuda import types
 
 import inspect
@@ -274,10 +277,6 @@ class TestExtendingLinkage(NumbaCUDATestCase):
 
             @lower_builtin(external_add, types.uint32, types.uint32)
             def lower_external_add(builder, target, args, kwargs):
-                from numba_cuda_mlir._mlir import ir as mlir_ir
-                from numba_cuda_mlir._mlir.dialects import func
-                from numba_cuda_mlir.lowering_utilities import convert, get_or_insert_function
-
                 builder.link_external_item(code_object)
                 i32 = builder.get_mlir_type(types.uint32)
                 fnty = mlir_ir.FunctionType.get([i32, i32], [i32])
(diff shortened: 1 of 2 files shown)


gmarkall:
> `LD_PRELOAD=/usr/local/cuda/lib64/libnvJitLink.so.13`

Do you do this because you get an nvJitLinkError about incorrect versions if you don't do this? I always have to do `pip install nvidia-nvvm==13.2.* nvidia-nvjitlink==13.2.*` after doing `pip install -e .[cu13,dev]` for Numba-CUDA-MLIR, and I wonder if you get the same (I assumed I just had something weird about my local setup).

tests/numba_cuda_tests/cudapy/test_extending.py
```diff
@@ -275,12 +276,18 @@ def typer(x, y):
...
 
             @lower_builtin(external_add, types.uint32, types.uint32)
-            def lower_external_add(context, builder, sig, args):
-                context.active_code_library.add_linking_file(code_object)
-                i32 = ir.IntType(32)
-                fnty = ir.FunctionType(i32, [i32, i32])
-                fn = cgutils.get_or_insert_function(builder.module, fnty, "add_cabi")
-                return builder.call(fn, args)
+            def lower_external_add(builder, target, args, kwargs):
+                builder.link_external_item(code_object)
```
gmarkall, line 280: There's at least something counter-intuitive here: if `link_external_item` is a method of `MLIRLower`, then calling the `MLIRLower` a `builder` is misleading. The alternative (I'm just waiting for the tests) is that whatever `builder` is doesn't have this method.

isVoid: Should we call this `context` instead? I've always named this `builder` following previous conventions.

isVoid:
> > `LD_PRELOAD=/usr/local/cuda/lib64/libnvJitLink.so.13`
> 
> Do you do this because you get an nvJitLinkError about incorrect versions if you don't do this? I always have to do `pip install nvidia-nvvm==13.2.* nvidia-nvjitlink==13.2.*` after doing `pip install -e .[cu13,dev]` for Numba-CUDA-MLIR, and I wonder if you get the same (I assumed I just had something weird about my local setup).

Frankly it's agent's decision to patch the environment. One thing I noticed is that I haven't been using `[cu13]` optional dependency package.

isVoid:
On my github portal I see that the merge status of this PR cannot be shown. So I assume this PR is probably in bad state. Going to close and recreate a new PR.

PR #7, opened 2026-05-04
[CI] Enable Benchmarks to run on every pull request

atmnp:
The benchmarks CI has not been enabled for every PR, this enables it. This also removes the llm.py benchmark since it's not supported until the next release of CCCL.


leofang:
Q: How long does the benchmark test run? If it takes more than say 5 mins, we should be conservative because the runner resource is limited and shared across many teams/projects

atmnp:
5 minutes and 40 seconds including container setup/teardown, the actual test time is 4 minutes. We can also make it nightly/weekly, the CI workflow wasn't running at all before this, just wanted to test it before we decided on the frequency.

.github/workflows/ci.yaml
```diff
@@ -153,7 +151,7 @@ jobs:
 
   test-expensive-linux-64:
     name: Expensive Test linux-64
-    if: ${{ github.repository_owner == 'nvidia' && github.ref_name == 'main' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
+    if: ${{ github.repository_owner == 'nvidia' && (startsWith(github.ref_name, 'pull-request/') || github.ref_name == 'main') }}
```
leofang: I think this can be simplified?
```suggestion
    if: ${{ github.repository_owner == 'nvidia' }}
```

.github/workflows/ci.yaml
```diff
@@ -164,7 +162,7 @@ jobs:
 
   test-expensive-linux-aarch64:
     name: Expensive Test linux-aarch64
-    if: ${{ github.repository_owner == 'nvidia' && github.ref_name == 'main' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
+    if: ${{ github.repository_owner == 'nvidia' && (startsWith(github.ref_name, 'pull-request/') || github.ref_name == 'main') }}
```
leofang:
```suggestion
    if: ${{ github.repository_owner == 'nvidia' }}
```

leofang (approved):
5m40s is fine, modulo recent CI instability

PR #8, opened 2026-05-04
Add array debug info

jiel-nv:
This PR adds debug info support for Numba array type.

- Emits the 7-field Numba array descriptor: `meminfo`, `parent`, `nitems`, `itemsize`, `data`, `shape`, and `strides`.
- Supports both kernel-argument arrays and fixed-size `cuda.local.array` values.
- Adds FileCheck coverage for 1D kernel args, 2D kernel args, and local arrays.

Notes:
- Array kernel arguments already arrive as stable LLVM pointers, so they use direct `dbg.declare`.
- Local arrays arrive as `memref` values, the `_build_array_debug_descriptor` extracts the memref metadata and builds an LLVM struct matching the Numba descriptor layout. The synthesized descriptor folds memref element offsets into the `data` pointer, converts element strides to byte strides, computes `nitems`, and fills `meminfo` / `parent` with null pointers for local scratch storage.


tests/test_debuginfo.py
```diff
@@ -544,3 +544,118 @@ def k_record_local(records, out):
...
+        CHECK-SAME: elements = {{.*}}#[[ITEMSIZE]], #[[DATA]], #[[SHAPE]], #[[STRIDES]]
+        CHECK: #[[ARRAY_VAR:di_local_variable[0-9]*]] = #llvm.di_local_variable<{{.*}}name = "a"
+        CHECK-SAME: type = #[[ARRAY_TYPE]]
+        CHECK: llvm.intr.dbg.declare #[[ARRAY_VAR]] = %{{[0-9]+}} : !llvm.ptr
+        """,
+        mlir,
+    )
+
+
+def test_mlir_array_2d_kernel_arg_type():
```
VijayKandiah: Could we add a test for Fortran-order ([::1, :])? The stride computation in `_build_array_debug_descriptor` reads strides from the memref, so it should handle column-major correctly and it would be good to verify.

src/numba_cuda_mlir/mlir_debuginfo.py
```diff
@@ -275,6 +275,67 @@ def _record_di_type(numba_type):
...
+
+    members = [
+        member("meminfo", i8_ptr_di, meminfo_offset, _POINTER_BITS),
+        member("parent", i8_ptr_di, parent_offset, _POINTER_BITS),
+        member("nitems", int64_di, nitems_offset, _INT_LITERAL_BITS),
+        member("itemsize", int64_di, itemsize_offset, _INT_LITERAL_BITS),
+        member("data", elem_ptr_di, data_offset, _POINTER_BITS),
+        member("shape", tuple_di, shape_offset, ndim * _INT_LITERAL_BITS),
+        member("strides", tuple_di, strides_offset, ndim * _INT_LITERAL_BITS),
+    ]
```
VijayKandiah: The 7-field descriptor layout is now hardcoded in `_array_di_type`, `_build_array_debug_descriptor` here. Could we derive the field names/order from `ArrayModel` in `numba_cuda/datamodel/models.py` instead? That way, if something were to change, we don't silently drift away here.

VijayKandiah (changes requested):

VijayKandiah (approved):
LGTM now, thank you!

PR #10, opened 2026-05-04
Expose variables_used linker option

isVoid:
Currently in Numba-CUDA's linker option, `optimize_unused_variables` option is set to default (False). So in nvshmem workflow, the required global variables are preserved after compilation. However, in numba-cuda-mlir, this option is set to true. Therefore we need to manually set them to make sure those global variables are preserved. We expose the setting function via `variables_used` attribute.


atmnp (approved):
LGTM, if you wanted to get this merged in to get it out of the backlog.

PR #21, opened 2026-05-06
Fix ExternFunction capture in overloads and linker cuda.core compat

atmnp:
1.) Register `typeof_impl` for `ExternFunction` so that captured extern function
references (from `cuda.declare_device`) can be typed during closure
analysis. Add a dummy `lower_builtin` for `ExternFunction` calls so numba's
legacy LLVM trial-compilation in `_build_impl` can lower overload
implementations that call extern functions without hitting a
`NotImplementedError`. Together these let `ExternFunction` work as a captured
value inside `@overload` and `@overload_method` implementations. (Fixes #117)

2.) Override `_get_jit_decorator` in a new `_NumbaCudaMlirOverloadFunctionTemplate`
subclass so that the dispatcher created during overload probing uses the
MLIR compilation pipeline. Without this, `get_call_type` triggers
`CUDADispatcher.compile_device` through the default numba-cuda pipeline,
which looks up argument data models in the legacy LLVM data model manager
and crashes with a `KeyError` for types that are only registered with
`numba_cuda_mlir`'s MLIR model registry. The `make_overload_template` helper
gains a `base=` parameter to support this injection. (Fixes #118)

These are bundled since 2.) also partially fixes the bug causing 1.)

### DIFF
--- a/src/numba_cuda_mlir/compiler.py
+++ b/src/numba_cuda_mlir/compiler.py
@@ -11,6 +11,8 @@ from numba_cuda_mlir.numba_cuda.typing.templates import (
     AttributeTemplate,
     Registry,
 )
+from numba_cuda_mlir.numba_cuda.typing.typeof import typeof_impl
+from numba_cuda_mlir.numba_cuda.core.imputils import lower_builtin
 from pathlib import Path
 from numba_cuda_mlir.mlir_optimization import optimize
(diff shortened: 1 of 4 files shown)


VijayKandiah (approved):

ZzEeKkAa:
Works locally for me, thank you!

ZzEeKkAa (approved):

PR #22, opened 2026-05-06
Add discriminated-union debug info for polymorphic scalar variables

jiel-nv:
- Emit DWARF discriminated-union metadata for supported polymorphic scalar locals using `DW_TAG_variant_part` and per-variant `extraData` discriminator values.
- Allocate one canonical LLVM stack slot per polymorphic source variable and route debug declaration, stores, and loads through that storage.
- Adds tests: 
  - test_mlir_poly_scalar_var with validation for the emitted discriminated-union DI
  - test_mlir_poly_scalar_var_runtime_debug with validation for runtime output value when `debug=True, opt=False`, so the canonical debug slot load/store path is covered.

### DIFF
--- a/src/numba_cuda_mlir/mlir_lowering.py
+++ b/src/numba_cuda_mlir/mlir_lowering.py
@@ -96,6 +96,8 @@ KERNEL_ERROR_CODES = {
     RuntimeError: 4,
     ZeroDivisionError: 5,
 }
+# MLIR LLVM dialect dynamic GEP sentinel, LLVM::GEPOp::kDynamicIndex / INT32_MIN.
+_GEP_DYNAMIC_INDEX = -2147483648
 
 
 def _is_omitted_arg(argty):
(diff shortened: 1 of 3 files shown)


src/numba_cuda_mlir/mlir_lowering.py
```diff
@@ -2904,11 +2984,61 @@ def _load_var(self, var: numba_ir.Var) -> Any:
...
     def store_var(self, var, value):
         """
         Store the value (MLIR Op) into the given variable.
         """
         trace("var=%s value=%s", var, value)
+        if self._debug_full:
+            base_name = self._canonical_dbg_var_name(var.name)
+            if self._poly_dbg_alloca.get(base_name) is not None:
+                self._store_poly_dbg_var(var.name, value)
+                return
```
VijayKandiah, line 3041: The `store_var` path for polymorphic debug variables returns early after writing to the debug slot, bypassing the normal varmap/stack storage. This means debug path use a completely different load/store path than default non-debug path, so a bug in the poly slot GEP/type logic would silently produce wrong results only under debug=True. Can we add a runtime test (with debug=True, assert output correctness per branch) to cover this?

VijayKandiah (changes requested):

VijayKandiah (approved):
LGTM now, thanks!

PR #26, opened 2026-05-06
Add mypy/pyright support

ZzEeKkAa:
Add support for mypy/pyright syntax highlight. Tested only with editable install, but wheel should also work.

Screenshot 2026-05-06 at 7 27 09 PM


atmnp (approved):
LGTM on green CI

PR #27, opened 2026-05-07
[Compile-Time] Re-design the extending API + dispatcher changes

atmnp:
These were all changes necessary to achieve the final benchmark results in the updated README.

- Kernels now default to LTOIR output when nvJitLink is available (matches numba-cuda)
- Regular PTX and post-LTO PTX are materialized lazily only when inspection is requested
- Bug Fix: MLIR cache identity now includes LTO/output mode and the resolved GPU target
- Default GPU target resolution moved to compile/cache time
- Vector constructor typing/lowering no longer relies on combinatorial Numba-CUDA-compatible registrations, and the ones added strictly for numba-cuda compatibility are now removed.
- Repeated kernel launches are faster: dispatch uses cached Python type signatures, a validated array fast path, and C++ scalar fast extractors for common scalar-only launches.
- Benchmark scripts support `--compile-mode {cold,warm}` and `--backend` and reports the geomean

# Major Change: Extending API Changes
- `numba_cuda_mlir.extending` APIs now require explicit registries:
  - `overload(..., typing_registry=...)`
  - `overload_method(..., typing_registry=...)`
  - `overload_attribute(..., typing_registry=..., lowering_registry=...)`
  - `register_jitable(..., typing_registry=...)`

Before: extensions would register into shared global state.  
After: extension authors must choose the registry they are extending, which makes registration order and ownership explicit.

Co-authored-by: Vijay Kandiah <vkandiah@nvidia.com>


VijayKandiah:
This PR includes commit ca14a33ca426d29ca5ea546a09d799b67d2a363f which contains changes to optimize kernel launch latency and introduces a microbenchmark to measure and compare numba-cuda-mlir's kernel launch latency with numba-cuda's. These changes reduce numba-cuda-mlir's kernel launch latency to be 1.0–17x faster than numba-cuda across all tested configurations (scalar and array arguments).

```
--------------------------------------------------------------------------------------
Benchmark                |  numba_cuda (ns) |  numba_cuda_mlir (ns) |    Speedup
--------------------------------------------------------------------------------------
launch_empty             |           4101.9 |                3871.2 |      1.06x
launch_1_array_arg       |           5839.0 |                2962.3 |      1.97x
launch_16_scalar_args    |          15204.4 |                4306.9 |      3.53x
launch_16_array_args     |          29780.4 |               12370.7 |      2.41x
launch_256_scalar_args   |         177007.1 |               10108.6 |     17.51x
--------------------------------------------------------------------------------------
```

src/numba_cuda_mlir/extending/__init__.py
```diff
@@ -59,8 +71,11 @@ def overload(
     strict=True,
     inline="never",
     prefer_literal=False,
+    typing_registry=None,
+    lowering_registry=None,
```
VijayKandiah: `overload()` accepts `lowering_registry=` but silently discards it — should it be removed, or is it kept intentionally for call-site/ API surface uniformity with `overload_attribute` which does accept and use `lowering_registry=`?

VijayKandiah (approved):
LGTM after rebase and CI passing.

PR #13, opened 2026-05-05
Add User guide

gmarkall:
This adds docs based on the existing Numba-CUDA documentation. It includes:

- Copying the documentation from Numba-CUDA.
- Editing the User Guide for relevance to Numba-CUDA.
- Copying in the doctests, but setting them to be ignored for now. They need to be present for the documentation to build, but I would like to fix them up in a subsequent PR as this one already contains a lot.

The actual edits are in the commit https://github.com/NVIDIA/numba-cuda-mlir/pull/13/changes/907dd5bb522fd0ec83e75685721493fb398b935f. The changes are broadly:

- Remove a few sections that don't make sense to keep:
  - Bindings notes: we have only used cuda.core bindings for a long time now, so the bindings notes are no longer relevant.
  - Caching notes: I think we want to re-think the way caching works, so I believe it's best left undocumented.
  - CUDA Array Interface: This is a standard, and there's no need to duplicate its docs here.
  - EMM Plugin interface: The relevance of the interface disappears if we recommend people use other libraries for allocating and managing arrays.
  - The FAQ: It only contained irrelevant notes about the old profiler.
  - Simulator docs: There is no simulator.
  - MVC notes: These are no longer relevant with nvjitlink in use at all times
- Update references to the name (Numba CUDA MLIR) and remove references to Numba.
- Update links to the CUDA Programming Guide.
- Update package names to use `numba_cuda_mlir.cuda`.
- Add some introduction explaining what Numba CUDA MLIR is.

Future work still includes:

- Updating the reference documentation. This would ideally also be complete now, but I think it's more important to have a consistent user guide now.
- Update the doctests.
- Set up the doc build as described by Leo.

PR #28, opened 2026-05-07
Add docs build

gmarkall:
Adds a multi-version build of the docs as outlined in https://github.com/NVIDIA/numba-cuda-mlir/pull/13#issuecomment-4390192540

- Addition of the versions.json file for the multi-version dropdown.
- Adds the build_docs.sh file to build the docs and then copy the built version to the correct subfolder, plus other files (generating the index.html for the root, and copying objects.inv).
- Adds a Github Actions workflow to build and deploy the docs, based on the one from CUDA Python with modifications for the numba-cuda-mlir setup.
- Remove the original docs-build.yaml copied from Numba-CUDA.


gmarkall:
I've downloaded the artifact from the build-docs job, and I think it looks as expected:

- `index.html` redirects to the `latest` subfolder
- `latest` subfolder contains the built docs
- The root dir also contains `objects.inv` (for intersphinx?) and `versions.json` (for the version drop-down)

I believe this is ready for review now.

docs/build_docs.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
```
leofang: Two things in this header:

1. The license should be `Apache-2.0` to match the rest of the repo. `LicenseRef-NVIDIA-SOFTWARE-LICENSE` was carried over from cuda-python's `cuda_core` (which is under a different license), but this repo is Apache-2.0.
2. Copyright year — this is a new file for this project, so should probably be `2026` rather than `2024-2025`.

docs/source/conf.py
```diff
@@ -40,3 +42,24 @@
...
+    "navbar_center": [
+        "version-switcher",
+        "navbar-nav",
+    ],
+}
+
+# Add warning to docs built from main
+if int(os.environ.get("BUILD_LATEST", 0)):
+    html_theme_options["announcement"] = (
+        "Warning: This documentation is built from the development branch\!"
```
leofang: Nit: the `\!` is not a recognized Python escape sequence, so the backslash will appear literally in the rendered HTML (`...branch\!`). Should just be `"!"`:
```python
        "Warning: This documentation is built from the development branch!"
```

leofang (commented):
Overall this looks great — closely follows the cuda-python pattern and the multi-version setup is all wired up correctly. Just a couple of minor fixups in the inline comments.

-- Leo's bot

docs/build_docs.sh
leofang: If this file is a copy-pasta, keeping the start year in 2024 seems fine. But we should update the end year to 2026. 

The license identifier is a good catch.

docs/source/conf.py
gmarkall: Thanks!

```suggestion
        "Warning: This documentation is built from the development branch!"
```

docs/build_docs.sh
gmarkall: I think it's part original and part copy-pasted, so I'll update the year and license ref.

gmarkall:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0
```

leofang (approved):

PR #32, opened 2026-05-11
Add float16 vector support

ZzEeKkAa:
Add float16 vector support. Minimal changes. Almost everything was already there

What I did not check - if the generated instructions are performant:
https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH____HALF2__ARITHMETIC.html
However, I guess that should've been handled by mlir cuda dialect


atmnp (approved):
LGTM

PR #33, opened 2026-05-11
Add vector view support

ZzEeKkAa:
Add vector view support. Most of the support was done in #35 and #36 . This one adds manipulations on the array length when dtype size does not match. Plus add robust vector view tests coverage


ZzEeKkAa:
It's working but I'm not particularly happy with `VectorTypeStub`/`VectorType` casts. `cuda.float32x2` or whatever should be registered as `VectorType` by default, not as `VectorTypeStub`. It adds complexity later for `.view()`, `cuda.local.array()` etc, as we need to cast it everywhere.

ZzEeKkAa:
> It's working but I'm not particularly happy with `VectorTypeStub`/`VectorType` casts. `cuda.float32x2` or whatever should be registered as `VectorType` by default, not as `VectorTypeStub`. It adds complexity later for `.view()`, `cuda.local.array()` etc, as we need to cast it everywhere.

Targeted at #36

src/numba_cuda_mlir/lowering_utilities/__init__.py
```diff
@@ -151,6 +151,12 @@ def get_type_width(ty: ir.Type) -> int:
             return ty.width
         case ir.ComplexType():
             return ir.ComplexType(ty).element_type.width
```
ZzEeKkAa, line 153: Looks like complex also need this fix. But let's target it in a separate MR. Credits to @Jlisowskyy for finding this bug in #34

src/numba_cuda_mlir/lowering_utilities/__init__.py
ZzEeKkAa, line 145: But I guess this should not be done manually. Instead we have to pull up datamodel manager, find corresponding model and get bitsize for it

src/numba_cuda_mlir/lowering/numpy.py
```diff
@@ -965,7 +965,39 @@ def lower_array_view_cg(builder, target, args, kwargs):
...
+        offset = metadata[1]
+
+        target_rank = memref_type.rank
+        dynamic_sizes = list(metadata[2 : 2 + target_rank])
+        dynamic_strides = list(metadata[2 + target_rank :])
+
+        source_width_val = arith.constant(source_width, index=True)
+        target_width_val = arith.constant(target_width, index=True)
+        bytes_size = arith_dialect.muli(dynamic_sizes[-1], source_width_val)
+        dynamic_sizes[-1] = arith_dialect.divui(bytes_size, target_width_val)
```
atmnp, line 985: Missing divisibility check?

src/numba_cuda_mlir/lowering_utilities/__init__.py
atmnp: Yep!

src/numba_cuda_mlir/lowering/numpy.py
atmnp, line 956: it looks to me like the higher dimensions keep the wrong strides?

ZzEeKkAa:
Looks like suppressed by #35 . However I want to double check tests

atmnp (changes requested):
Sorry I thought I submitted this a while ago, better late than never.

ZzEeKkAa:
Suppressed by #35

PR #34, opened 2026-05-11
Fix array.view dtype resolution, typing, and lowering

VijayKandiah:
### Summary
- Route `array.view()` dtype resolution through the `to_mlir_type` singledispatch instead of the bare `np_dtype_to_mlir_type` handler, making externally-registered type converters reachable
- Short-circuit same-element-type views as a no-op to avoid emitting an identity `tensor.bitcast` that downstream MLIR pass `one-shot-bufferize` cannot handle
- Fix `NumpyArrayViewMethodTemplate` to return `Array(new_dtype, ndim, layout)` instead of the receiver's type, so the view carries the correct element type through typing and into lowering
- Replace the `tensor.bitcast` lowering path with a `unrealized_conversion_cast` between memref types, which resolves cleanly through the MLIR pass pipeline that includes `one-shot-bufferize`
- Add 3 new tests:
  - `test_array_view_same_dtype` — `int64` viewed as `int64`
  - `test_array_view_reinterpret_dtype` — `float32` viewed as `int32`
  - `test_array_view_custom_dtype` — custom boxed dtype via `@to_mlir_type.register`

### DIFF
--- a/src/numba_cuda_mlir/typing/numpy.py
+++ b/src/numba_cuda_mlir/typing/numpy.py
@@ -990,14 +990,7 @@ class NumpyArrayViewMethodTemplate(AbstractTemplate):
         if len(args) == 1:
             arr = self.this
             dtype = args[0]
-
-            if isinstance(dtype, types.DTypeSpec):
-                element_type = dtype.dtype
-            elif isinstance(dtype, types.NumberClass):
-                element_type = dtype.dtype
-            else:
-                element_type = dtype
-
+            element_type = dtype.dtype if isinstance(dtype, types.DTypeSpec) else dtype
             ret = types.Array(element_type, arr.ndim, arr.layout)
             return signature(ret, dtype, recvr=self.this)
         return None
--- a/src/numba_cuda_mlir/typing/numpy.py
+++ b/src/numba_cuda_mlir/typing/numpy.py
@@ -990,8 +990,16 @@ class NumpyArrayViewMethodTemplate(AbstractTemplate):
         if len(args) == 1:
             arr = self.this
             dtype = args[0]
-            # Return array with potentially different dtype
-            return signature(arr, dtype, recvr=self.this)
+
+            if isinstance(dtype, types.DTypeSpec):
+                element_type = dtype.dtype
+            elif isinstance(dtype, types.NumberClass):
+                element_type = dtype.dtype
+            else:
+                element_type = dtype
+
+            ret = types.Array(element_type, arr.ndim, arr.layout)
+            return signature(ret, dtype, recvr=self.this)
         return None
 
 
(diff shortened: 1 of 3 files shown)


src/numba_cuda_mlir/typing/numpy.py
```diff
@@ -990,8 +990,9 @@ def generic(self, args, kws):
         if len(args) == 1:
             arr = self.this
             dtype = args[0]
-            # Return array with potentially different dtype
-            return signature(arr, dtype, recvr=self.this)
+            element_type = dtype.dtype if isinstance(dtype, types.DTypeSpec) else dtype
```
ZzEeKkAa, line 993:
```suggestion
            parsed_dtype = parse_dtype(dtype)
            retty = arr.copy(dtype=parsed_dtype)
            # Return array with potentially different dtype
            return signature(retty, dtype, recvr=self.this)
```

With this import on top

```python
from numba_cuda_mlir.numba_cuda.typing.npydecl import parse_dtype
```

In this way we will be actually aligned with others dtype behavior, like at local/shared array

ZzEeKkAa: Here is the reference implementation of `parse_dtype`. It also handles string (like `dtype="i32"`) and `TypeRef` (Used when a type is passed as a value.)


```python
def parse_dtype(dtype):
    """
    Return the dtype of a type, if it is either a DtypeSpec (used for most
    dtypes) or a TypeRef (used for record types).
    """
    if isinstance(dtype, types.DTypeSpec):
        return dtype.dtype
    elif isinstance(dtype, types.TypeRef):
        return dtype.instance_type
    elif isinstance(dtype, types.StringLiteral):
        dtstr = dtype.literal_value
        try:
            dt = np.dtype(dtstr)
        except TypeError:
            msg = f"Invalid NumPy dtype specified: '{dtstr}'"
            raise TypingError(msg)
        return from_dtype(dt)
```

Jlisowskyy (approved):
Looks good! It definitely solves all three issues I reported, but two major problems are left unresolved. First, it will fail for bitwidth changes (e.g., a view from complex64 to float32). Second, it will fail when view() is performed on a shared memory array, which will result in a runtime error. Both of these issues are addressed in #35, but those changes are more complex. I'm fine with merging this now and working on the remaining issues afterward

VijayKandiah:
> Looks good! It definitely solves all three issues I reported, but two major problems are left unresolved. First, it will fail for bitwidth changes (e.g., a view from complex64 to float32). Second, it will fail when view() is performed on a shared memory array, which will result in a runtime error. Both of these issues are addressed in #35, but those changes are more complex. I'm fine with merging this now and working on the remaining issues afterward

Sounds good, let's get this in first.

PR #35, opened 2026-05-11
Full .view() implementation including dtype bitwidth change

Jlisowskyy:
This PR completes ndarray.view(dtype) lowering on the MLIR backend, including dtype changes that alter element bitwidth. Motivation: while enabling cuFFTDx in nvmath-python I produced this implementation - it passes the new test suite added here and the nvmath-python tests run cleanly on top of it.

AI summary of changes:
Code changes

- src/numba_cuda_mlir/lowering/numpy.py:959 — Rewrote lower_array_view_cg. Now handles three cases:
  - Same dtype → no-op store.
  - Same bitwidth, different dtype → unrealized_conversion_cast (pure retype).
  - Different bitwidth → extract_strided_metadata + recomputed offset/sizes/strides through reinterpret_cast on the appropriate axis, with new helper _view_transform_axis.
- src/numba_cuda_mlir/typing/numpy.py:990 — NumpyArrayViewMethodTemplate now returns a properly typed types.Array whose dtype reflects the target element type (unwraps DTypeSpec / NumberClass), instead of returning the original array type.
- src/numba_cuda_mlir/lowering_utilities/__init__.py:150 — get_type_width fixes: complex is now 2 * element width (was element width only) and adds a VectorType case (product of shape × element width). Promotion rule for complex uses scalar-only widths to avoid double-counting.
- tests/test_array_methods.py — removed @pytest.mark.xfail on the prior view test (renamed test_array_view).
- tests/test_array_view.py (new, 295 lines) — full parametrized matrix.

What is supported

- Same-dtype .view() (identity).
- Same-width reinterpret of any rank (e.g. i32↔f32, u16↔i16, f64↔i64, including 3-D).
- Different-width .view() for 1-D arrays (narrowing and widening: i64→i8, i8→i32, f64→f32, …).
- Different-width .view() for "C" and "F" contiguous N-D arrays — adjusts the trailing axis for C, leading axis for F.
- "A" layout, rank ≥ 2 — runtime dispatch: picks the first axis with stride == 1; if none, sets the kernel ValueError error code.
- .view() after slicing (1-D and 2-D trailing-contiguous slice).
- Chained .view() calls (view-of-view).
- Half-precision (u16↔f16, f16 widening to i32) and special floats (±inf, nan, signed zeros) preserved bit-exactly.

Not covered / explicit failure paths

- 0-D arrays — raised at lowering (ValueError).
- "A" layout rank ≥ 2 with no unit-stride axis — runtime error code set, no transformation applied.


Jlisowskyy:
This approach could serve as inspiration for #34. If this code looks promising, I would be happy to work on it, or someone can take over from here

src/numba_cuda_mlir/lowering_utilities/__init__.py
```diff
@@ -150,7 +150,13 @@ def get_type_width(ty: ir.Type) -> int:
         case ir.FloatType():
             return ty.width
         case ir.ComplexType():
-            return ir.ComplexType(ty).element_type.width
+            return 2 * ir.ComplexType(ty).element_type.width
```
ZzEeKkAa, line 153: Nice. Should we add test somewhere to catch this. I feel like fix for the `get_type_width` (along with  `case ir.VectorType():`) deserve separate MR

src/numba_cuda_mlir/typing/numpy.py
```diff
@@ -990,8 +990,12 @@ def generic(self, args, kws):
         if len(args) == 1:
             arr = self.this
             dtype = args[0]
-            # Return array with potentially different dtype
-            return signature(arr, dtype, recvr=self.this)
+            element_type = (
+                dtype.dtype if isinstance(dtype, (types.DTypeSpec, types.NumberClass)) else dtype
```
ZzEeKkAa: `NumberClass` inherits from `DTypeSpec`, so we don't need to check both. Let's merge #34 first

tests/test_array_methods.py
```diff
@@ -20,8 +20,7 @@ def slicing(byte_arr: cuda.DeviceNDArray, start: int, stop: int, output: cuda.De
     assert output[0] == expect, f"output: {output} != {expect}"
 
 
-@pytest.mark.xfail
-def tests_array_view():
+def test_array_view():
```
ZzEeKkAa, line 23: I feel like this should be also part of #34 . I'm wondering how did the tests pass there, if it was marked `xfail`?

UPD: Ho, okay. It passed because prefix was `tests`, not `test`

atmnp:
#38 should help triage the internal compiler failures more easily in the future, it was added to the expensive benchmarks but forgotten in the regular test suite.

Jlisowskyy:
Apparently `test_reinterpret_array_type` is working now after applying the changes

atmnp (approved):

PR #36, opened 2026-05-12
Refactor: Replace VectorTypeStub with unified VectorType and VectorTypeClass

ZzEeKkAa:
### Overview
This MR simplifies the CUDA vector type architecture by eliminating the dynamically generated `VectorTypeStub` classes. 

Previously, the codebase maintained separate stub classes to act as user-facing constructors (e.g., `cuda.float32x2`), which added unnecessary indirection and complexity when resolving types for operations like `np.array.view()`. We now use `VectorType` instances directly as the user-facing objects, backed by a newly introduced `VectorTypeClass` in the Numba type system. This allows vector types to seamlessly act as both multi-argument constructors and standard `dtype` specifiers.

### Key Changes

* **Removed `VectorTypeStub`**: Completely removed the dynamically generated stub classes from `src/numba_cuda_mlir/cuda/vector_types.py`.
* **Direct `VectorType` Exports**: The `cuda` module now directly exposes instances of `VectorType` (e.g., `cuda.float32x2` is now a `VectorType` instance rather than a stub class).
* **Introduced `VectorTypeClass`**: Added `VectorTypeClass` in `typing/cuda_vector_types.py` to represent `VectorType` instances within Numba's type system. It is properly registered with Numba's data model manager (`OpaqueModel`).
* **Unified Type Resolution**: 
  * Updated `typeof_impl` to correctly map `VectorType` instances to `VectorTypeClass`.
  * Refactored constructor templates to build directly from `VectorType` instances.
  * Updated `typing/cuda.py` and `typing/numpy.py` to resolve attributes against the new `vector_types_by_name` and `vector_types_by_alias` dictionaries.
* **Lowering Updates**: Updated `lowering/cuda_vector_types.py` to register constructor lowering directly against the `_vector_types` list.
* **Test Suite Alignment**: Updated internal test references in `tests/numba_cuda_tests/cudapy/test_vector_type.py` to iterate over `_vector_types` instead of `_vector_type_stubs`.

### Impact
* **Cleaner Architecture**: Reduces the number of internal representations for vector types from three (Stub, internal Type, Numba Type) down to two (`VectorType` instance, `VectorTypeClass`).
* **Better NumPy Integration**: Makes it much more straightforward to use vector types as `dtype` arguments for operations like `array.view()`, aligning their behavior with standard types like `np.float32` or `np.complex64`.
* **Stability**: All existing unit tests and the full `numba_cuda_tests` suite pass with the new architecture.


src/numba_cuda_mlir/cuda/vector_types.py
```diff
@@ -93,32 +78,32 @@ def map_vector_type_stubs_to_alias(vector_type_stubs):
...
-# Export all stubs as module-level attributes
-for stub in _vector_type_stubs:
-    globals()[stub.__name__] = stub
-    for alias in stub.aliases:
-        globals()[alias] = stub
+vector_types_by_name = {vec_type.__name__: vec_type for vec_type in _vector_types}
+vector_types_by_alias = {}
+for vec_type in _vector_types:
+    for alias in vec_type.aliases:
+        vector_types_by_alias[alias] = vec_type
```
Jlisowskyy:
```suggestion
vector_types_by_alias = {alias: vec_type for vec_type in _vector_types for alias in vec_type.aliases}
```

src/numba_cuda_mlir/cuda/vector_types.py
```diff
@@ -45,37 +46,21 @@ def make_vector_type_stubs():
...
-
-    return vector_type_stubs
-
-
-def map_vector_type_stubs_to_alias(vector_type_stubs):
-    """Create C-compatible aliases for vector type stubs (e.g., float4 -> float32x4)."""
+        base_type = BASE_TYPE_MAP[prefix]
+        vec_type = VectorType(base_type, nelem)
+        vec_type._base_type_name = prefix  # Keep for alias mapping
+        vec_type._num_elements = nelem
```
Jlisowskyy: I belive base class already exposes such property by `length`

src/numba_cuda_mlir/cuda/vector_types.py
```diff
@@ -45,37 +46,21 @@ def make_vector_type_stubs():
...
+        vec_type.__name__ = f"{prefix}x{nelem}"
+        vec_type.aliases = []
+        vector_types.append(vec_type)
```
Jlisowskyy: We are losing an __init__() exception here, it may be confusing for the user

Jlisowskyy: Maybe we could create a subclass of `VectorType`. It would be a little more readable and robust

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -13,50 +13,62 @@
...
+
+    @property
+    def dtype(self):
+        return self.instance_type
+
+    def get_call_type(self, context, args, kws):
+        return self._template(context).apply(args, kws)
+
+    def get_call_signatures(self):
+        return (), True
```
Jlisowskyy: I think this tuple should contain allowed signatures:
```
    @abstractmethod
    def get_call_signatures(self):
        """
        Returns a tuple of (list of signatures, parameterized)
        """
```

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -13,50 +13,62 @@
...
-    "uint8": types.uint8,
-    "uint16": types.uint16,
-    "uint32": types.uint32,
-    "uint64": types.uint64,
-    "float16": types.float16,
-    "float32": types.float32,
-    "float64": types.float64,
-}
+    def get_impl_key(self):
+        return self.typing_key
```
Jlisowskyy: I think there is signature mismatch:
```
    @abstractmethod
    def get_impl_key(self, sig):
        """
        Returns the impl key for the given signature
        """
```

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -131,19 +143,11 @@ def resolve_w(self, ty):
...
+# Register typeof_impl for all vector types so they can be used as closure variables
 
 
-for stub in _vector_type_stubs:
-    typeof.typeof_impl.register(stub)(_make_typeof_impl(stub))
+@typeof.typeof_impl.register(VectorType)
+def typeof_vector_type(val, c):
+    """Create a typeof implementation for a vector type."""
+    template = make_constructor_template(val)
+    return VectorTypeClass(val, template)
```
Jlisowskyy, line 155: Hmm, maybe we could cache the `VectorTypeClass` instead?

src/numba_cuda_mlir/cuda/vector_types.py
ZzEeKkAa: What do you mean? I guess we should just move this to VectorType.__init__?

src/numba_cuda_mlir/typing/cuda_vector_types.py
ZzEeKkAa: What do you mean?

VectorTypeClass is the type of type(VectorType instance)

src/numba_cuda_mlir/cuda/vector_types.py
Jlisowskyy: Oops, I combined two thoughts here. We could definitely do that, but it would require some refactoring of existing usages?. As for the exception, I just wanted to point out that we lose the stub's exception here

ZzEeKkAa: Please check how it looks now. I guess it should resolve this thread

src/numba_cuda_mlir/typing/cuda_vector_types.py
ZzEeKkAa: TBH, I have never touched it. Just asked AI to fill it properly. Please check if that something you would expect. It works as expected, so we can leave that for later refactoring

atmnp (approved):
LGTM, looks sane. @Jlisowskyy is this good to go?

src/numba_cuda_mlir/cuda/vector_types.py
Jlisowskyy: Looks great, thanks!

Jlisowskyy (approved):
LGTM

src/numba_cuda_mlir/typing/cuda_vector_types.py
Jlisowskyy: I was just wondering, is it sensible to cache the whole `VectorTypeClass`? Just looking to learn something here

ZzEeKkAa: I don't see the reason. Numba typing mechanism should be smart enough to avoid duplicates

PR #37, opened 2026-05-12
Add vector support for local and shared memory

ZzEeKkAa:
# Support VectorType as dtype for `cuda.local.array` and `cuda.shared.array`

## Summary
This PR adds support for using CUDA vector types (e.g., `cuda.float32x4`) as the `dtype` argument when allocating `cuda.local.array` and `cuda.shared.array`. 

Previously, the lowering decorators for local and shared arrays strictly expected `types.NumberClass` (or string literals) for the `dtype` argument, which caused compilation failures when passing vector types (which use `VectorTypeClass` inheriting from `types.DTypeSpec`).

## Changes
- **Broadened `dtype` support:** Updated the `@lower` decorators for `cuda.local_array` and `cuda.shared_memory` in `src/numba_cuda_mlir/lowering/cuda.py` to accept `types.DTypeSpec` instead of `types.NumberClass`. This allows any valid dtype specifier to be used, including `VectorTypeClass`.
- **Vector element assignment:** Registered `VectorType` in `lower_array_setitem` within `src/numba_cuda_mlir/lowering/numpy.py` so that vector elements can be correctly assigned to arrays (e.g., `arr[0] = cuda.float32x4(...)`).
- **Cleanup:** Removed duplicate lowering decorators for `cuda_local_array`.
- **Testing:** Added `test_vector_local_array` and `test_vector_shared_array` to `tests/numba_cuda_tests/cudapy/test_vector_type.py` to ensure vector types can be successfully allocated, written to, and read from in both memory spaces.

## Test Plan
- Ran the newly added `test_vector_local_array` and `test_vector_shared_array` tests successfully.
- Verified that `ruff check` passes.

Blocked by #36


Jlisowskyy (approved):
Looks good, pretty straightforward

atmnp (approved):

src/numba_cuda_mlir/lowering/cuda.py
```diff
@@ -302,43 +302,31 @@ def cuda_static_shared_memory(lower: MLIRLower, target, static_shape, dtype, ali
...
-@lower(cuda.shared.array, types.Tuple, types.NumberClass, types.IntegerLiteral)
-@lower(cuda.shared.array, types.Tuple, types.DType, types.IntegerLiteral)
+@lower(cuda.shared.array, types.Tuple, types.DTypeSpec, types.IntegerLiteral)
 @lower(cuda.shared.array, types.Tuple, types.StringLiteral, types.IntegerLiteral)
-@lower(cuda.shared.array, types.Tuple, types.NumberClass, types.NoneType)
-@lower(cuda.shared.array, types.Tuple, types.DType, types.NoneType)
+@lower(cuda.shared.array, types.Tuple, types.DTypeSpec, types.NoneType)
 @lower(cuda.shared.array, types.Tuple, types.StringLiteral, types.NoneType)
 def cuda_shared_memory(lower: MLIRLower, target, args: list[Any], kwargs: list[tuple[str, Any]]):
     shape, dtype, alignas = _extract_shape_and_dtype(*args, **dict(kwargs))
```
Jlisowskyy, line 332: I noticed one thing regarding alignment. The default is 8, but since vector types usually need specific alignments for specialized instructions, should we be enforcing that for them?

ZzEeKkAa: Array alignment is always greater or equal to 8. I think it is overkill and we should get rid of it. However it will be at least alignment of dtype, so no extra steps needed here

ZzEeKkAa: But what you said is true for the view. We either need to provide mechanism to prevent users from shooting into the leg, or warn them that with views they are on their own with alignment. The last option is fine, until we provide robust alignment support for all operations. I know there was some work around that in numba-cuda, but I'm not sure what is the current state

PR #39, opened 2026-05-12
MLIR Translation: Implement Barrier0Op

gmarkall:
**EDIT:** I think maybe the git history was rewritten after I built llvm-modern, so although there was only one commit to llvm-version.env, I think the version changed out from under my existing build.

Locally I'm getting a lot of failures like:

```
FAILED tests/numba_cuda_tests/cudapy/test_atomics.py::TestCudaAtomics::test_atomic_add - RuntimeError: llvm70 translation failed: unsupported NVVM op: nvvm.barrier0
```

I see it was removed last week in https://github.com/llvm/llvm-project/pull/195608, so I'm not sure if this is because I have some misconfiguration in my setup (non-matching versions of libraries / LLVMs) or because my setup is just a little different from that in CI and I'm running into a legitimate issue, but adding an implementation of the `NVVM::Barrier0Op` seems to resolve it - making this PR for comments / discussion.

### DIFF
--- a/cext/mlir-llvm70/include/llvm70/LLVM70Target.h
+++ b/cext/mlir-llvm70/include/llvm70/LLVM70Target.h
@@ -153,6 +153,7 @@ private:
   llvm::Error translateClusterArriveOp(mlir::Operation *op, bool isRelaxed);
   llvm::Error translateClusterWaitOp(mlir::Operation *op);
   llvm::Error translateBarrierOp(mlir::Operation *op);
+  llvm::Error translateBarrier0Op(mlir::Operation *op);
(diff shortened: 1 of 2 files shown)


VijayKandiah:
Yes, barrier0 op was recently removed and we bumped the llvm version we pull in to be newer than [this change ](https://gitlab-master.nvidia.com/nvhpc-eng/mlir-nvvm7/-/commit/a10ea62059edb6009bc29c306d507662a652a564)to account for Jie's upstream changes. We shouldn't need to add support for it. This is most likely a local environment misconfiguration. Perhaps you have bindings from an older LLVM installed locally?

PR #40, opened 2026-05-12
[CompileTime] Replace implicit context refresh with explicit initialization and version-tracked registries + misc changes to registrations

atmnp:
The typing and target contexts previously refreshed themselves implicitly on every compilation, re-walking all registries and re-filtering builders each time. This change introduces version counters on both typing and target registries so that install_registry and refresh become no-ops when nothing has changed. It adds ensure_initialized() to both MLIRTarget and CUDATarget to perform one-time setup at well-defined points (first compilation, declare_device), and replaces the _assume_builder_is_usable_for_mlir module-name heuristic in the lowering layer with a clean registry-only dispatch that no longer needs to filter out Numba-origin builders at lookup time. Device declarations (declare_device) are refactored into a deferred DeviceDeclaration dataclass that records registrations globally and applies them when a target initializes, eliminating import-order coupling between bf16/fp16 modules and the target descriptors. The MLIR module string serialization is also deferred via get_mlir_module_str, and the optimization pass now operates directly on the in-memory module rather than re-parsing from a string.

For users of the extending registries (extending.typing_registry, extending.lowering_registry, @overload, @intrinsic, register_jitable), the key change is that registrations made after the compiler contexts have already been initialized are no longer picked up automatically. After adding new entries to the registries, callers must now invoke extending.refresh_registries() to sync the compiler contexts. This function is exported from numba_cuda_mlir.extending and handles both the MLIR and CUDA target descriptors. Registrations that happen at import time before any kernel is compiled continue to work without an explicit refresh, since ensure_initialized() installs all pending registrations on first use.


src/numba_cuda_mlir/compiler.py
```diff
@@ -432,8 +436,10 @@ def compile_mlir(pyfunc, sig, optimized=False, **targetoptions):
         cres = _compile_and_optimize(pyfunc, sig, targetoptions)
         return cres.metadata["mlir_module_optimized"]
     else:
+        from numba_cuda_mlir.mlir_lowering import get_mlir_module_str
+
         cres = _compile_only(pyfunc, sig, targetoptions)
```
VijayKandiah, line 447:
```suggestion
        mlir_target.ensure_initialized()
        cres = _compile_only(pyfunc, sig, targetoptions)
```

`_compile_only` calls `mlir_compiler.compile_mlir()` directly, bypassing `MLIRDispatcher.compile()` which internally invokes `mlir_target.ensure_initialized()`. With the `get_function()` auto-refresh fallback removed [here](https://github.com/NVIDIA/numba-cuda-mlir/pull/40/changes#diff-0e567f5c1571178b6de141b942ce19f2c3a1499613ab5a4487b8d8197fee4f57L594-L599), this path has no initialization guarantee — type inference fails with `ValueError` because the typing context has zero registries loaded.


The `optimized=True` path is fine (goes through `_compile_and_optimize` -> `MLIRDispatcher.compile()`), but the `optimized=False` path is uncovered.

Repro:
```python
from numba_cuda_mlir import types
from numba_cuda_mlir.compiler import compile_mlir
def simple_kernel(x):
    x[0] = 1
sig = types.void(types.Array(types.int64, 1, 'C'))
compile_mlir(simple_kernel, sig, optimized=False)  # ValueError
```
gives error:
```
ValueError: Cannot determine Numba type of <class 'builtin_function_or_method'>
```

VijayKandiah (changes requested):
LGTM for the most part. Raised one concern above.

VijayKandiah (approved):
LGTM, thanks!

isVoid:
Corresponding change required in Numbast: https://github.com/NVIDIA/numbast/pull/376. All green testing against this PR.

PR #41, opened 2026-05-12
Try to fix publish workflow

leofang:
The previous attempt to trigger the publish workflow has failed:
https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/25764823361
because the old workflow was removed and the filename does not pass the syntax check. Try to bring the workflow up-to-date to unblock v0.3.0.


leofang:
Since this is not testable by the CI, let me admin-merge and retry.

PR #44, opened 2026-05-13
Add support for type casting between CUDA vector types and complex numbers

ZzEeKkAa:
## Summary
This PR introduces the ability to cast 2-element CUDA vector types (e.g., `float2`, `double2`, `int2`) to Python `complex` types and vice versa. This improves interoperability between CUDA vector types and complex math operations.

## Changes
* **Typing (`typing/cuda_vector_types.py`)**: 
  * Registered the `complex` builtin for 2-element vector types. It automatically resolves to `complex64` for `float32` based vectors and `complex128` for other types.
  * Updated the vector constructor templates to accept `types.Complex` as an argument when constructing 2-element vectors.
* **Lowering (`lowering/cuda_vector_types.py`)**: 
  * Implemented MLIR lowering for `complex(vector)` which extracts the `x` and `y` components, converts them to the target precision, and constructs an MLIR complex value using the `complex` dialect.
  * Updated vector constructor lowering to properly extract the real and imaginary parts when a complex number is passed as an argument.
* **Testing (`tests/numba_cuda_tests/cudapy/test_vector_type_complex_cast.py`)**: 
  * Added a new test suite to verify vector-to-complex casts, complex-to-vector casts, and complex argument handling.
  * Added negative tests to ensure that invalid casts (e.g., trying to cast a 4-element vector like `float4` to `complex`) correctly raise a `TypingError`.

## Test Plan
- Run the newly added test suite: `pytest tests/numba_cuda_tests/cudapy/test_vector_type_complex_cast.py`
- All tests pass successfully.

### DIFF
--- a/src/numba_cuda_mlir/lowering/cuda_vector_types.py
+++ b/src/numba_cuda_mlir/lowering/cuda_vector_types.py
@@ -77,6 +77,11 @@ def _constructor_lowering(lower_ctx: MLIRLower, target, args: list[Any], kwargs)
 
         if isinstance(arg_type, VectorType):
             scalars.extend(_extract_vector_elements(val))
+        elif isinstance(arg_type, types.Complex):
+            from numba_cuda_mlir._mlir.dialects import complex as complex_dialect
+
+            scalars.append(complex_dialect.re(val))
+            scalars.append(complex_dialect.im(val))
         else:
             scalars.append(val)
 
@@ -111,3 +116,27 @@ lower_getattr(VectorType, "x")(_make_attr_lowering("x"))
 lower_getattr(VectorType, "y")(_make_attr_lowering("y"))
 lower_getattr(VectorType, "z")(_make_attr_lowering("z"))
 lower_getattr(VectorType, "w")(_make_attr_lowering("w"))
+
+
+@_raw_lower(complex, VectorType)
+def _complex_from_vector_lowering(lower_ctx: MLIRLower, target, args: list[Any], kwargs):
+    """Lowering for complex(vector_type)."""
+    target_type = lower_ctx.get_numba_type(target.name)
+    mlir_target_type = lower_ctx.get_mlir_type(target_type)
+
+    val = lower_ctx.load_var(args[0])
+
+    real = vector.extract(val, [], [0])
+    imag = vector.extract(val, [], [1])
+
+    real = convert(real, mlir_target_type.element_type)
+    imag = convert(imag, mlir_target_type.element_type)
+
+    from numba_cuda_mlir._mlir.dialects import complex as complex_dialect
+
+    result = complex_dialect.create_(
+        complex=mlir_target_type,
+        real=real,
+        imaginary=imag,
+    )
+    lower_ctx.store_var(target, result)
--- a/src/numba_cuda_mlir/lowering/cuda_vector_types.py
+++ b/src/numba_cuda_mlir/lowering/cuda_vector_types.py
@@ -19,6 +19,7 @@ from numba_cuda_mlir.cuda.vector_types import _vector_types
 from numba_cuda_mlir.type_defs.vector_types import VectorType
 from numba_cuda_mlir import types
 from numba_cuda_mlir._mlir.dialects import vector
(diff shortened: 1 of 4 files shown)


tests/test_vector_type_complex_cast.py
atmnp, line 1: Let's move this test to `tests/`, and make this apache licensed, all new test files are numba-cuda-mlir, hence Apache, not BSD.

src/numba_cuda_mlir/lowering/cuda_vector_types.py
```diff
@@ -111,3 +116,27 @@ def attr_lowering(context, builder: MLIRLower, target, value):
...
+
+    val = lower_ctx.load_var(args[0])
+
+    real = vector.extract(val, [], [0])
+    imag = vector.extract(val, [], [1])
+
+    real = convert(real, mlir_target_type.element_type)
+    imag = convert(imag, mlir_target_type.element_type)
+
+    from numba_cuda_mlir._mlir.dialects import complex as complex_dialect
```
Jlisowskyy: I think we can move this imports to module, same for `_constructor_lowering`

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -153,3 +156,14 @@ def typeof_vector_type(val, c):
...
+
+
+@registry.register_global(complex)
+class ComplexBuiltinTemplate(AbstractTemplate):
+    def generic(self, args, kws):
+        if len(args) == 1 and isinstance(args[0], VectorType) and args[0].length == 2:
+            if args[0].dtype == types.float32:
+                return signature(types.complex64, args[0])
+            else:
+                return signature(types.complex128, args[0])
```
Jlisowskyy: This looks a bit off. Shouldn't we map types <= 32 bits to `complex64` and wider types to `complex128`?

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -153,3 +156,14 @@ def typeof_vector_type(val, c):
     """Create a typeof implementation for a vector type."""
     template = make_constructor_template(val)
     return VectorTypeClass(val, template)
+
+
+@registry.register_global(complex)
+class ComplexBuiltinTemplate(AbstractTemplate):
```
Jlisowskyy, line 162: Are we missing `complex32` in `numba-cuda-mlir`? I believe we need it for nvmath-python

src/numba_cuda_mlir/typing/cuda_vector_types.py
ZzEeKkAa: Nope. That looks correct. int32 won't fit into float32 without loosing precision.

ZzEeKkAa: Keep in mind. This is only for `complex`, not for `np.complex64` or `np.complex128`

Jlisowskyy: Right, good point on int32. But what about float16? It widens losslessly to f32

ZzEeKkAa: True. I have an ongoing work to add support for complex32 from ml_dtype package. Once it is done I would love to cast float16 to complex32 directly

src/numba_cuda_mlir/typing/cuda_vector_types.py
ZzEeKkAa: Yes, it is not exposed in numpy, but it is a part of ml_dtypes. I have working in progress for that support

ZzEeKkAa: But it is not a blocker now, as we did not have that for legacy numba

src/numba_cuda_mlir/typing/cuda_vector_types.py
ZzEeKkAa: Added support for f<=32 and i<=16

Jlisowskyy (approved):
LGTM

PR #45, opened 2026-05-13
Add basic vector/number ops

ZzEeKkAa:
## Summary
- Added typing and lowering support for basic arithmetic operations between CUDA vector types and scalars (`+`, `-`, `*`, `/`, `//`, `%`, and their in-place variants).
- Added typing and lowering support for vector-vector addition and subtraction.
- Added typing and lowering support for unary operators (`-`, `abs`) on CUDA vector types.
- Restricted vector-vector multiplication and division, raising a `TypingError` at compile time because their behavior is undefined or context-dependent.
- Implemented type promotion and unification for vector operations, handling cases like `float32x4 + float64` correctly.
- Added comprehensive tests for all supported vector-scalar and vector-vector operations, as well as tests verifying that unsupported operations fail to compile.

## Test plan
- [x] Run `pytest tests/test_vector_types_ops.py` to ensure all vector arithmetic tests pass.
- [x] Verify that `vector * vector` and `vector / vector` operations correctly raise a `TypingError`.
- [x] Ensure that scalar broadcasting works correctly in both `scalar <op> vector` and `vector <op> scalar` cases.
- [x] Ensure that type promotion (e.g., `float32x4` + `float64` -> `float64x4`) works correctly.


atmnp (approved):
LGTM

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -122,6 +122,124 @@ def generic(self, args, kws):
...
+
+                if op == operator.truediv and isinstance(target_dtype, types.Integer):
+                    bitwidth = max(target_dtype.bitwidth, 32)
+                    target_dtype = float_of_width(bitwidth)
+
+                restype = _get_vector_type(target_dtype, lhs.length)
+                if restype is None:
+                    print(f"restype is None for {target_dtype} and {lhs.length}")
+                    return None
+                print(f"Returning signature {restype} {lhs} {rhs}")
```
Jlisowskyy: Some print leftovers from debugging

src/numba_cuda_mlir/lowering_utilities/__init__.py
```diff
@@ -606,6 +611,34 @@ def unverified_basic_mlir_convert(
...
+                return (
+                    arith.sitofp(out=target_type, in_=value)
+                    if elem1.width > 1
+                    else arith.uitofp(out=target_type, in_=value)
+                )
+            elif isinstance(elem1, ir.FloatType) and isinstance(elem2, ir.IntegerType):
+                return (
+                    arith.fptosi(out=target_type, in_=value)
+                    if elem2.width > 1
+                    else arith.fptoui(out=target_type, in_=value)
```
Jlisowskyy, line 648: Will it work when `elem1.width == elem2.width`? When I tried to perform noop casts, I had some failures on MLIR verification

src/numba_cuda_mlir/typing/cuda_vector_types.py
```diff
@@ -122,6 +122,124 @@ def generic(self, args, kws):
...
 
+def _get_vector_type(dtype, length):
+    for vt in _vector_types:
+        if vt.dtype == dtype and vt.length == length:
+            return vt
+    return None
+
+
+import operator
+from numba_cuda_mlir.lowering_utilities.type_conversions import float_of_width
```
Jlisowskyy: Shoudn't those be on top of the file?

src/numba_cuda_mlir/lowering/cuda_vector_types.py
```diff
@@ -111,3 +111,103 @@ def attr_lowering(context, builder: MLIRLower, target, value):
 lower_getattr(VectorType, "y")(_make_attr_lowering("y"))
 lower_getattr(VectorType, "z")(_make_attr_lowering("z"))
 lower_getattr(VectorType, "w")(_make_attr_lowering("w"))
+
+
+import operator
+from numba_cuda_mlir.lowering_utilities import _get_mlir_bin_op_for_operator
+from numba_cuda_mlir._mlir.dialects import arith
```
Jlisowskyy: Same here

src/numba_cuda_mlir/lowering/cuda_vector_types.py
```diff
@@ -111,3 +111,103 @@ def attr_lowering(context, builder: MLIRLower, target, value):
...
+    operator.truediv,
+    operator.itruediv,
+    operator.floordiv,
+    operator.ifloordiv,
+    operator.mod,
+    operator.imod,
+]:
+    _raw_lower(op, VectorType, VectorType)(_make_vector_binop_lowering(op))
+    _raw_lower(op, VectorType, types.Number)(_make_vector_binop_lowering(op))
+    _raw_lower(op, types.Number, VectorType)(_make_vector_binop_lowering(op))
```
Jlisowskyy, line 175: I think it would be less confusing if we maintained the same lowering support as typing (vector-vector is only supported for add/sub).

Jlisowskyy (dismissed):

src/numba_cuda_mlir/lowering_utilities/__init__.py
ZzEeKkAa: I'm not sure what you mean. https://mlir.llvm.org/docs/Dialects/ArithOps/#arithfptoui-arithfptouiop

src/numba_cuda_mlir/lowering/cuda_vector_types.py
ZzEeKkAa: The supported functionality is the intersection of typing and lowering. Lowering is the second stage. If the code is cleaner, but lowering is a superset of typing I think it's fine

src/numba_cuda_mlir/lowering_utilities/__init__.py
Jlisowskyy: For example, `arith.extf` requires the destination element type to be strictly wider than the source: [docs](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithextf-arithextfop) 

With the current:

```
if elem1.width > elem2.width: 
    truncf
else:
    extf
```

the equal-width case falls into the else and emits extf, which would probably fail verification. I'm not sure whether that branch is actually reachable in practice, but if it is, it may be an issue

PR #50, opened 2026-05-15
[DOCS] Add Extension API Documentation

gmarkall:
Adds a new Extension API section, derived from a combination of the existing Numba Extension API docs, and the implementations of extensions internally within Numba-CUDA-MLIR.

The sections of the documentation are:

- High-level API: `@overload`, etc - similar to Numba's high-level API.
- Low-level API: Typing and lowering - the lowering is new to Numba-CUDA-MLIR (as it is not llvm-based).
- MLIR dialect primer - a brief overview of the dialects used in Numba-CUDA-MLIR.
- MLIR dialect reference - Lists of common operations with the dialects used in Numba-CUDA-MLIR, with links to upstream dialect documentation.

The last two sections are mainly provided for those new to MLIR-based lowering to find their way through the APIs - many of those used in Numba-CUDA-MLIR are linked, to aid in searching for examples of their use in Numba-CUDA-MLIR, and as a starting point for research. The MLIR documentation doesn't aim to provide a comprehensive introduction to MLIR.

I'm open to whether either of the two last sections should exist - I get the impression that they will be helpful outlines, even though they're not detailed and comprehensive, but I am open to suggestions otherwise.

The documentation was initially generated with Claude Code, with the following changes on top:

- High- and low-level API documentation text was extensively edited and re-written. Although it contained facts, I did not think it was written in a helpful manner for someone trying to use the docs to get started with the extension APIs.
- The MLIR dialect documentation was lightly edited, and I checked a subset of the references it makes to ensure that they appeared to be correct. I think this section needed less editing because it is more of a summary / list of items rather than an explanation of concepts / a tutorial.

Some docstrings were also added to the high-level API code, to make it a little easier to work with inside IDEs.


VijayKandiah (approved):
LGTM to merge after landing the changes we discussed offline.

PR #51, opened 2026-05-18
Test to-be-merged cuDF numba-cuda-mlir extension backend

brandon-b-miller:
This PR creates a CI job that builds an unmerged branch of _my fork_ of `cudf` from source. This branch contains an MLIR version of cuDF's numba-cuda extension. We'll need to build it from source to test it until it's officially merged and the code appears in nightlies we can pull rather than build.

### DIFF
--- a/ci/test-matrix.yml
+++ b/ci/test-matrix.yml
@@ -69,4 +69,7 @@ linux:
   expensive:
     - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'h100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'arm64', PY_VER: '3.13',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
+  cudf-source:
+    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
   nightly: []
--- a/ci/test-matrix.yml
+++ b/ci/test-matrix.yml
@@ -72,4 +72,6 @@ linux:
   cudf-source:
     - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'arm64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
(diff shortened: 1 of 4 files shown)


ci/tools/run-cudf-source-tests
```diff
@@ -0,0 +1,178 @@
...
+CUDF_BRANCH="${CUDF_BRANCH:-cudf-numba-cuda-mlir-extension-backend}"
+
+export CUDA_VERSION="${TEST_CUDA_MAJOR}.${TEST_CUDA_MINOR}"
+
+MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-$(uname -m).sh"
+CONDA_ENV_NAME="cudf_build"
+
+declare -A CUDA_TO_ENV_YAML=(
+  ["12.9"]="all_cuda-129_arch-x86_64.yaml"
+  ["13.2"]="all_cuda-132_arch-x86_64.yaml"
```
VijayKandiah, line 27: Should we have the same CTK matrix for cuDF CI as other CI tests? i.e., [12.2.2, 12.9.1, 13.0.2, 13.2.1]

ci/test-matrix.yml
```diff
@@ -69,4 +69,7 @@ linux:
   expensive:
     - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'h100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'arm64', PY_VER: '3.13',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
+  cudf-source:
+    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
```
VijayKandiah, line 74: The cudf-source test matrix only has `amd64`  and no `arm64`. Is this intentional? Our main CI tests both architectures. Similarly, should we add other GPU archs (or atleast one older arch volta) like the main CI instead of testing only on `sm_89`?

ci/tools/run-cudf-source-tests
```diff
@@ -0,0 +1,178 @@
...
+  -o addopts= \
+  "$@" || true
+
+conda run -n "${CONDA_ENV_NAME}" python -m pytest \
+  "${GROUPBY_UDF}" \
+  -k "test_groupby_apply" \
+  "${COMMON_ARGS[@]}" \
+  --junit-xml=cudf-source-junit-groupby.xml \
+  -o addopts= \
+  "$@" || true
```
VijayKandiah, line 132: The `|| true` on both pytest invocations could make infrastructure failures like import error, broken conda env, wrong test path get silently swallowed as it moves on to the next step. The JUnit XML parser below seems to only warn if an XML file is missing and doesn't fail. So if both test runs crash before producing any output, the job exits 0. Would CI be green in this case even when no tests are run? If so, it would make sense to add a sanity check like 
```
if total_passed + total_failures + total_errors == 0:
    print("ERROR: No test results found — tests may not have run")
    sys.exit(1)
```

VijayKandiah (changes requested):

ci/tools/run-cudf-source-tests
brandon-b-miller: Unfortunately these are the only env files shipped to support a cuDF source build right now. When we go to packages, we should be able to delete this.

ci/test-matrix.yml
brandon-b-miller: Added.

brandon-b-miller: Let's wait on packages for older arch's. I worry the cudf build will hog runners here.

ci/tools/run-cudf-source-tests
brandon-b-miller: Fixed

ci/test-matrix.yml
VijayKandiah: Sounds good!

VijayKandiah (approved):
LGTM!

PR #52, opened 2026-05-18
Fix ICE for raise-only kernel

jiel-nv:
A kernel whose only statement is `raise` ICE'd with `entry block of region may not have predecessors`.

MLIRLower.lower_static_raise unconditionally emitted `cf.br ^bb<max>` assuming the highest-offset block was a Return block; for a raise-only kernel that block is the entry block, producing an illegal self-loop.

Fix: branch to the highest-offset Return block when one exists; otherwise terminate the current block in place with `gpu.return` (kernels) or `func.return` (device functions).

Add a regression test.


VijayKandiah (approved):
LGTM

PR #54, opened 2026-05-19
Prevent Numba-CUDA JIT being invoked and address resulting issues

gmarkall:
This is a first step towards removing the llvmlite dependency. The central change is in the first commit (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/7c51dfe2eada6bc26917d64f6f93fee191e7aff3), which changes the typing template for overloads to be hardcoded to return the `mlir_jit` decorator, instead of the `cuda.jit` decorator from vendored Numba-CUDA.

This change causes various issues to surface, that were previously hidden by the fact that the vendored Numba-CUDA was used to compile various things that could not necessarily be compiled with Numba-CUDA-MLIR, and were not present as "internal extensions" to Numba-CUDA-MLIR. Several implementations are added to Numba-CUDA-MLIR to support these items - including `np.nditer`, to make many NumPy function implementations work again.

There are some Numba-CUDA extensions that duplicate already-existing ones in Numba-CUDA-MLIR - these are a hasattr implementation (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/37e3d5ff1a9d61cc8db5a30e1adb17ff9ade7398) and cooperative groups APIs (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/25fafc1a150d0e92e4b5f94c75972418bd758546). To get around this, these need to be removed from the vendored Numba-CUDA.

Some new support is needed for the extensions to work - these include a data model for unicode types and getattr on structs (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/144a5d118d43d61e51306642c8ce4439b499d9dd), and lowering for dtype objects (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/d0a622973f60938f65ae55f3a4d7cf38a3011eed). 

With all this fixed up, we seem to hit a path where using extensions can use MLIR functions before a context has been created. I coudn't find a great way to ensure that this doesn't happen, so I placed it at the same time as we override the target before compilation in https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/bfe73009b80d29846b659a83994a3338f7e9ff63.

There are some tests that had to be marked as xfails, which were previously depending on Numba-CUDA keeping them passing:

- `compile_all` tests. This is not actually implemented in Numba-CUDA-MLIR, it was just calling Numba-CUDA's `compile_all`. (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/9d041dc7b6d78eefe92f6e955aab7fb66293a474)
- Tests that used `np.array` - this is not implemented in Numba-CUDA-MLIR. They only worked before because the `numpy.array` overload was compiled by Numba-CUDA during typing, then the arrays were optimized away. (https://github.com/NVIDIA/numba-cuda-mlir/pull/54/changes/73c99abb60402b08a9088948c044ce2fb065d2ee)

Following this, the next step is to delete all `@intrinsic` functions in `numba_cuda_mlir.numba_cuda`, but I will put that in a separate PR as this one is already sizeable. The `@intrinisic` functions in Numba-CUDA-MLIR should also be removed (see #74), but that is tangential to removing llvmlite.


src/numba_cuda_mlir/mlir_compiler.py
```diff
@@ -348,10 +348,12 @@ def compile_mlir(pyfunc, return_type, args, targetoptions: Dict[str, Any]):
...
     # Run compilation pipeline
     from numba_cuda_mlir.numba_cuda.core.target_extension import target_override
+    from numba_cuda_mlir.lowering_utilities import context as mlir_ctx
+    from numba_cuda_mlir._mlir import ir as mlir_ir
 
     flags.lto = is_lto
 
-    with target_override("numba_cuda_mlir"):
+    with target_override("numba_cuda_mlir"), mlir_ctx.get_context(), mlir_ir.Location.unknown():
         cres = compiler.compile_extra(
```
gmarkall, line 357: May also be used when we set up the MLIR module for the kernel, and in the runtime. Need to break in get_context() and check the stack trace to see where else it gets called from.

atmnp (approved):
LGTM, nothing too weird going on here. Thanks for all this work! Going through the commits, it seems like a mess of bugs you've uncovered.

PR #56, opened 2026-05-19
Fix array.real/imag on shared-memory arrays losing address space

VijayKandiah:
`array.real` / `array.imag` on shared-memory complex arrays fails at runtime with `CUDA_ERROR_ILLEGAL_ADDRESS`.

Root cause: `_lower_array_complex_real_imag` extracted the data pointer via `extract_aligned_pointer_as_index` then rebuilt it with `llvm.inttoptr(llvm.PointerType.get(), ...)` — creating a generic (addrspace 0) pointer regardless of where the source array lived.

Fix: Replace the manual LLVM struct construction with `memref.reinterpret_cast`, which preserves the source memref's memory space. Follows the same pattern used by `lower_array_view_cg`.


Jlisowskyy (approved):
lgtm

PR #57, opened 2026-05-19
Fix multi-assign tuple, Optional lowering, string constant-folding, and VectorType→complex setitem

VijayKandiah:
This PR addresses four MLIR lowering issues:

1. **Multi-assigned tuple-typed locals** — Allocate per-element memref stack slots for `UniTuple` variables assigned from multiple branches, avoiding a `MemRefType.get` TypeError on tuple types.
2. **Device functions returning `Optional` (int | None)** — Register `OptionalModel`, add cast-to/from-optional via `llvm.insertvalue`/`extractvalue`, handle None assignments to Optional-typed variables, and unwrap Optional values in setitem lowering.
3. **String literal equality constant-folding ** — In the UnicodeType equality lowering, Constant-fold `==` when both operands resolve to Python strings at lowering time, emitting a single `arith.constant` instead of an `scf.for` byte-comparison loop.
4. **Missing `setitem` for `float64x2` → `complex128[:]`** — Extend `ArraySetitemTemplate` to accept 2-lane `VectorType` values when the target array has complex dtype.

Also removed `xfail` from 2 tests using `Optional` type that now pass.

### DIFF
--- a/src/numba_cuda_mlir/lowering/builtins.py
+++ b/src/numba_cuda_mlir/lowering/builtins.py
@@ -800,6 +800,21 @@ def operator_is_none_none_lower(builder, target, args, kwargs):
     builder.store_var(target, result)
 
 
+@lower(operator.is_, types.Optional, types.NoneType)
+@lower(operator.is_, types.NoneType, types.Optional)
+def operator_is_optional_none_lower(builder, target, args, kwargs):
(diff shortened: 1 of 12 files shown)


Jlisowskyy (approved):
LGTM! Thanks!

atmnp (approved):

PR #67, opened 2026-05-20
[codex] Add Numbast MLIR CI integration

isVoid:
## Summary
- add a CI helper that installs the just-built numba-cuda-mlir wheel and runs Numbast's experimental MLIR tests
- checkout public NVIDIA/numbast at pinned commit fb743408b07e52beb4980f858ceb648b4e15f619 only for the Numbast MLIR lane
- add one linux-64 and one linux-aarch64 CI job, both on Python 3.12 / CUDA 13.2.1 / local CTK, and include them in aggregate checks

## Validation
- bash -n ci/tools/run-numbast-mlir-tests
- parsed changed workflow YAML with Python yaml.safe_load
- simulated both Numbast matrix filters with jq
- git diff --check
- commit hooks: SPDX, YAML, EOF, whitespace, merge-conflict, large-file checks

PR #68, opened 2026-05-20
Add Numbast MLIR source CI tests

isVoid:
## Summary
- add a dedicated Numbast MLIR reusable workflow modeled after the cuDF source workflow from #51
- add `numbast-mlir` matrix coverage for Python 3.12 on CUDA 12.9.1 and 13.2.1 for both Linux architectures: `linux-64`/L4 and `linux-aarch64`/A100
- checkout public `NVIDIA/numbast` at pinned commit `f779d7d7f834a7e8b411b4d0bfe6c2fb36a58859` from Numbast PR #365 and run its experimental MLIR test subtree against the just-built `numba-cuda-mlir` wheel
- provision the Numbast MLIR test lane with Miniforge and `clangdev=18`, which matches the supported ast_canopy LLVM range and provides Clang resource headers
- make `fetch_ctk` parse CUDA redistrib metadata with `jq` instead of ambient `python`, so Miniforge-only jobs can fetch CUDA 12.9.1 mini CTK components
- include the Numbast MLIR jobs in the aggregate `checks` job while preserving the cuDF source jobs from #51

## Validation
- `bash -n ci/tools/run-numbast-mlir-tests`
- YAML parse for `.github/actions/fetch_ctk/action.yml`, `.github/workflows/test-numbast-mlir-linux.yml`, and `ci/test-matrix.yml`
- local `jq` selector test for `.[$component][$subdir].relative_path`
- Python matrix assertion confirms four Numbast MLIR rows: amd64/12.9.1, amd64/13.2.1, arm64/12.9.1, arm64/13.2.1
- `git diff --check`
- verified the Numbast pin matches the current PR #365 head `f779d7d7f834a7e8b411b4d0bfe6c2fb36a58859`

Supersedes #67, which was based on a stale local `main` and did not include the cuDF source CI jobs from #51.

### DIFF
--- a/.github/actions/fetch_ctk/action.yml
+++ b/.github/actions/fetch_ctk/action.yml
@@ -54,8 +54,8 @@ runs:
       uses: ./.github/actions/install_unix_deps
       continue-on-error: false
       with:
-        dependencies: "zstd curl xz-utils"
-        dependent_exes: "zstd curl xz"
+        dependencies: "zstd curl xz-utils jq"
+        dependent_exes: "zstd curl xz jq"
 
     - name: Download CTK cache
       id: ctk-get-cache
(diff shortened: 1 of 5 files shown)


isVoid:
Depends on https://github.com/NVIDIA/numbast/pull/363

isVoid:
The tests finally seems to pass! But I forgot to add cuda 12.9.1 in the test grid 😓

atmnp (approved):
LGTM

PR #71, opened 2026-05-20
Keep default CUDA JIT output on PTX

### LINKED ISSUE #64: Keep default CUDA JIT output on PTX unless LTO is requested
## Problem

When nvJitLink is available, default `cuda.jit` configuration can implicitly route kernels through LTO/nvJitLink. For large generated-kernel workloads, this makes every ordinary kernel pay the LTO path cost even when the user did not request LTO or `output="ltoir"`.

Cooperative CUDA code generation can create many specialized kernels in one process. Keeping the default path on PTX avoids unnecessary nvJitLink overhead and leaves LTO as an explicit opt-in.

## Example

```python

@cuda.jit
def kernel(out):
    out[0] = 1

# Expected default policy: this should compile on the stable PTX path.
assert kernel.targetoptions["lto"] is False

@cuda.jit(lto=True)
def lto_kernel(out):
    out[0] = 1

assert lto_kernel.targetoptions["lto"] is True

@cuda.jit(output="ltoir")
def ltoir_kernel(out):
    out[0] = 1

assert ltoir_kernel.targetoptions["lto"] is True
```

## Expected behavior

`cuda.jit` should default to PTX linking unless the user explicitly requests LTO with `lto=True` or requests LTO IR with `output="ltoir"`.

Existing explicit LTO requests should continue to work. The target option should still track whether `lto` was explicit so downstream behavior can distinguish implicit defaults from user intent.

## Implementation notes

The local fix has already been implemented and a PR will be forthcoming.

The implementation changes `verify_target_options()` so default `lto` is `False`, while preserving explicit `lto=True` and `output="ltoir"` behavior. The local implementation is currently represented by commit `aa8d389` (`Keep MLIR JIT default linking on PTX`) on the cooperative CUDA work branch.

Tests should cover:

- default `cuda.jit` sets `lto` to `False`;
- explicit `lto=True` is preserved;
- `output="ltoir"` enables LTO;
- `_lto_explicit` continues to reflect direct user intent.

Issue created with GPT-5.5 xhigh, Codex 0.131.0.

tpn:
Closes #64.

## Summary

- Keep default MLIR CUDA JIT kernels on the PTX linking path by setting implicit `lto` to `False` for kernels with no external link inputs.
- Preserve explicit user policy for `lto=True`, `lto=False`, `output="ptx"`, and `output="ltoir"`.
- Enable implicit LTO only when an LTOIR link input is present or discovered during lowering, unless callbacks/debug or explicit output/LTO choices require PTX.
- Keep plain CUDA source link inputs (`.cu` / `CUSource`) on the PTX path by default unless the user explicitly enables LTO. This avoids forcing regular source links through nvJitLink.
- Preserve LTO when `lineinfo=True` and LTOIR link inputs require the LTO path, so profiling paths can keep optimized codegen.
- Preserve `inspect_lto_ptx()` for PTX-default kernels by recompiling optimized MLIR to LTOIR on demand and replaying discovered external link items into the diagnostic LTO linker.
- Preserve Numbast-style retained globals by forwarding `kernels_used` through the linker wrapper and rooting the generated entry kernel when `variables_used` is configured.
- Reject contradictory LTOIR requests when the resolved JIT options leave LTO disabled, including `output="ltoir", lto=False`, `output="ptx"` with LTOIR link inputs, `.ltoir` link inputs with `lto=False`, unavailable `nvjitlink`, debug-disabled implicit LTO, and callback fallback paths.
- Keep MLIR disk-cache keys stable when implicit LTO is discovered during lowering, so cache load/save behavior tracks explicit user intent rather than a late target-option mutation.
- Treat `cuda.shared.array(0, dtype=...)` as dynamic shared memory and keep shared-memory LLVM GEPs conservative. This fixes existing dynamic shared-memory tests exposed by running the default path through PTX instead of LTO.

## Current status

- Rebased on `main` at `ef2009e70d39379fd98939489d14c88a8e00d14a`.
- Current PR head is `20c1713bab12a743f64bdc8900dd8cb6cf136cbc`.
- GitHub reports the PR as blocked while Graham's review remains changes-requested and the newly-triggered Actions run is pending.
- Prior head `3ff23cd795a8529834191f20c5ac9abd31e2e28d` exposed a cp314t/H100 CI failure in `tests/infra/test_link_libraries.py` when plain `.cu` device-source links were auto-promoted to nvJitLink LTO. The current head keeps implicit LTO narrowed to actual LTOIR inputs.
- Prior head `65c43e75098889a26e566a5b56c1d5f14a1c70a9` exposed rtxpro6000 failures in `TestInspect.test_inspect_lto_asm` because `inspect_lto_ptx()` fell back to regular PTX after plain source links stopped enabling normal LTO. The current head restores diagnostic LTO inspection without changing the normal PTX runtime path.
- Prior head `53adead7c33be3fd3703d48bc808b1d4a4044938` exposed Numbast MLIR failures in `test_module_link_variables_used_is_passed_to_mlir_linker`: `variables_used` retained globals but the CUDA linker pruned the entry kernel. The current head forwards `kernels_used` and roots the generated kernel when variables are explicitly retained.

## Validation

- `PYTHONPATH=$PWD/src /home/trentn/miniforge3/envs/coopcusimt312/bin/python -m pytest -o addopts='' tests/test_lto_defaults.py tests/numba_cuda_tests/cudadrv/test_linker.py::TestLinker::test_variables_used_linker_option tests/numba_cuda_tests/cudadrv/test_linker.py::TestLinker::test_variables_used_passed_to_cuda_core_linker tests/numba_cuda_tests/cudadrv/test_linker.py::TestLinker::test_public_linker_preserves_variables_used_when_recreated_with_lto tests/numba_cuda_tests/cudapy/test_inspect.py::TestInspect::test_inspect_lto_asm -q` (`40 passed`)
- `PYTHONPATH=$PWD/src /home/trentn/miniforge3/envs/coopcusimt312/bin/python -m pytest -o addopts='' tests/numba_cuda_tests/cudadrv/test_linker.py -q` (`32 passed, 1 skipped, 1 xfailed`)
- `/home/trentn/miniforge3/envs/coopcusimt312/bin/python -m ruff check src/numba_cuda_mlir/numba_cuda/cudadrv/driver.py src/numba_cuda_mlir/linker.py src/numba_cuda_mlir/mlir_optimization.py tests/numba_cuda_tests/cudadrv/test_linker.py tests/test_lto_defaults.py` (`All checks passed!`)
- `/home/trentn/miniforge3/envs/coopcusimt312/bin/python -m ruff format --check src/numba_cuda_mlir/numba_cuda/cudadrv/driver.py src/numba_cuda_mlir/linker.py src/numba_cuda_mlir/mlir_optimization.py tests/numba_cuda_tests/cudadrv/test_linker.py tests/test_lto_defaults.py` (`5 files already formatted`)
- `git diff --check origin/main...HEAD`

Full GitHub Actions matrix is being re-triggered for this head.


gmarkall (changes requested):
I don't think "LTO off by default" is likely to be a good blanket policy because it will result in performance regressions for anything that links to external objects and doesn't explicitly enable LTO in the kernel definition. The user isn't necessarily going to know when to enable LTO, because linkage of external code will often be hidden inside an extension.

I think a safer policy would be to leave LTO off when nothing external is linked to a kernel, and as soon as anything external (e.g. CUDA C++ code, another LTOIR library, etc.) is linked in, to enable LTO.

tpn:
I wholeheartedly agree re not making LTO off by default.  I'll iterate.

tpn:
/ok to test

-- GPT-5.5 xhigh, Codex v0.130.0
Signed-off-by: Trent Nelson <trent@trent.me>

src/numba_cuda_mlir/mlir_lowering.py
```diff
@@ -2048,6 +2050,13 @@ def link_external_item(self, link_item):
                 self.linker = self._create_linker()
                 for prior_link_item in self._linked_external_link_items:
                     self.linker.add_file_guess_ext(prior_link_item)
+        elif self._can_enable_implicit_lto_for_external_link_item():
+            self.targetoptions["lto"] = True
+            if not self._linker_config["lto"]:
+                self._linker_config["lto"] = True
+                self.linker = self._create_linker()
```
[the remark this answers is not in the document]
tpn: Addressed in 5366ca1 by moving the MLIR disk-cache option key to explicit user intent: implicit LTO no longer changes the load/save key when hidden external link items are discovered during lowering, while explicit lto=False remains distinct. Added focused cache-key coverage in tests/test_lto_defaults.py.

atmnp (commented):
@gmarkall does this look sane?

tests/test_lto_defaults.py
```diff
@@ -0,0 +1,285 @@
...
+    _set_nvjitlink_available(monkeypatch, False)
+
+    dispatcher = cuda.jit(link=["external.ltoir"])(_kernel)
+
+    assert dispatcher.targetoptions["lto"] is False
+    assert dispatcher.targetoptions["_lto_explicit"] is False
+
+
+def test_mlir_jit_preserves_explicit_lto_false_with_external_link_items():
+    dispatcher = cuda.jit(link=["external.ltoir"], lto=False)(_kernel)
```
gmarkall: I think this is an invalid use case - you cannot link LTOIR if you have LTO off. Maybe we should report an error if the user tries to do that.

tests/test_lto_defaults.py
```diff
@@ -0,0 +1,285 @@
...
+    dispatcher = cuda.jit(link=["external.ltoir"], debug=True, opt=False)(_kernel)
+
+    assert dispatcher.targetoptions["lto"] is False
+    assert dispatcher.targetoptions["_lto_explicit"] is False
+
+
+def test_mlir_jit_disables_implicit_lto_for_lineinfo_external_link_items(monkeypatch):
+    _set_nvjitlink_available(monkeypatch, True)
+
+    dispatcher = cuda.jit(link=["external.ltoir"], lineinfo=True)(_kernel)
```
gmarkall, line 479: This is the wrong behaviour. If the user requested lineinfo then the use case is going to be profiling with NSight Compute. If we disable LTO, that forces them to be profiling suboptimally-compiled code, which will make the profiling useless.

tests/test_lto_defaults.py
```diff
@@ -0,0 +1,285 @@
...
+
+    assert dispatcher.targetoptions["lto"] is False
+    assert dispatcher.targetoptions["_lto_explicit"] is False
+
+
+def test_mlir_jit_explicit_output_ptx_takes_precedence_over_link_items(monkeypatch):
+    _set_nvjitlink_available(monkeypatch, True)
+
+    dispatcher = cuda.jit(link=["external.ltoir"], output="ptx")(_kernel)
```
gmarkall: I think that this isn't an issue specific to this PR, but looking at this particular example, I do find it dubious that we have an "output=" keyword argument at all. In numba-cuda, if you want some code to be generated for your own use, the [compilation APIs](https://nvidia.github.io/numba-cuda/reference/host.html#compilation) are intended to be the way to go. Or, do I misunderstand the purpose of the `output` kwarg here? (cc @atmnp )

gmarkall (changes requested):
I was going to approve this PR on the basis that I think it's doing what it says it's doing, but I did notice one issue I mentioned on the tests in the diff - we should not be disabling LTO because of the user requesting lineinfo.

Other than that, I was going to approve the PR with the notes that this adds to the patterns of "do something that's not quite right, and then go and try and correct it / throw away the wrong thing later", which is in general not a good architecture and compounded with future changes will balloon complexity further:

- Using `drop_shared_gep_no_wrap_flags` endevaors to remove a flag that ideally should not have been present at the time we generated the GEP, if we knew that the GEP pertained to dynamic shared memory.
- We already have instances where the a new linker for the `MLIRLowering` is created, discarding the previous linker and the work that had been done in its internals. This adds another path where we call `self._create_linker()`, when LTO is opportunistically enabled.

There should be a time at which we tidy these things up - ideally we ought to know prior to creating any linker the complete set of things we're going to link and the options we're going to use, then create the linker and link everything, instead of repeatedly creating linkers, doing half the job with them, then throwing them away until we finish getting everything right.

I don't think the decision of when to do that is for me to make - I would defer to @atmnp on this (and anyone else he suggests 🙂).

tests/test_lto_defaults.py
tpn: Addressed in `20c1713bab12a743f64bdc8900dd8cb6cf136cbc`. LTOIR link inputs now fail when the effective options leave LTO disabled instead of silently keeping `lto=False`: decorator-time coverage includes `link=["external.ltoir"], lto=False`, `cuda.LTOIR(...)` with `lto=False`, `output="ptx"` plus LTOIR, missing nvJitLink, and debug-implied disable; lowering-time discovered LTOIR with explicit `lto=False` / `output="ptx"` also raises.

tests/test_lto_defaults.py
tpn: Addressed in `20c1713bab12a743f64bdc8900dd8cb6cf136cbc`. `lineinfo=True` is no longer treated as a reason to disable LTO. The current tests assert `cuda.jit(link=["external.ltoir"], lineinfo=True)` keeps `targetoptions["lto"] is True`, and lowering-discovered LTOIR plus lineinfo recreates the linker with LTO enabled.

tests/test_lto_defaults.py
tpn: I left the existing `output=` keyword behavior intact in this PR. The changes here are limited to preserving and validating the existing policy: `output="ltoir"` still opts into LTO, `output="ptx"` keeps the PTX path, and contradictory `output="ltoir", lto=False` now raises. I agree the larger API question is separate from #64 and should be handled separately if we decide to revisit it.

tpn:
> I was going to approve this PR on the basis that I think it's doing what it says it's doing, but I did notice one issue I mentioned on the tests in the diff - we should not be disabling LTO because of the user requesting lineinfo.
> 
> Other than that, I was going to approve the PR with the notes that this adds to the patterns of "do something that's not quite right, and then go and try and correct it / throw away the wrong thing later", which is in general not a good architecture and compounded with future changes will balloon complexity further:
> 
> * Using `drop_shared_gep_no_wrap_flags` endevaors to remove a flag that ideally should not have been present at the time we generated the GEP, if we knew that the GEP pertained to dynamic shared memory.
> * We already have instances where the a new linker for the `MLIRLowering` is created, discarding the previous linker and the work that had been done in its internals. This adds another path where we call `self._create_linker()`, when LTO is opportunistically enabled.
> 
> There should be a time at which we tidy these things up - ideally we ought to know prior to creating any linker the complete set of things we're going to link and the options we're going to use, then create the linker and link everything, instead of repeatedly creating linkers, doing half the job with them, then throwing them away until we finish getting everything right.
> 
> I don't think the decision of when to do that is for me to make - I would defer to @atmnp on this (and anyone else he suggests 🙂).

Yeah, agreed in general on all points.  #71 in the current state unblocks a handful of cuda.coop things, so I'd prefer to merge as-is, but happy to create follow-up issues for the following items:

1. Generate dynamic-shared-memory GEPs correctly at lowering time, so no post-pass flag stripping is needed.
2. Refactor MLIR lowering/linking into a planned-link model: discover/record link requirements first, finalize LTO/output/linker policy once, then construct and populate a linker once.
3. Separately clarify whether cuda.jit(output=...) should remain a public-ish escape hatch or move behind compilation APIs.

@atmnp thoughts?

atmnp (approved):
I am on board with merging this as-is.

I agree with the points Graham brought up on why the `output` kwarg on the `jit` decorator might not be needed, and that this is further complicating our already complicated flow of passing linker options. I do think that these other changes are strictly outside of the scope of *this PR* and the issue (#64) it aims to resolve, which we need for unblocking cuda.coop sooner rather than later.

I would like both of your eyes on #151 however to see if that captures what we want here, and the other changes Graham has proposed - I am okay with resolving the merge conflicts on that one after we get this one, I just wanted to make sure at least I personally had a clear vision of the implementation we'd like to see going forward.

I am approving this for now, so we can get the ball rolling but let's look at #151 which will need to get rebased, but I'll leave it as-is until we confirm it covers all the cases/looks good in principle, then I will deal with the impl changes we'd need post the merge of this.

gmarkall:
> Separately clarify whether cuda.jit(output=...) should remain a public-ish escape hatch or move behind compilation APIs.

I don't even think there's anything to move - I think this is just duplicating what already exists in the `cuda.compile()` API, so it can just be removed if there's no other reason for it to exist.

gmarkall:
I've disabled auto-merge because it will result in the PR getting merged as soon as CI passes, because of Atmn's approving review - I am still to go through the last round of changes.

tests/test_lto_defaults.py
```diff
@@ -36,45 +38,90 @@ def test_mlir_jit_enables_implicit_lto_for_external_link_items(monkeypatch):
     assert dispatcher.targetoptions["_lto_explicit"] is False
 
 
+def test_mlir_jit_keeps_cu_link_items_on_ptx_by_default(monkeypatch):
+    _set_nvjitlink_available(monkeypatch, True)
+
+    dispatcher = cuda.jit(link=["external.cu"])(_kernel)
```
gmarkall, line 413: Shouldn't we be using LTO by default if we are linking external items? If we use PTX, then the result is suboptimal because function calls will be required in the final generated SASS.

gmarkall (commented):
I haven't read through all the extra code yet, but I think one of the tests highlights that we're still (or perhaps now, rather than still) too aggressively disabling LTO - whenever multiple items are being linked together, we should be using LTO wherever we can because separate compilation as PTX prevents any opportunity for inlining function calls across translation units and further optimizations that result from that. I've noted on the diff where I think a test is asserting the behaviour I think we don't want.

gmarkall:
> I'd prefer to merge as-is, but happy to create follow-up issues for the following items

I agree here, I think we should not increase the scope of this PR further.

tests/test_lto_defaults.py
tpn: Addressed in the current head: CUDA source link items now take the LTO path by default when nvJitLink is available, including the lowering-time discovered-link path. I also pushed a follow-up fix for the CI failure in the lineinfo + CUDA-source LTO case: `_prepare_llvm_ir()` now re-applies `Debug Info Version` module-flag neutralization after `downgrade_for_libnvvm()`, with a regression test covering that post-downgrade path. Local validation in `cccl313nr` passes `tests/test_lto_defaults.py` and the two previously failing `tests/infra/test_link_libraries.py` cases.

tpn: Follow-up after the first CI rerun: the same failure persisted only on the pre-Blackwell LLVM70 path (`sm_80`/`sm_89`/`sm_90`), which bypasses `_prepare_llvm_ir()`. I pushed `16f2097` so LLVM70 emits `Debug Info Version` with NVRTC-compatible Error behavior (`i32 1`) directly, and added a FileCheck assertion for that generated IR. Local validation now covers the forced `chip="sm_90"` CUDA-source LTO link reproducer as well as the normal focused link tests.

tpn:
@gmarkall @atmnp I think I've addressed everything here, and CI is green.  Graham, if it all looks good, can you resolve your "requesting changes" flag?  Thanks.

atmnp:
@tpn We resolved some of the #155 exposed issues for the v0.4.1 release, are we good to close this?

tpn:
Closing as obsolete, this was superseded by newer work.

PR #74, opened 2026-05-21
Remove `@intrinsic` implementations

gmarkall:
After some experimentation, I think that that `@intrinsic` doesn't yet work in Numba-CUDA-MLIR - none of the non-vendored intrinsics seem to be having any effect either.

This commit removes them, in order to avoid leaving dead / misleading code lying around. The entire `numba_cuda_mlir.lowering.mlir` module is removed as a result, as it only seemed to contain unused code / placeholders.


src/numba_cuda_mlir/lowering/mlir/types.py
```diff
@@ -3,55 +3,17 @@
...
 lower = registry.lower
 
 
-def mlir_convert(ctor):
-    """
-    Create an mlir builder that converts a value indiscriminately to the given type.
-    Note that the construction of the type is delayed until the builder itself
-    is called because we need to be within an MLIR context.
-    """
+lower(bool, types.Any)
```
gmarkall: I just noticed a typo here, this should have been turned into a decorator. If CI still passes, I'll delete the rest of this file because that will show that the remaining lowering also had no effect.

VijayKandiah (approved):
LGTM, thanks!

PR #77, opened 2026-05-22
[Lowering]  Fix multi-assigned heterogeneous tuple lowering

atmnp:
Spill multi-assigned heterogeneous BaseTuple values through per-element stack slots instead of keeping branch-local tuple elements in the varmap. This ensures values loaded after control-flow joins dominate their uses, avoiding MLIR verification failures such as "operand does not dominate this use".


Jlisowskyy (approved):
LGTM! It looks like the issue is resolved.

VijayKandiah (approved):

PR #78, opened 2026-05-26
Skip building `libcudf` from source in third party tests

brandon-b-miller:
The target branch we're testing here only contains python changes relative to main, so we should be able to avoid this piece of the CI job by just pulling a package and building/testing the unmerged code on top of that.


VijayKandiah (approved):
LGTM

PR #80, opened 2026-05-26
[Lowering] Introduce value/storage data model to fix float16/bool memory representation bugs

atmnp:
Separate MLIR value types (f16, bf16, i1) from their byte-addressable storage types (i16, i8) throughout the lowering pipeline. This fixes nvjitlink LTO incorrectly deleting half-precision stores to shared memory and ensures correct ABI representation for bool, float16, bfloat16, and exotic float types across function calls, array operations, and reductions.

Key changes:
- StoragePrimitiveModel with distinct value/data/argument/return types
- Automatic value<->storage conversion in load_var/store_var and at ABI boundaries
- Pre-codegen pass to use integer stores for shared-memory half types
- Shared-memory address space cast in ffi.from_buffer


src/numba_cuda_mlir/mlir_lowering.py
VijayKandiah: This allocates stack slots using `get_mlir_type` (value type, e.g. f16, i1) rather than `get_storage_type` (e.g. i16, i8). The direct scalar and UniTuple multi-assign paths at lines 731 and 742 in this file were updated to use `get_storage_type`, but this recursive helper for heterogeneous BaseTuple was not. Should this be updated too?

VijayKandiah, line 1882: `lower_overload_call` stores call results without `from_return`. Since `StoragePrimitiveModel.get_return_type() `returns the storage type (e.g. i16 for float16, i8 for bool), a callee that returns one of these types would produce a storage-typed value that gets placed directly into the varmap — downstream code expecting the value type (f16, i1) would see a type mismatch.

`build_user_defined_function_call` and `build_recursive_call` above both apply `from_return` to convert back to value types in your changes. Should this path do the same?

src/numba_cuda_mlir/models.py
VijayKandiah: The dynamic FP8 models registered here use plain `PrimitiveModel` (value == storage), while the static FP8 models earlier in the file (e.g. `Float8E5M2TypeModel` at line 681) use `StoragePrimitiveModel` with i8 storage. If both registration paths can be active for overlapping type instances, the data model would differ depending on init order -- `get_data_type()` would return `f8E5M2` from one but i8 from the other.

Should these also use `StoragePrimitiveModel(dmm, fe_type, T.f8E5M2(), T.i8())` etc. for consistency?

src/numba_cuda_mlir/lowering/builtins.py
```diff
@@ -337,39 +346,46 @@ def lower_broadcasted_floor_division(builder, target, args, kwargs):
...
+    lhs_type = builder.get_numba_type(lhs_arg.name)
+    rhs_type = builder.get_numba_type(rhs_arg.name)
+    lhs = builder.load_var(lhs_arg)
+    rhs = builder.load_var(rhs_arg)
+    if isinstance(lhs_type, types.Array):
+        lhs = memref_to_value_tensor(lhs_type, lhs)
+    if isinstance(rhs_type, types.Array):
+        rhs = memref_to_value_tensor(rhs_type, rhs)
+
     lhs, rhs = broadcast_shapes_for_binary_op(lhs, rhs, builder)
```
VijayKandiah, line 366: `lower_broadcasted_binary` (line 281) and `lower_broadcasted_div` (line 414) both call `expensive_coerce_tensor_type` after broadcasting to ensure operand element types match the result type. The floor division path skips that step. Is this intentional?

VijayKandiah (approved):
LGTM overall, added some minor concerns.

Jlisowskyy:
Looks like there's a small regression compared to main. This code works fine there, but fails on this branch.
```python
import numpy as np
from numba_cuda_mlir._mlir.dialects import llvm
from numba_cuda_mlir._mlir.extras import types as T
from numba_cuda_mlir.extending import lowering_registry, overload_attribute, typing_registry
from numba_cuda_mlir.lowering_utilities import to_mlir_type, unverified_convert
from numba_cuda_mlir.models import PrimitiveModel, register_model
from numba_cuda_mlir.numba_cuda import types
from numba_cuda_mlir.numba_cuda.extending import typeof_impl


class Dummy:
    pass


class DummyType(types.Type):
    def __init__(self) -> None:
        super().__init__(name="Dummy()")


@typeof_impl.register(Dummy)
def _typeof(val, c):
    return DummyType()


class _Model(PrimitiveModel):
    def __init__(self, dmm, fe_type) -> None:
        super().__init__(dmm, fe_type, T.i8())


register_model(DummyType)(_Model)


@overload_attribute(DummyType, "arrangement", inline="always",
                    typing_registry=typing_registry, lowering_registry=lowering_registry)
def _ol_str(api):
    return lambda api: "col_major"


_DUMMY = Dummy()


@cuda.jit
def k(out):
    out[0] = 1 if _DUMMY.arrangement == "col_major" else 0


k[1, 1](cuda.to_device(np.zeros(1, dtype=np.int32)))
cuda.synchronize()
```

src/numba_cuda_mlir/numba_cuda/cudadrv/driver.py
```diff
@@ -2416,8 +2417,10 @@ def _get_linker_options(self, ptx):
             prec_sqrt=self._prec_sqrt,
             fma=self._fma,
             variables_used=self.variables_used,
-            optimize_unused_variables=self._optimize_unused_variables,
-            optimization_level=self._optimization_level,
+            optimize_unused_variables=(
+                self._optimize_unused_variables if self.variables_used else None
+            ),
+            optimization_level=opt_level,
```
Jlisowskyy, line 2423: Is this change intended? Correct me if I'm wrong, it looks like `optimization_level=0` here disables all optimization on LTOIR (e.g. cusolverdx), it gets linked but not optimized (no cross-module inlining). If that's the case, is there a way we can work around it?

atmnp: Yes, it is intended. `numba-cuda` does not set the opt level to anything higher than 0 either, precisely to avoid the bug we were looking at earlier in cufft/cublas. We should fix this in a follow up where O3 in the linker options does *not* DCE out shared memory. There are work arounds that I partially explored, but it took away too much from this PR for the value vs storage model distinction.

Jlisowskyy: Thanks for clarification!

Jlisowskyy (approved):
Looks good, Thanks!

src/numba_cuda_mlir/mlir_lowering.py
```diff
@@ -3493,6 +3493,8 @@ def get_argument_type(self, ty):
         return self._lookup_model(ty).get_argument_type()
 
     def get_return_type(self, ty):
+        if isinstance(ty, types.DTypeSpec):
+            return ir.NoneType.get()
```
ZzEeKkAa, line 3497: Is it some what a place holder?

ZzEeKkAa (approved):
Functional wise works perfectly!

src/numba_cuda_mlir/mlir_lowering.py
atmnp: This was my fix for `VectorTypeClass` problem I believe @Jlisowskyy was referring to. Rather than having a data model registered for this, this is *a way* of making sure we don't try to lower it to MLIR.

ZzEeKkAa: I see. Should work then. It's just I did not expect this way. I thought we have to do something like `@register_model(VectorTypeClass)` with a dummy struct (that how we usually do it for nvmath).

PR #48, opened 2026-05-14
Windows LLVM 7 and latest support

tpn:
## Summary

- Build and cache both modern LLVM/MLIR and LLVM 7 in CI.
- Package modern MLIR Python bindings plus the LLVM 7 compatibility bridge into Windows wheels.
- Add explicit source-build selection via `NUMBA_CUDA_MLIR_BUILD_LLVM70`:
  - `auto`: build LLVM 7 bridge when `MLIR_DIR` is set
  - `1` / `on`: require and build the LLVM 7 bridge
  - `0` / `off`: build latest LLVM/MLIR support only
- Keep direct CMake selection through `-DBUILD_LLVM70=ON/OFF`.
- Fail the build if LLVM 7 bridge support is requested but `MLIRToLLVM70` is not produced.
- Remove stale LLVM 7 artifacts from reused build directories for latest-only builds.
- Update install docs and runtime diagnostics for the new selectable build modes.

## Testing

- `ci/build-windows.sh modern`
- `ci/build-windows.sh llvm7`
- `cmd /c .ci-verify\run-wheel-latest-only.cmd`
- `cmd /c .ci-verify\run-wheel-both-explicit.cmd`
- Verified latest-only wheel has `MLIRToLLVM70` count `0` and `MLIRPythonCAPI.dll` count `1`
- Verified explicit-both wheel has `MLIRToLLVM70` count `2` and `MLIRPythonCAPI.dll` count `1`
- Smoke-tested installed latest-only wheel: import modern MLIR and parse `module {}`
- Smoke-tested installed explicit-both wheel: import modern MLIR, load `MLIRToLLVM70.dll`, and parse `module {}`
- `git diff --check`
- `bash -n ci/build-llvm-modern.sh ci/build-windows.sh ci/build-llvm7.sh`
- `.ci-verify\py311\python.exe -m ruff check setup.py src\numba_cuda_mlir\tools.py`
- `.ci-verify\py311\python.exe -m ruff format --check setup.py src\numba_cuda_mlir\tools.py`
- `.ci-verify\py311\python.exe -m py_compile setup.py src\numba_cuda_mlir\tools.py`

PR #81, opened 2026-05-26
[CI] Introduce Windows CI

atmnp:
This PR adds end-to-end Windows support (build + wheel + GPU test CI) and introduces a new Windows-safe MLIR translation path, while keeping Linux behavior intact. It also refactors LLVM artifact handling/caching, updates packaging/runtime loading for Windows DLLs, and adjusts tests for cross-platform reliability.

Conceptual changes:
  1. CI now includes Windows LLVM builds, wheel builds, and wheel tests.
  2. New Windows test workflow (test-wheel-windows.yml) runs GPU tests in a Windows devcontainer, including driver install/mode
     setup (WDDM/MCDM), artifact restore, and matrix-driven execution.
  3. Main CI orchestration (ci.yaml) adds Windows build/test jobs and gates final status on Windows test outcomes.
  4. LLVM artifact handling is unified via a new composite action (restore_llvm_artifacts) that restores from cache when possible and falls back to workflow artifacts, with validation checks for expected binaries.
  5. LLVM cache key generation is centralized (ci/llvm-cache-key.sh) and now includes Windows-specific build inputs to avoid stale/incompatible cache reuse.
  6. LLVM build workflows were refactored to use explicit cache restore/save steps and always upload LLVM artifacts after restore/build.
  7. A full Windows LLVM build pipeline was added (ci/build-windows.sh + PowerShell wrappers), including:
      - building LLVM7 and modern LLVM in Windows containers,
      - generating LLVM-C.dll exports from LLVM static libs (gen-llvm-c-exports.py),
      - staging MLIR Python artifacts,
      - building/staging a new modern bridge DLL,
      - smoke-import validation of native MLIR artifacts.
  8. CTK fetch action was made cross-platform:
      - robust component filtering by CUDA major version,
      - support for Windows package layout/zip extraction,
      - Windows CUDA env/path propagation.
  9. Native C/C++ build system was made cross-platform:
      - MSVC warning/error flags and compatibility fixes,
      - reusable helper for Python extension module targets,
      - platform-safe module init/type object initialization,
      - Windows dynamic loading (LoadLibrary/GetProcAddress) for CUDA and MLIR CAPI paths,
      - Windows-safe aligned allocation and minor portability cleanups.
  10. A new modern MLIR bridge library (cext/mlir-modern) was introduced:
      - translates MLIR GPU module text to libnvvm-compatible LLVM IR/bitcode,
      - applies intrinsic/metadata/attribute adaptations for libnvvm compatibility,
      - includes smoke tests.
  11. Python runtime/optimization path now has OS-aware behavior:
      - Linux keeps pointer-based MLIR->LLVM flow,
      - Windows uses the new modern bridge translation path,
      - optional debug-info-preserving text IR emission path was added.
  12. Packaging/build staging was updated for Windows:
      - package data now includes .dll/.pyd,
      - setup.py uses platform-specific shared-library names,
      - staged MLIR package is patched to register DLL directories on Windows,
      - LLVM70 bridge and LLVM runtime library staging supports Windows naming/layout.
  13. Test infrastructure and tests were updated for cross-platform stability:
      - filecheck logic moved to in-process parser/matcher (no external FileCheck dependency) - more similar to Numba-CUDA in behavior
      - helpers added for subprocess-based output capture,
      - printf and AST consteval/device-func tests moved to subprocess execution,
      - Windows-specific skips/xfails added where needed (bfloat16, selected LLVM70 vector-path cases).

Additional functional fixes included:
  - Safe tuple-iterator indexing in lowering utilities to avoid invalid accesses.
  - typeconv overload selection now validates/guards sequence sizes and integer conversions.
  - Debug info string escaping/normalization improved (notably Windows path handling).
  - run-tests now includes Windows wheel install path handling and explicit native MLIR import diagnostics.

Known limitations:
  - bfloat16 tests are currently skipped on Windows. Our lowering for bf16 is currently lowering through MLIR which *does not* use the correct libdevice calls, we support it via `truncf`/`extf` which has flaky behavior on Windows.
  - Some vector-type tests are expected-fail only on the Windows LLVM70 path (pre-SM100 path).
  - Windows TCC Driver Mode is untested

Overall impact
  - Adds first-class Windows CI and packaging support.
  - Preserves Linux flows while consolidating artifact/caching logic.
  - Introduces a new bridge-based translation strategy needed for reliable Windows MLIR/libnvvm interop. Without this, we introduce incompatibilities between the translation tool and the LLVMs we build.

Notes:
- We need the devcontainer for building LLVM to get a copy of MSVC, we likely do not need it at test-time but it's kept that way for now to maintain uniformity in the environment.
- Thanks to @tpn for his initial PR #48 which helped confirm which of the intermediate steps in the PR were necessary
- The Windows LLVM build/cext process might be more complex than necessary, but I would prioritize it as a clean up for our Windows platform users.
- cuDF and numbast are not supported on Windows, so they can not be added


cext/mlir-modern/ModernBridge.cpp
```diff
@@ -0,0 +1,624 @@
...
+static void adapt_for_libnvvm(llvm::Module &mod, llvm::LLVMContext &ctx) {
+    adapt_barrier_sync(mod, ctx);
+    adapt_barrier_reduction(mod, ctx);
+    adapt_inline_asm_intrinsics(mod, ctx);
+    adapt_atomicrmw(mod, ctx);
+    adapt_trunc(mod, ctx);
+    adapt_nvvm_annotations(mod, ctx);
+    adapt_nvvmir_version(mod, ctx);
+    adapt_debug_info_version(mod, ctx);
+}
```
VijayKandiah, line 455: Both `adapt_for_libnvvm` and `downgrade_for_libnvvm` here duplicate the logic in `cext/launcher/llvm_downgrade.cpp` (which uses the LLVM C API dynamically loaded via dlsym). The duplication seems necessary given the different API surfaces (C++ here vs C there), but it does incur a maintenance burden to keep both in sync. Would it be feasible to share the adaptation logic somewhere — e.g., via a common file with platform/API `#ifdef _WIN32`s, or a shared header with the transform logic templated over the API?

cext/mlir-llvm70/include/llvm70/CAPILoader.h
```diff
@@ -26,32 +31,69 @@ class CAPILoader {
 
   static llvm::Expected<std::unique_ptr<CAPILoader>>
   create(llvm::StringRef libPath) {
+#ifdef _WIN32
+    HMODULE handle = nullptr;
+    std::string path = libPath.str();
+    if (path.empty()) {
+      GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS,
+                         reinterpret_cast<LPCSTR>(&CAPILoader::create), &handle);
+      path = "<current module>";
```
VijayKandiah, line 43: Should we set `ownsHandle=false` for this branch instead of setting it always to be `true` below at line 123? Otherwise, the destructor will call `FreeLibrary` on a handle whose reference count was never incremented, which could cause undefined behavior. Note that this branch might not be hit today at all because all callers would be passing real paths and `path.empty()` never returns true.

src/numba_cuda_mlir/testing/__init__.py
```diff
@@ -57,57 +87,40 @@ def filecheck_with_comments(module):
...
+    matcher = Matcher(opts, input_file, parser)
+    matcher.stderr = StringIO()
+    try:
+        result = matcher.run()
+    except AssertionError as e:
+        err = matcher.stderr.getvalue()
+        raise ValueError(f"\n{err}") from e
+    if result != 0:
+        err = matcher.stderr.getvalue()
+        raise ValueError(f"\n{err}")
```
VijayKandiah: `get_filecheck_path` is no longer called after the switch to in-process Python `filecheck` library. Should we remove it?

.github/actions/fetch_ctk/action.yml
```diff
@@ -30,27 +30,29 @@ runs:
...
-        echo "CTK_CACHE_KEY=mini-ctk-${{ inputs.cuda-version }}-${{ inputs.host-platform }}-$HASH" >> $GITHUB_ENV
-        echo "CTK_CACHE_FILENAME=mini-ctk-${{ inputs.cuda-version }}-${{ inputs.host-platform }}-$HASH.tar.gz" >> $GITHUB_ENV
-        echo "CTK_CACHE_COMPONENTS=${CTK_CACHE_COMPONENTS}" >> $GITHUB_ENV
+        HASH=$(printf '%s' "${CTK_CACHE_COMPONENTS}" | sha256sum | awk '{print $1}')
+        echo "CTK_CACHE_KEY=mini-ctk-${{ inputs.cuda-version }}-${{ inputs.host-platform }}-$HASH" >> "$GITHUB_ENV"
+        echo "CTK_CACHE_FILENAME=mini-ctk-${{ inputs.cuda-version }}-${{ inputs.host-platform }}-$HASH.tar.gz" >> "$GITHUB_ENV"
+        echo "CTK_CACHE_COMPONENTS=${CTK_CACHE_COMPONENTS}" >> "$GITHUB_ENV"
 
     - name: Install dependencies
+      if: ${{ startsWith(inputs.host-platform, 'linux') }}
```
VijayKandiah, line 55: `install_unix_deps` is ran for linux only and skipped on Windows, but the cache-miss download path below (line 117-118) unconditionally uses `jq` to parse the CTK redistrib JSON. If a `LOCAL_CTK=1` Windows job hits a cold GHA cache, will `jq` be available on the windows runner?

tests/numba_cuda_tests/cudapy/test_bfloat16.py
```diff
@@ -17,6 +18,10 @@
     uint64,
 )
 
+pytestmark = pytest.mark.skipif(
+    sys.platform == "win32", reason="bfloat16 tests are disabled on Windows"
```
VijayKandiah: Should we also mention in the comment about why these are skipped on windows here?

`bfloat16 tests are disabled on windows due to current lowering path via truncf/extf having flaky behavior on Windows.`

tests/numba_cuda_tests/cudapy/test_vector_type.py
```diff
@@ -32,6 +36,31 @@ def _vector_types():
...
+
+def _xfail_windows_llvm70_vector_path(reason):
+    def decorator(fn):
+        @wraps(fn)
+        def wrapper(*args, **kwargs):
+            if not _is_windows_llvm70_vector_path():
+                return fn(*args, **kwargs)
+            try:
+                fn(*args, **kwargs)
+            except Exception:
```
VijayKandiah: Catching the bare `Exception` here could mask unrelated failures (like `ImportError`, `TypeError`) as expected xfails. What should be the expected xfail here? We should ideally narrow the capture scope to the expected failure.

VijayKandiah (changes requested):

cext/mlir-modern/ModernBridge.cpp
atmnp: To reduce the maintenance burden without that larger refactor, I added `cext/libnvvm_compat_passes.def` as the shared ordered pass list. Both `ModernBridge.cpp` and `llvm_downgrade.cpp` now include that file when running libnvvm compatibility passes, so future pass additions/removals have one ordering source even though each API keeps its own implementation body.

cext/mlir-llvm70/include/llvm70/CAPILoader.h
atmnp: We now explicitly avoid incrementing the refcount and set ownsHandle = false; the destructor only calls FreeLibrary when ownsHandle is true.

VijayKandiah (approved):
LGTM but i'd wait for @leofang to take a look before merging.

ci/tools/env-vars
```diff
@@ -20,17 +20,24 @@ PYTHON_VERSION_FORMATTED=$(echo "${PY_VER}" | tr -d '.')
 if [[ "${HOST_PLATFORM}" == linux* ]]; then
   REPO_DIR=$(pwd)
   TOOLS_PATH="${REPO_DIR}/ci/tools"
+elif [[ "${HOST_PLATFORM}" == win* ]]; then
+  REPO_DIR=$(pwd)
+  TOOLS_PATH="${REPO_DIR}/ci/tools"
```
leofang: nit: this can be merged with the above block and set unconditionally

ci/tools/install_gpu_driver.ps1
leofang, line 1: FYI I'll handle this in a separate PR to make driver versions customizable

ci/tools/run-tests
```diff
@@ -30,16 +34,79 @@ if [[ "${TEST_CUDA_MAJOR}.${TEST_CUDA_MINOR}" == "12.2" ]]; then
...
     "cuda-toolkit[cccl,cudart]==${TEST_CUDA_MAJOR}.${TEST_CUDA_MINOR}.*" \
     "${coverage_packages[@]}"
 fi
 
+if [[ "${RUNNER_OS:-}" == "Windows" || "$(python - <<'PY'
+import os
+print(os.name)
+PY
+)" == "nt" ]]; then
+  python - <<'PY'
```
leofang: Q: Why do we need this, instead of just letting the same tests below run (which would cover the import test)?

ci/setup-sccache.sh
```diff
@@ -25,10 +25,6 @@ cat <<EOF > ~/.config/sccache/config
 bucket = "rapids-sccache-devs"
 region = "us-east-2"
 key_prefix = "numba_cuda_mlir"
-
-[cache.s3.preprocessor_cache_mode]
-use_preprocessor_cache_mode = true
-key_prefix = "numba_cuda_mlir/preprocessor"
```
leofang, line 31: Q: Why do we remove these? I think it should work for Windows too?

.github/actions/fetch_ctk/action.yml
leofang, line 1: Note: I'll propagate recent updates to this action in a separate PR.

leofang (approved):
Approve — thanks for getting Windows CI on the board, this is a substantial lift.

Inline comments capture the issues / follow-ups I noticed. Nothing blocks merging; nits, follow-up refactors, and one backport candidate to cuda-python (`install_gpu_driver.ps1` error handling).

Two positive notes:
- `ci/llvm-cache-key.sh` is a clean centralized cache-key generator — good pattern to keep consistent if we add similar caches elsewhere.
- `.github/actions/restore_llvm_artifacts/action.yml` is well-designed: SHA-pinned actions, restore-then-fallback-to-artifact pattern, and post-restore validation.

The main follow-up worth flagging is the devcontainer at test stage — per the PR description it isn't strictly required there, and dropping it in a future PR would simplify the workflow and bring cuSIMT's Windows test job closer to cuda-python's structure.

.github/workflows/build-wheel.yml
```diff
@@ -82,43 +83,52 @@ jobs:
...
+          host-platform: ${{ inputs.host-platform }}
+          llvm-py-tag: ${{ steps.wheel-tags.outputs.llvm-py-tag }}
 
-      - name: Download LLVM 7 artifacts
-        if: ${{ startsWith(inputs.host-platform, 'linux') }}
-        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
-        with:
-          name: llvm7-install-${{ inputs.host-platform }}
-          path: llvm7-install
+      - name: Validate Windows MLIR Python ABI
```
leofang: Non-blocking: this step duplicates the validation already performed by `.github/actions/restore_llvm_artifacts/action.yml`, which checks the same `_mlir.${py-tag}-win_amd64.pyd` and prints the same `ls` on failure. Suggest dropping this inline step and letting the composite action own the validation — that way any other caller of `restore_llvm_artifacts` benefits too.

.github/workflows/test-wheel-windows.yml
```diff
@@ -0,0 +1,200 @@
...
+      fail-fast: false
+      max-parallel: 5
+      matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
+    runs-on: "windows-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.DRIVER }}-${{ matrix.GPU_COUNT }}"
+    if: ${{ github.repository_owner == 'nvidia' && !cancelled() }}
+    steps:
+      - name: Checkout ${{ github.event.repository.name }}
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+      - name: Update driver
```
leofang, line 77: Non-blocking: `setup-proxy-cache` is missing from this test job. cuda-python's `test-wheel-windows.yml` uses it (`enable-apt: true`) right around here. Even with tests running inside a devcontainer, host-side proxy still helps with `docker pull` and `actions/download-artifact`. Add it, or add a one-line comment explaining why it's intentionally omitted.

.github/workflows/test-wheel-windows.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: Set up mini CTK
+        if: ${{ matrix.LOCAL_CTK == '1' }}
+        uses: ./.github/actions/fetch_ctk
+        continue-on-error: false
+        with:
+          host-platform: ${{ inputs.host-platform }}
+          cuda-version: ${{ matrix.CUDA_VER }}
+          cuda-components: "cuda_nvcc,cuda_cudart,cuda_crt,libnvvm,cuda_nvrtc,cuda_cccl,libnvjitlink,cuda_cuobjdump"
+
+      - name: Resolve devcontainer image
```
leofang, line 143: Non-blocking: the test stage runs inside a devcontainer (this step + `Pull devcontainer image` + `Run numba-cuda-mlir tests in devcontainer` below). The PR description already notes this isn't strictly required at test time and was kept for uniformity. Suggest dropping the devcontainer at test stage in a follow-up PR — would simplify this workflow, align cuSIMT's Windows test job with cuda-python's host-runner approach, and unblock host-side `setup-proxy-cache`.

ci/test-matrix.yml
```diff
@@ -80,3 +80,26 @@ linux:
     - { ARCH: 'arm64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'arm64', PY_VER: '3.12',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
   nightly: []
+
+windows:
+  pull-request:
+    # win-64
+    # - { ARCH: 'amd64', PY_VER: '3.10',  CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'rtx2080',    GPU_COUNT: '1', DRIVER: 'latest', DRIVER_MODE: 'WDDM' }
```
leofang: Non-blocking: eight Windows matrix entries are commented out here (TCC mode + several CUDA/Python combos, per the PR's "Windows TCC Driver Mode is untested" note). Commented-out lines tend to rot — either delete them with a single-line comment pointing to a tracking issue, or open the issue for the TCC limitation and reference it from this file.

.github/workflows/build-llvm.yml
```diff
@@ -101,6 +96,104 @@ jobs:
...
+          path: llvm-modern-install/
+          key: ${{ steps.cache-key.outputs.key }}
+
+      - name: Upload LLVM Modern artifacts
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+        with:
+          name: llvm-modern-install-${{ matrix.py-tag }}-${{ inputs.host-platform }}
+          path: llvm-modern-install/
+
+  build-llvm-modern-windows:
```
leofang: Non-blocking follow-up: this `build-llvm-modern-windows` job and `build-llvm7-windows` below duplicate ~90% of the corresponding Linux jobs (cache-key, restore, build, save, upload). Refactoring to a matrix or per-platform composite action would roughly halve this file. Worth a separate PR.

.github/workflows/build-llvm.yml
```diff
@@ -101,6 +96,104 @@ jobs:
...
+
+      - name: Pull devcontainer image
+        if: steps.gha-cache.outputs.cache-hit != 'true'
+        shell: pwsh
+        run: |
+          . ./ci/windows-common.ps1
+          Import-EnvFile ci/windows-devcontainer.env
+          docker pull $env:WINDOWS_DEVCONTAINER_IMAGE
+
+      - name: Build modern LLVM in devcontainer
```
leofang: Non-blocking follow-up: the Windows path is missing tiers 2 and 3 of our LLVM cache hierarchy. The Linux `build-llvm-modern` job above (lines 72-92) installs sccache, configures S3 credentials via `ci/setup-sccache.sh`, and runs `cache_download` / `cache_upload` via `ci/llvm-cache.sh` — so on GHA-cache miss it falls back to the S3 artifact cache, and on full miss the actual build is sccache-accelerated. This Windows job has neither. `ci/build-windows.sh` has zero sccache references; the `docker run` only propagates `--env PARALLEL`, so AWS creds wouldn't reach an sccache call even if added. Result: any GHA-cache key invalidation forces a full ~40min from-scratch build per matrix entry. Same gap applies to `build-llvm7-windows`. Wiring up tier 2 (S3 LLVM artifact cache around `docker run`) is the cheaper fix; tier 3 (sccache inside the container with `--env AWS_*` and `-DCMAKE_C/CXX_COMPILER_LAUNCHER=sccache`) is the bigger lift but pays off on every key change.

.github/workflows/build-llvm.yml
```diff
@@ -101,6 +96,104 @@ jobs:
...
+          echo "key=${KEY}" >> "$GITHUB_OUTPUT"
+
+      - name: Restore or check LLVM Modern GHA cache
+        id: gha-cache
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
+        with:
+          path: llvm-modern-install/
+          key: ${{ steps.cache-key.outputs.key }}
+
+      - name: Pull devcontainer image
```
leofang, line 159: Non-blocking follow-up: `docker pull` of the rapidsai/devcontainers image takes ~8 min per job on cache miss. With 5 Python matrix entries that's ~40 min of compute spent re-pulling the same image. Options: pre-pull on the runner VM image, mirror to a closer registry, or `docker save | docker load` from a GHA cache (image is the same across all 5 entries). Mostly a runner-side optimization.

ci/windows-llvm-container-build.ps1
```diff
@@ -0,0 +1,68 @@
...
+
+$bash = Resolve-Bash
+$repoRootUnix = Convert-ToUnixPath -WinPath $repoRoot
+& git config --global --add safe.directory $repoRootUnix
+
+$parallel = if ($env:PARALLEL) { $env:PARALLEL } else { '16' }
+$pythonForBash = ($pythonExe -replace '\\', '/')
+$modeArg = $Mode
+
+$bashCmd = @"
```
leofang: Optional polish: this multi-line bash heredoc embedded inside a PS1 script is hard to debug if it ever fails. Consider extracting the bash body to a small `ci/build-windows-driver.sh` and invoking via `bash -lc ci/build-windows-driver.sh`. Not blocking.

ci/tools/install_gpu_driver.ps1
```diff
@@ -0,0 +1,87 @@
...
+    if ($data_center_gpus -contains $gpu_type) {
+        Write-Output "Data center GPU detected: $gpu_type"
+        $filename = "$version-data-center-tesla-desktop-winserver-2022-2025-dch-international.exe"
+        $server_path = "tesla/$version"
+    } elseif ($desktop_gpus -contains $gpu_type) {
+        Write-Output "Desktop GPU detected: $gpu_type"
+        $filename = "$version-desktop-win10-win11-64bit-international-dch-whql.exe"
+        $server_path = "Windows/$version"
+    } else {
+        throw "Unsupported GPU type '$gpu_type'. Known data center GPUs: $($data_center_gpus -join ', '); known desktop GPUs: $($desktop_gpus -join ', ')"
```
leofang, line 28: Backport candidate to cuda-python: this `throw` with a list of known GPUs (plus `$ErrorActionPreference = 'Stop'` at the top of the file, and the matching `throw` for unknown driver modes at line 72) is strictly better than cuda-python's current `Write-Output "Unknown ..."; exit 1`. Could fold into the planned "make driver versions customizable" PR or land separately.

leofang (commented):
Re-positioning my earlier comments to land at the specific steps/lines they refer to (the previous batch all anchored at line 1 by mistake), plus two new follow-up notes on the LLVM Modern Windows build timing: the multi-tier LLVM cache (S3 artifact + sccache) wired up on Linux is missing on the Windows path, and the ~8 min devcontainer image pull repeats per matrix entry. All non-blocking — PR remains approved.

.github/workflows/build-wheel.yml
```diff
@@ -82,43 +83,52 @@ jobs:
           SHA: ${{ github.sha }}
         run: ./ci/tools/env-vars build
 
-      - name: Dump environment
```
leofang, line 85: Nit: this `Dump environment` step was removed in the PR. It's a useful diagnostic for CI failures (env-var content is one of the first things one checks). Consider restoring it.

.github/workflows/build-wheel.yml
```diff
@@ -82,43 +83,52 @@ jobs:
...
-        run: |
-          env
-
       - name: Determine cibuildwheel and LLVM Python tags
         id: wheel-tags
-        if: ${{ startsWith(inputs.host-platform, 'linux') }}
         run: |
```
leofang, line 92: Correction to my earlier comment (I was wrong about `CUDA_VER` being removed — it was never in this build job's env block, only in the test job). Actual finding: `inputs.cuda-version` is declared at line 11 and passed by `ci.yaml`, but it isn't consumed anywhere in this build job (the `Set environment variables` step at line 79 above doesn't include `CUDA_VER`, and `ci/tools/env-vars build` doesn't read it). Per Leo, the input should stay for parity with numba-cuda's build-then-move-tests flow — addressable in a follow-up to actually consume it; flagging so it's not lost.

leofang (commented):
Fixing two earlier comments on `build-wheel.yml`: the `Dump environment` one now anchors on the removed (LEFT) line, and the `CUDA_VER` one — whose premise was wrong (I conflated build-wheel.yml with test-wheel-windows.yml's env block) — is corrected.

.github/workflows/test-wheel-windows.yml
atmnp: added

.github/workflows/build-llvm.yml
atmnp: saving this for a follow up - unclear which solution is optimal

.github/workflows/build-llvm.yml
atmnp: Agreed, this is worth a follow-up. I’m not wiring S3 LLVM artifact caching or in-container sccache in this PR to keep the ball rolling.

PR #82, opened 2026-05-27
Add cuda-gdb CI workflow

jiel-nv:
This PR integrates `cuda-gdb` into the CI pipeline to verify that `numba-cuda-mlir` emits correct DWARF debug information that a debugger can actually digest under a real debugging session.

- **`tests/test_cuda_gdb.py`** — smoke test that compiles a `debug=True, opt=False` kernel, runs it under `cuda-gdb` in batch mode, hits the breakpoint and inspects arguments and local variables to verify they have the correct values.
- **`ci/tools/run-cuda-gdb-tests`** — CI runner locates `cuda-gdb` via `$CUDA_PATH`, installs the `numba-cuda-mlir` wheel from the build artifact, and invokes `pytest`.
- **`.github/workflows/test-cuda-gdb-linux.yml`** — reusable workflow that fetches the necessary CTK components (`cuda_gdb`, `cuda_cuobjdump`, `cuda_nvdisasm`), and runs the tests in a container with `--security-opt seccomp=unconfined` (required for `ptrace`). `LOCAL_CTK: 1` is always set because `cuda-gdb` is not distributed as a PyPI wheel and comes from the CTK.
- **`ci/test-matrix.yml`** — new `cuda-gdb` section targeting L4/amd64 and A100/arm64 on CUDA 13.2.1.
- **`.github/workflows/ci.yaml`** — two new jobs (`test-cuda-gdb-linux-64`, `test-cuda-gdb-linux-aarch64`) wired into the `checks` gate.


.github/workflows/test-cuda-gdb-linux.yml
leofang, line 1: **Recommend deleting this whole file.** Folding cuda-gdb directly into `test-wheel-linux.yml` is cleaner and was the original intent — the commented-out compute-sanitizer scaffolding inside `test-wheel-linux.yml` shows the pattern, and cuda-python does this for real today:

https://github.com/NVIDIA/cuda-python/blob/788bbad1ae8a901241e978d15edda3ee954ed55c/.github/workflows/test-wheel-linux.yml#L289-L299

The shape we want here is a single new optional input on `test-wheel-linux.yml`:

```yaml
inputs:
  extra-cuda-components:
    type: string
    default: ""
...
- name: Set up extra CTK components
  if: ${{ inputs.extra-cuda-components != '' }}
  uses: ./.github/actions/fetch_ctk
  with:
    host-platform: ${{ inputs.host-platform }}
    cuda-version: ${{ matrix.CUDA_VER }}
    cuda-components: ${{ inputs.extra-cuda-components }}
```

**Critical:** the overlay `fetch_ctk` call passes *only* the extra components (`cuda_gdb,cuda_nvdisasm`), **not** the full CTK list. The first `fetch_ctk` keeps its existing component list and cache key, so it stays shared across all test variants. Only the small overlay is cached separately under its own key. This is what keeps total cache footprint small and miss cost low when more variants (compute-sanitizer, etc.) get added on top later — the cuda-python sanitizer overlay is the precedent.

With that single input, plus the existing `test-script` and `matrix_filter` inputs, this entire 138-line workflow file goes away.

.github/workflows/ci.yaml
```diff
@@ -215,6 +215,28 @@ jobs:
     with:
       host-platform: linux-aarch64
 
+  test-cuda-gdb-linux-64:
```
leofang, line 263: With the `extra-cuda-components` input added to `test-wheel-linux.yml` (see comment on the new workflow file), these two jobs collapse to the shape below — note this exactly mirrors the commented-out `coverage-report` job further down in this same file:

```yaml
test-cuda-gdb-linux-64:
  name: cuda-gdb Test linux-64
  if: ${{ github.repository_owner == 'nvidia' }}
  permissions:
    contents: read
  needs:
    - build-linux-64
  uses: ./.github/workflows/test-wheel-linux.yml
  with:
    build-type: pull-request
    host-platform: linux-64
    test-script: run-cuda-gdb-tests
    extra-cuda-components: "cuda_gdb,cuda_nvdisasm"
    matrix_filter: 'map(select(.PY_VER == "3.12" and .CUDA_VER == "13.2.1" and .LOCAL_CTK == "1" and .GPU == "l4")) | .[0:1]'

test-cuda-gdb-linux-aarch64:
  name: cuda-gdb Test linux-aarch64
  if: ${{ github.repository_owner == 'nvidia' }}
  permissions:
    contents: read
  needs:
    - build-linux-aarch64
  uses: ./.github/workflows/test-wheel-linux.yml
  with:
    build-type: pull-request
    host-platform: linux-aarch64
    test-script: run-cuda-gdb-tests
    extra-cuda-components: "cuda_gdb,cuda_nvdisasm"
    matrix_filter: 'map(select(.PY_VER == "3.12" and .CUDA_VER == "13.2.1" and .LOCAL_CTK == "1" and .GPU == "a100")) | .[0:1]'
```

The `checks` job wiring below stays as-is — these two job names don't change.

ci/test-matrix.yml
```diff
@@ -79,4 +79,7 @@ linux:
     - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'arm64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'arm64', PY_VER: '3.12',  CUDA_VER: '13.2.1', LOCAL_CTK: '1', GPU: 'a100',       GPU_COUNT: '1', DRIVER: 'latest' }
+  cuda-gdb:
```
leofang: With the `matrix_filter` approach in `ci.yaml`, this `cuda-gdb:` section isn't needed at all — the two desired rows (py3.12 / CUDA 13.2.1 / LOCAL_CTK=1, `l4` amd64 and `a100` arm64) already exist verbatim in the `pull-request:` block above. Selecting them via `matrix_filter` keeps the matrix source of truth in one place and avoids drift between the cuda-gdb rows and the corresponding PR rows.

Recommend dropping these 3 lines.

ci/tools/run-cuda-gdb-tests
```diff
@@ -0,0 +1,44 @@
...
+wheel="${wheels[0]}"
+
+pip install "${wheel}" --group "test-cu${TEST_CUDA_MAJOR}" \
+  "cuda-bindings==${TEST_CUDA_MAJOR}.${TEST_CUDA_MINOR}.*" \
+  "cuda-core[cu${TEST_CUDA_MAJOR}]"
+
+echo "Running cuda-gdb tests"
+export NUMBA_CUDA_MLIR_GDB_BINARY="${CUDA_GDB_BINARY}"
+export NUMBA_CUDA_MLIR_ICE_FULL_TB=1
+pytest tests/test_cuda_gdb.py -v --tb=short
```
leofang: The script itself is fine and can stay almost as-is — it'll be invoked via `test-script: run-cuda-gdb-tests`. Two small cleanups once it's no longer driven by a dedicated workflow:

1. **`LOCAL_CTK`**: comes from the matrix row (already `'1'` for the rows we'll be filtering), so it doesn't need to be set explicitly by the calling workflow's `env:` block anymore. The `: "${LOCAL_CTK:?...}"` assertion at the top of this script still works.
2. **JUnit XML**: add `--junit-xml=cuda-gdb-junit-results.xml` to the `pytest` invocation on this line so the results land in a file the standard `actions/upload-artifact` step in `test-wheel-linux.yml` can pick up — same way `cudf-source` and `numbast-mlir` results are surfaced today.

tests/test_cuda_gdb.py
```diff
@@ -0,0 +1,109 @@
...
+    gdb_commands: str,
+    *,
+    timeout: int = 120,
+) -> subprocess.CompletedProcess:
+    """Write *python_src* to a temporary file and invoke cuda-gdb in batch
+    mode, passing each line of *gdb_commands* as a separate ``-ex`` argument.
+    """
+    ex_args = [
+        arg for raw in gdb_commands.splitlines() if (line := raw.strip()) for arg in ("-ex", line)
+    ]
```
leofang: Nit, optional: the walrus + double-`for` comprehension is dense for a test helper —

```python
ex_args = [
    arg for raw in gdb_commands.splitlines() if (line := raw.strip()) for arg in ("-ex", line)
]
```

reads more naturally as:

```python
ex_args = []
for raw in gdb_commands.splitlines():
    line = raw.strip()
    if line:
        ex_args.extend(("-ex", line))
```

Take it or leave it.

leofang (approved):

leofang:
_(Review summary — the body field on my [APPROVED review](https://github.com/NVIDIA/numba-cuda-mlir/pull/82#pullrequestreview-4460802591) ended up empty due to a posting quirk; reposting here. Inline threads on that review carry the per-file detail.)_

---

Approving — CI is green, so this isn't blocking. But before we merge, can we take the opportunity to do this the right way?

**The `cudf-source` / `numbast-mlir` satellite pattern is an anti-pattern, not the long-term shape we want.** Each one duplicates ~140 lines of `test-wheel-linux.yml` to add a small overlay, and they're already starting to drift from each other (matrix-key handling, `LOCAL_CTK` guarding, JUnit upload, use of `restore_llvm_artifacts`, etc.). I missed that drift when those two went in — we should not entrench it by adding a fourth copy. cuda-gdb is in fact the *lightest* of the three (no extra repo checkout, no Miniforge env, just one extra CTK component) and folds into `test-wheel-linux.yml` very cleanly.

The intended pattern is already proven in cuda-python — see [`cuda-python` test-wheel-linux.yml L289–L299](https://github.com/NVIDIA/cuda-python/blob/788bbad1ae8a901241e978d15edda3ee954ed55c/.github/workflows/test-wheel-linux.yml#L289-L299) for the compute-sanitizer overlay. Combined with the `test-script` and `matrix_filter` inputs that `test-wheel-linux.yml` already exposes here (the commented-out `coverage-report` job in `ci.yaml` exercises exactly this), it fully covers cuda-gdb. Net result: this PR's new `test-cuda-gdb-linux.yml` (+138) and the new `cuda-gdb:` block in `test-matrix.yml` (+3) both go away, and we stop perpetuating the satellite-workflow anti-pattern.

The conda-env satellites (`cudf-source`, `numbast-mlir`) need more parameterization to fold in cleanly (extra repo checkout + Miniforge env build); I'll track those as a separate follow-up — but we should stop the bleeding here while the cuda-gdb diff is still small.

Details inline.

.github/workflows/test-cuda-gdb-linux.yml
jiel-nv: This approach is much cleaner. I'll take it. Thanks!

.github/workflows/ci.yaml
jiel-nv: The recommended shape looks good. One issue: the filter returns an empty matrix, because the existing matrix has no py3.12 / 13.2.1 / LOCAL_CTK=1 combination (those 3.12/13.2.1 rows are LOCAL_CTK=0, and cuda-gdb needs a local CTK).

Would you be OK with me adding two new rows to `test-matrix.yml` to cover it?

ci/test-matrix.yml
jiel-nv: Yes, the `matrix_filter` approach is good, but without adding two new rows to `test-matrix.yml`, the preferred py3.12 / 13.2.1 / LOCAL_CTK=1 combination can't be set up.

ci/tools/run-cuda-gdb-tests
jiel-nv: One issue on the **JUnit XML** upload: `test-wheel-linux.yml` doesn't actually have an `actions/upload-artifact` step today. The `cudf-source` and `numbast-mlir` uploads live in their own dedicated workflow, but since we're going to delete the dedicated cuda-gdb workflow and folding it into the shared `test-wheel-linux.yml`, there's no existing upload step to "pick up" the XML.

tests/test_cuda_gdb.py
jiel-nv: Thanks and I'll take it.

jiel-nv:
@leofang Thanks for the detailed review! I followed your suggested approach, folding the cuda-gdb tests into `test-wheel-linux.yml`, with two small add-ons to cover the issues I flagged inline:

Add-on 1: Added two new matrix rows (py3.12 / 13.2.1 / LOCAL_CTK=1 for l4/amd64 and a100/arm64). That exact combination didn't exist in the pull-request matrix. Please let me know if the added CI cost on regular test jobs is a concern.

Add-on 2: Added an artifact upload step for the cuda-gdb test results in `test-wheel-linux.yml`, since the dedicated cuda-gdb workflow was deleted but the shared workflow didn't already have one to pick up the JUnit XML.

leofang:
LGTM!

PR #84, opened 2026-05-27
Add support for custom dtype

### LINKED ISSUE #83: [BUG] cuda.local.array fails when dtype is assigned a Python object that maps to types.DTypeSpec
# `cuda.local.array` fails when `dtype` is assigned a Python object that maps to `types.DTypeSpec`

## Describe the bug
When passing a custom Python object (that represents a datatype and maps to a subclass of Numba's `types.DTypeSpec`) to `cuda.local.array(..., dtype=my_custom_dtype)`, the MLIR compiler fails with a `TypeError: No conversion found for type <class 'MyCustomDType'>`.

## To Reproduce
```python
import numpy as np
from numba.core import types
from numba.extending import typeof_impl

class MyCustomDType:
    pass

my_dtype = MyCustomDType()

@typeof_impl.register(MyCustomDType)
def typeof_my_dtype(val, c):
    return types.NumberClass(types.float32)

@cuda.jit
def kernel(out):
    # Fails during compilation inside `cuda.local.array` lowering
    arr = cuda.local.array((10,), dtype=my_dtype)
    arr[0] = 42.0
    out[0] = arr[0]

def main():
    out = np.zeros(1, dtype=np.float32)
    kernel[1, 1](out)
    print("Result:", out[0])
    assert out[0] == 42.0
    print("Success!")

if __name__ == "__main__":
    main()
```

## Error Output
```
Traceback (most recent call last):
...
  File "/local/home/yhavrylko/Projects/numba-simt-mlir-compiler/src/numba_cuda_mlir/lowering/cuda.py", line 437, in cuda_local_array
    mlir_dtype = to_mlir_type(np_dtype)
                 ^^^^^^^^^^^^^^^^^^^^^^
...
  File "/local/home/yhavrylko/Projects/numba-simt-mlir-compiler/src/numba_cuda_mlir/lowering_utilities/type_conversions.py", line 192, in to_mlir_type
    raise TypeError(f"No conversion found for type {type(obj)}")
TypeError: No conversion found for type <class '__main__.MyCustomDType'>
```

## Root Cause
The issue is deeply rooted in how `src/numba_cuda_mlir/mlir_lowering.py` handles constant/global assignments for custom type objects. 

Currently, `lower_global_assign` and `lower_var_assign` check if the python value is a subclass of `types.Type`. However, if the python object is a custom type instance (like `MyCustomDType()`), it does not inherit from `types.Type` (even though the Numba frontend infers it as `types.NumberClass` which inherits from `types.DTypeSpec`). 

As a result, the MLIR lowering fallback simply stores the raw Python object into `self.varmap` instead of materializing an MLIR type token (e.g. `arith.constant 0`). 

When `cuda_local_array` calls `_resolve_dtype(lower, dtype_var)`, it attempts to use `lower.load_var(dtype_var)` which returns the raw Python object (`MyCustomDType`) back to the MLIR memory allocation layer, where `to_mlir_type()` crashes.

## The Fix
Since the Numba frontend already successfully determined that the object acts as a type specifier (by resolving it to `types.DTypeSpec`), `mlir_lowering.py` should materialize an MLIR type token based on the inferred Numba type, regardless of what the original Python object's class inheritance looks like.

By modifying the type token check in `mlir_lowering.py` to:
```python
        elif isinstance(target_type, types.DTypeSpec):
            self.store_var(target, self._materialize_type_token(target_type))
```
`load_var()` properly returns the MLIR type value (`arith.constant 0 : f32`), which `to_mlir_type` naturally resolves to the correct MLIR Type when `_resolve_dtype` returns it to `cuda.local.array`.

ZzEeKkAa:
## Summary

Fixes a lowering bug where custom Python objects acting as data types (resolving to `types.DTypeSpec` in Numba's frontend) would crash the MLIR compiler when passed to `cuda.local.array` or `cuda.shared.array`.

## Root Cause

When a custom Python object (e.g. `Complex(np.float32)`) is assigned to a variable and acts as a data type, Numba's frontend correctly infers its type as a `types.NumberClass` (which inherits from `types.DTypeSpec`). 

However, in `mlir_lowering.py`, `lower_global_assign` and `lower_var_assign` previously checked `isinstance(var_value, types.Type)`. Since the custom Python object itself is not a subclass of `types.Type`, the lowering phase fell back to storing the raw Python object in `self.varmap` instead of materializing a proper MLIR type token (like `arith.constant 0`).

When `cuda_local_array` later called `_resolve_dtype(lower, dtype_var)`, it loaded this raw Python object from the varmap and passed it to `to_mlir_type()`, which crashed with `TypeError: No conversion found for type <class '...'>`.

## Fix

1. **`mlir_lowering.py`**: Changed the fallback type token check in `lower_var_assign` and `lower_global_assign` from checking the Python value's inheritance to checking the Numba frontend type:
   ```python
   case _ if isinstance(self.get_numba_type(target.name), types.DTypeSpec):
       target_type = self.get_numba_type(target.name)
       self.store_var(target, self._materialize_type_token(target_type))

Closes https://github.com/NVIDIA/numba-cuda-mlir/issues/83

### DIFF
--- a/src/numba_cuda_mlir/mlir_lowering.py
+++ b/src/numba_cuda_mlir/mlir_lowering.py
@@ -936,7 +936,11 @@ extern "C" __global__ void
         elif hasattr(free_var.value, "__cuda_array_interface__"):
             self.lower_captured_array_to_memref(target, free_var.value)
         else:
-            self.store_var(target, free_var.value)
+            target_type = self.get_numba_type(target.name)
+            if isinstance(target_type, types.DTypeSpec):
+                self.store_var(target, self._materialize_type_token(target_type))
+            else:
+                self.store_var(target, free_var.value)
 
     def lower_captured_array_to_memref(self, target, pyval):
         """Build an MLIR memref value from __cuda_array_interface__ metadata.
@@ -1159,7 +1163,7 @@ extern "C" __global__ void
             )
         elif hasattr(glob.value, "__cuda_array_interface__"):
             self.lower_captured_array_to_memref(target, glob.value)
-        elif isinstance(target_type, types.NumberClass) and isinstance(glob.value, types.Type):
+        elif isinstance(target_type, types.DTypeSpec):
             self.store_var(target, self._materialize_type_token(target_type))
         else:
             self.store_var(target, glob.value)
@@ -1180,7 +1184,7 @@ extern "C" __global__ void
                 self.store_var(target, var_value)
             case str() | bytes():
                 self.store_var(target, var_value)
-            case types.Type() if isinstance(self.get_numba_type(target.name), types.NumberClass):
+            case _ if isinstance(self.get_numba_type(target.name), types.DTypeSpec):
                 target_type = self.get_numba_type(target.name)
                 self.store_var(target, self._materialize_type_token(target_type))
             case None:
--- a/src/numba_cuda_mlir/models.py
+++ b/src/numba_cuda_mlir/models.py
@@ -133,11 +133,11 @@ class StructModel(DataModel):
         return self._models
 
 
-@register_model(types.NumberClass)
-class NumpyDataTypeModel(PrimitiveModel):
+@register_model(types.DTypeSpec)
+class DTypeSpecModel(PrimitiveModel):
     """
-    Convert from a Numba NumberClass type instance
-    into the MLIR type that this NumberClass instance represents.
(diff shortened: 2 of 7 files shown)


atmnp (approved):

Jlisowskyy (approved):
LGTM, works fine locally

PR #85, opened 2026-05-27
Fix to_numba_type for np dtypes

ZzEeKkAa:
## Description

This PR improves the robustness of type conversions in `type_conversions.py` for NumPy and `ctypes` types, and adds comprehensive, parameterized test coverage for these utilities, including extension API support.

### Key Changes
- **Improved NumPy Type Resolution**: Updated `to_numba_type` and `to_mlir_type` to use `issubclass(obj, np.generic)` instead of checking `__module__ == "numpy"`. This properly catches NumPy scalar types (like `np.float32`, `np.complex64`, etc.) passed as types and maps them correctly to their respective Numba/MLIR types.
- **Removed Invalid Ctype**: Removed `ctypes.c_float16` from `ctypes_type_to_numba_type` as it does not exist in standard Python `ctypes` and was causing an `AttributeError`.
- **Parameterized Unit Tests**: Added `tests/test_type_conversions.py` with `pytest.mark.parametrize` to assert that `to_numba_type` correctly round-trips for NumPy scalar types, `np.dtype` wrappers, and `ctypes` scalar equivalents. 
- **Extension API Testing**: Added a test asserting that custom frontend types dynamically registered via the `@to_numba_type.register` extension API are accurately resolved.

## Testing
- [x] Added `test_type_conversions.py` containing:
  - `test_numpy_scalar_type_conversion` (Parameterized)
  - `test_numpy_dtype_conversion` (Parameterized)
  - `test_ctypes_conversion` (Parameterized)
  - `test_custom_type_conversion`
- [x] Verified `pytest -v tests/test_type_conversions.py` passes successfully locally.
- [x] Tested against `nvmath-python`'s `cufftdx` example test suite.

### DIFF
--- a/tests/test_type_conversions.py
+++ b/tests/test_type_conversions.py
@@ -1,0 +1,34 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""
+Typing conversion tests.
+"""
+
+import ctypes
+import numpy as np
+import pytest
+
+from numba_cuda_mlir.numba_cuda import types
+from numba_cuda_mlir.lowering_utilities.type_conversions import to_numba_type
+
+def test_numpy_scalar_type_conversion():
+    """Verify that numpy scalar types convert to correct numba types."""
+    assert to_numba_type(np.float32) == types.float32
+    assert to_numba_type(np.float64) == types.float64
+    assert to_numba_type(np.float16) == types.float16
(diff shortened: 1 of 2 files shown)


tests/test_type_conversions.py
atmnp, line 1: Should we also add test for unsigned integer types?

ZzEeKkAa: Good point. Actually there was something missing in the src for uint support. Added support there as well

Jlisowskyy (commented):
LGTM! I think since this touches the compiler directly, we should get @atmnp sign-off before merging

atmnp (approved):

PR #86, opened 2026-05-28
cuda vector -> np.complex

Jlisowskyy:
Completes np.complex64/np.complex128 cast

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/typing/builtins.py
+++ b/src/numba_cuda_mlir/numba_cuda/typing/builtins.py
@@ -811,17 +811,13 @@ class NumberClassAttribute(AttributeTemplate):
                 if isinstance(val, types.Array) and val.ndim == 0 and val.dtype == ty:
                     # This is 0d array -> scalar degrading
                     return ty
-
-                if (
+                elif (
                     isinstance(val, VectorType)
                     and val.length == 2
                     and isinstance(ty, types.Complex)
                     and isinstance(val.dtype, types.Float)
                 ):
-                    if (ty.bitwidth == 64 and val.dtype.bitwidth <= 32) or (
-                        ty.bitwidth == 128 and val.dtype.bitwidth <= 64
-                    ):
-                        return ty
+                    return ty
 
                 # unsupported
                 msg = f"Casting {val} to {ty} directly is unsupported."
--- a/src/numba_cuda_mlir/numba_cuda/typing/builtins.py
+++ b/src/numba_cuda_mlir/numba_cuda/typing/builtins.py
@@ -27,6 +27,8 @@ from numba_cuda_mlir.numba_cuda.extending import (
     make_attribute_wrapper,
 )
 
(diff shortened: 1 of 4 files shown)


tests/test_vector_type_complex_cast.py
```diff
@@ -115,3 +115,59 @@ def kernel(matrix, out):
...
+        @cuda.jit("void(complex128[:])")
+        def kernel(out):
+            v = float32x2(1.0, 2.0)
+            out[0] = np.complex128(v)
+
+        res = np.zeros(1, dtype=np.complex128)
+        kernel[1, 1](res)
+        self.assertEqual(res[0], np.complex128(1.0 + 2.0j))
+
+    def test_np_complex64_from_float64x2_rejected(self):
```
ZzEeKkAa: I'm wondering if we should reject it here. You can cast from int64 to int32 without problems. It was always users responsibility to avoid overflow...

src/numba_cuda_mlir/numba_cuda/typing/builtins.py
```diff
@@ -809,13 +811,24 @@ def typer(val):
...
-                        msg += f" Try doing '<array>.astype(np.{ty})' instead"
-                    raise errors.TypingError(msg)
+
+                if (
+                    isinstance(val, VectorType)
+                    and val.length == 2
+                    and isinstance(ty, types.Complex)
+                    and isinstance(val.dtype, types.Float)
+                ):
+                    if (ty.bitwidth == 64 and val.dtype.bitwidth <= 32) or (
```
ZzEeKkAa: Could we just combine these two ifs togeather?

src/numba_cuda_mlir/numba_cuda/typing/builtins.py
```diff
@@ -809,13 +811,24 @@ def typer(val):
...
+                        ty.bitwidth == 128 and val.dtype.bitwidth <= 64
+                    ):
+                        return ty
+
+                # unsupported
+                msg = f"Casting {val} to {ty} directly is unsupported."
```
ZzEeKkAa, line 823: I love not having else, but let's be consistent with other side of the function (+ keep git blame clean). Maybe `if ... elif ... else`?

ZzEeKkAa:
Looks good, just curious about `complex64(float64x2)`

src/numba_cuda_mlir/numba_cuda/typing/builtins.py
Jlisowskyy: Good point, I agree. Fixed

src/numba_cuda_mlir/numba_cuda/typing/builtins.py
Jlisowskyy: Oh. Sure, done!

tests/test_vector_type_complex_cast.py
Jlisowskyy: Right, let's allow that

ZzEeKkAa (approved):

PR #88, opened 2026-05-28
Add complex 2args constructor support

ZzEeKkAa:
## Summary
Adds support for constructing complex numbers inside CUDA kernels using the two-argument form of NumPy type constructors, e.g., `np.complex64(real, imag)` and `np.complex128(real, imag)`. Previously, only the single-argument addition form (`np.complex64(real + imag * 1j)`) was supported.

### Key Changes:
- **Typing**: Updated `NumberClassAttribute.resolve___call__` in `typing/builtins.py` to accept an optional `imag` argument. The typing logic has been refactored into `typer_real` and `typer_complex` to cleanly handle single-argument vs. two-argument numeric construction. It raises a `TypingError` if a user attempts to pass two arguments to non-complex type constructors.
- **Lowering**: Mapped `@lower(types.NumberClass, types.Number, types.Number)` to the existing `complex_constructor_2args_cg` function in `lowering/cmath.py`, which directly constructs the MLIR complex type (`complex.create`) from the provided `real` and `imag` components.
- **Cleanup**: Ensured all imports in `lowering/builtins.py` are at the top of the file according to workspace rules.
- **Testing**: Added a parameterized test suite in `tests/test_complex.py` to verify the 2-argument constructors behave correctly for `np.complex64`, `np.complex128`, and the built-in `complex`. 

## Test Plan
- Ran the newly added `test_complex_constructor_2args` test locally via the GPU wrapper (`./internal/cursor/gpu_run.sh 'PYTHONPATH=src .venv/bin/python -m pytest tests/test_complex.py -k test_complex_constructor_2args'`).
- Ran the full `ci/tools/run-tests` test suite to ensure no regressions were introduced.

### DIFF
--- a/src/numba_cuda_mlir/lowering/cmath.py
+++ b/src/numba_cuda_mlir/lowering/cmath.py
@@ -43,6 +43,7 @@ SQRT2 = 1.414213562373095048801688724209698079
 
 
 @lower(complex, types.Number, types.Number)
+@lower(types.NumberClass, types.Number, types.Number)
 def complex_constructor_2args_cg(mlir_lower, target, args, kwargs):
     """Code generator for complex(real, imag) constructor."""
     assert not kwargs, "complex constructor does not accept keyword arguments"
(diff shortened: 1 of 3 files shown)


Jlisowskyy (approved):
LGTM!

PR #90, opened 2026-06-01
Optimize CUDA array interface launch caching

atmnp:
Summary:
- pass CUDA Array Interface objects directly to the C++ launcher to avoid rebuilding DeviceNDArray wrappers on hot launches
- cache array launch signatures by dtype, ndim, layout, and readonly state so cached argtypes match Numba typing semantics
- reject masked CUDA Array Interface objects before pass-through launch, preserving the previous unsupported-mask behavior
- add regression coverage for pass-through launches, dtype/ndim cache reuse, readonly cache separation, and masked-array rejection

### DIFF
--- a/src/numba_cuda_mlir/descriptor.py
+++ b/src/numba_cuda_mlir/descriptor.py
@@ -29,6 +29,7 @@ from numba_cuda_mlir.numba_cuda import serialize, typing
 from numba_cuda_mlir.numba_cuda.core.base import BaseContext
 from numba_cuda_mlir.numba_cuda.core.callconv import MinimalCallConv
 from numba_cuda_mlir.numba_cuda.cudadrv.devicearray import DeviceNDArrayBase
+from numba_cuda_mlir.numba_cuda.np import numpy_support
 from numba_cuda_mlir.numba_cuda.core.descriptors import TargetDescriptor
 from numba_cuda_mlir.numba_cuda.core.compiler_lock import global_compiler_lock
 from numba_cuda_mlir.numba_cuda.dispatcher import Dispatcher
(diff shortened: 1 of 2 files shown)


leofang (approved):
Verified locally on an RTX 6000 Ada (sm_89), CUDA 13.2, driver 595.45.04, against this PR's branch head applied to a stock 0.3.0 install (`src/numba_cuda_mlir/descriptor.py` swapped in).

Benchmark: vector add `c = a + b` on three CuPy arrays (`block=256`, `n_repeat=200`, `max_duration=2 s` per cell). Reporting `cpu_median` — host-side per-launch wall clock — since that isolates the launch overhead this PR targets.

Small-N float32 (launch-overhead-dominated):

| framework | N=1024 baseline | N=1024 patched | speedup |
|---|---:|---:|---:|
| numba_cuda | 89.25 µs | 87.58 µs | 1.02× |
| numba_cuda_cached | 89.77 µs | 87.61 µs | 1.02× |
| numba_cuda_mlir | **154.46 µs** | **50.13 µs** | **3.08×** |
| numba_cuda_mlir_cached | 153.76 µs | 49.86 µs | 3.08× |

`numba_cuda` is within run-to-run noise (PR doesn't touch its code path), confirming the speedup is targeted.

Consistent across dtypes (`numba_cuda_mlir` @ N=1024): float32 3.08×, float64 2.93×, int32 2.76×, int64 2.64×.

Consistent across sizes (`numba_cuda_mlir` @ float32): N=1024 3.08×, N=1M 3.06×, N=64M 3.00×.

Bottom line: post-PR, `numba_cuda_mlir`'s per-launch floor is roughly 40% lower than `numba_cuda`'s (50 vs 88 µs on this host), where it was previously ~75% higher. The fix is dtype- and N-independent in this measurement, as expected for a launch-path change.

Approving. Leaving a follow-up comment with a Cython-vs-Python micro-bench suggesting an architectural end-state for the same code path.

— Leo's bot

leofang:
Going one step further, the architectural end-state for this code path is to drive *both* sides — Python-side cache key and C++ launcher in `_cext` — onto `cuda.core.utils.StridedMemoryView` (Cython) as the single view abstraction.

Per-arg micro-bench on the same Ada host (single contiguous CuPy float32 array, just the metadata-extraction step):

| per-arg cost | median |
|---|---:|
| this PR's `_array_cache_key + _sync_cuda_array_interface_stream` (pure Python) | 3350 ns |
| `StridedMemoryView.from_any_interface` (Cython, prefers DLPack) | 1730 ns |
| `StridedMemoryView.from_dlpack` (Cython) | 1630 ns |

DLPack wins cleanly because CuPy's `__dlpack__` returns a C-struct directly; CAI still requires constructing a Python dict each call.

What the coordinated change looks like:

- **Python side**: replace per-arg `_array_cache_key + _sync_cuda_array_interface_stream` with `StridedMemoryView.from_any_interface(arg, stream_ptr=...)`. Aligns with numba-cuda 0.30, which already uses `StridedMemoryView`.
- **C++ launcher side (`_cext`)**: ingress from `StridedMemoryView` (or accept pre-parsed `(ptr, shape, strides)`) instead of re-reading `__cuda_array_interface__` from the original Python object.

Today the CAI parse happens twice — once in Python for caching (this PR), once in C++ for the actual launch. Driving both onto SMV kills the duplicate parse and gets the full Cython benefit on both sides.

Projected end-to-end on the same N=1024 float32 cell as the review (50.13 µs above): the Python-side change alone projects to ~45 µs (3 × (3.35 − 1.73) ≈ 5 µs saved per launch); further savings on the C++ side depend on how much the current re-parse costs.

— Leo's bot

PR #92, opened 2026-06-02
[Compilation] Avoid finalizing internal device callees

atmnp:
Route nested device function calls through a lightweight callee compile path so MLIR callees can be cloned into parent kernels without eagerly optimizing and retaining cubins for every device function.

We keep public compile_device() finalized so inspect_asm() continues to produce PTX.


Jlisowskyy (approved):
LGTM

PR #93, opened 2026-06-02
Remove llvmlite dependency

gmarkall:
This removes the llvmlite dependency by either removing code that uses llvmlite, or creating a placeholder module for it that allows everything to import successfully, without us going through the whole process of actually removing all the builder code whilst trying to keep everything working consistently.

This work was done in stages - see individual commit messages for details of each stage / category.

A follow-up PR will address dead code removal.


gmarkall:
I think this now fails because Numbast has a dependency on llvmlite that does not now get installed:

```
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/tmp/miniforge/envs/numbast-mlir/lib/python3.12/site-packages/numbast/experimental/mlir/__init__.py", line 4, in <module>
    from numbast.experimental.mlir.struct import bind_cxx_struct, bind_cxx_structs
  File "/tmp/miniforge/envs/numbast-mlir/lib/python3.12/site-packages/numbast/experimental/mlir/struct.py", line 7, in <module>
    from llvmlite import ir
ModuleNotFoundError: No module named 'llvmlite'
```

**Follow-up / edit:** I addressed this in https://github.com/NVIDIA/numba-cuda-mlir/pull/93/commits/450a03933fb3ffa8044d87b4461cd7555c062f36, by adding llvmlite to the Numbast dependencies that the CI script installs.

atmnp (approved):
LGTM

leofang:
CI changes LGTM!

PR #96, opened 2026-06-02
Permit `NamedTuple` in a kernel + fix `None` start in shmem view

atmnp:
This removes 4 xfails as well as fixes two bugs needed for nvmath support.

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/api_util.py
+++ b/src/numba_cuda_mlir/numba_cuda/api_util.py
@@ -8,6 +8,21 @@ import numpy as np
 import functools
 
 
+def resolve_cuda_array_interface_dtype(typestr, owner_dtype=None):
+    dtype = np.dtype(typestr)
+    if owner_dtype is None:
+        return dtype
+
+    try:
+        owner_dtype = np.dtype(owner_dtype)
+    except TypeError:
+        return dtype
+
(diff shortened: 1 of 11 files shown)


src/numba_cuda_mlir/numba_cuda/api_util.py
```diff
@@ -8,6 +8,21 @@
 import functools
 
 
+def resolve_cuda_array_interface_dtype(typestr, owner_dtype=None):
```
gmarkall, line 11: I think this addresses the shortcoming of the CAI that the owner's dtype might be something not expressible in CAI typestrs - is that understanding correct?

gmarkall (approved):
This looks good to me. I have a question on the diff to ensure my understanding is correct, but it's not a blocking question.

src/numba_cuda_mlir/numba_cuda/api_util.py
atmnp: Yes, that’s the issue this fixes. Some newly enabled tests exposed that using only the CAI typestr can lose the dtype the owner actually carries. We now prefer owner.dtype when NumPy can interpret it and its itemsize matches the CAI dtype. If the owner dtype is absent, invalid, or a different size, we fall back to the CAI typestr.

Jlisowskyy (approved):
LGTM!

PR #100, opened 2026-06-03
Fold literal string attribute comparisons before materialization

atmnp:


atmnp:
Will re-open a better fix once I triage the problem this is intended to solve for nvmath more deeply. This has only partially fixed the problem.

PR #101, opened 2026-06-03
Defer CUDA source compilation in linker

isVoid:
## Motivation

This is a required (correctness) feature for nvshmem4py, and IMO a useful (performance) feature to improve the toolchains in numba-cuda-mlir.

## nvshmem4py correctness

The compilation of nvshmem4py numba bindings uses a single compilation unit that includes the nvshmem headers and shim functions. The headers contains a single static declaration of device state variable. In the current state, multiple CUs will lead to multiple definitions of the device state variable and causing duplicate symbols at link time.

## Performance improvement

Numba-cuda's CUSource object is designed in a way that the underlying data can be backed by a `StringIO` object. Allowing user to batching CUDA C++ functions to a point where `.data` attribute is accessed. This reduces the total number of times the `nvrtc` instance is called upon.

## Implementation

The implementation is straight forward - for every call on the linker to add a cusource, don't compile it immediately but cache it in a buffer. At the completion time of the Linker, call a new function that materialize these source codes. The completion boundary is defined at these funcitons: `get_linked_ptx`, `complete`, `recreate_with_lto`. These boundaries, in the compilation pipeline, all takes place after the lowering phase has completed (or is not in the pipeline), so the correctness of the deferred source can be guaranteed.

## Tests

Tests are written in a way that mocks the linker objects with additional call counters. They are there to prevent cusource materialization prior to linker completion.


gmarkall:
Thanks - I had been looking at the linker logic to try and review #71, and I did not like that we create linkers then discard them and the objects we might have added if we decide we need to change the LTO "state".

atmnp (approved):

PR #102, opened 2026-06-03
Drop implicit `nvjitlink` dependency derived from `cudatoolkit` 

brandon-b-miller:
Drops `nvjitlink` from the list of extras installed via the `cudatoolkit` pypi metapackages. With the current setup, a downstream library that pins the `cudatoolkit` package implicitly pins `nvjitlink` as well, which is a problem if they need 12.9 for instance to ensure MVC. 

The explicit nvjitlink dependency listed should be sufficient.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,12 +31,12 @@ dependencies = [
 [project.optional-dependencies]
 cu12 = [
     "cuda-bindings>=12.9.1,<13.0.0",
-    "cuda-toolkit[cudart,nvcc,nvrtc,nvjitlink,cccl]==12.*",
+    "cuda-toolkit[cudart,nvcc,nvrtc,cccl]==12.*",
     "nvidia-nvjitlink-cu12>=12.3.0,<13.0.0",
 ]
 cu13 = [
     "cuda-bindings==13.*",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -31,12 +31,12 @@ dependencies = [
...
 cu12 = [
     "cuda-bindings>=12.9.1,<13.0.0",
-    "cuda-toolkit[cudart,nvcc,nvrtc,nvjitlink,cccl]==12.*",
+    "cuda-toolkit[cudart,nvcc,nvrtc,cccl]==12.*",
     "nvidia-nvjitlink-cu12>=12.3.0,<13.0.0",
 ]
 cu13 = [
     "cuda-bindings==13.*",
-    "cuda-toolkit[cudart,nvcc,nvrtc,nvjitlink,cccl]==13.*",
+    "cuda-toolkit[cudart,nvcc,nvrtc,cccl]==13.*",
```
leofang, line 46: We need to add a generous nvjitlink pin here.

leofang (changes requested):

pyproject.toml
brandon-b-miller: Added for cuda 13 and vendored the comments from numba-cuda

brandon-b-miller:
ok to test

leofang (approved):

PR #107, opened 2026-06-08
Delete llvmlite code

gmarkall:
Following on from the removal of the llvmlite dependency, this removes a lot of code that used llvmlite and was otherwise dead.

The changes / deletions are generated with Claude Code - I guided it through the process of removing the code and reviewed the changes as they were made, so I believe them all to be reasonable / sensible.


atmnp (approved):
LGTM! Thanks Graham

PR #114, opened 2026-06-10
BUG: Remove packaging dependency from numba_cuda._compat

### LINKED ISSUE #106: [BUG] Unable to import cuda from numba_cuda_mlir after following INSTALL.md instructions
**Describe the bug**
When trying to run a simple import command, I was surprised to come across this error: `ModuleNotFoundError: No module named 'packaging'`

The `packaging` module is imported [here](https://github.com/NVIDIA/numba-cuda-mlir/blob/main/src/numba_cuda_mlir/numba_cuda/_compat.py#L3), but it's missing from `pyproject.yaml`.

Follow steps in [INSTALL.md](https://github.com/NVIDIA/numba-cuda-mlir/blob/main/INSTALL.md#option-2-editable-install-with-pre-built-llvm) using the uv package manager, and then run `python -c "from numba_cuda_mlir import cuda"`.

**Expected behavior**
Importing `cuda` from `numba_cuda_mlir` should silently succeed and not throw the `ModuleNotFoundError` exception

**Environment details (please complete the following information):**
 - Environment location: local WSL2 session
 - Method of numba-cuda-mlir install: [conda, Docker, or from source]
   - Run `uv venv venv --python 3.12`
   - `source venv/bin/activate`
   - Download LLVM pre-built binaries
   - `MLIR_DIR=$PWD/llvm-modern-install/lib/cmake/mlir LIBLLVM7=$PWD/llvm7-install/lib/libLLVM-7.so  uv pip install -e '.[cu13]' --group dev --config-settings editable_mode=compat`

**Additional context**
This originally came up when I was going to see if I could figure out what was going on with https://github.com/NVIDIA/numba-cuda-mlir/issues/76

acosmicflamingo:
Fixes #106 

Rather than add another runtime dependency to numba-cuda-mlir to fix the import issue, we can utilize the `cuda.core._version.__version__tuple` attribute and Python's already powerful tuple comparison functionality to mimic the previous version-comparing behavior.

It also appears that `CUDA_CORE_VERSION` isn't used anywhere else in the codebase, so there aren't any places where the interpreter would expect a `Version` object. Since it's only used to ensure `CUDA_CORE_GT_0_6` and `CUDA_CORE_GE_1_0` are correctly configured before being imported in files like `./src/numba_cuda_mlir/numba_cuda/cudadrv/driver.py`, this PR should be enough to address the issue at hand.

This PR also assumes that any string after the patch number is seen as a dev build:
```python
(1, 0, 0, 'dev', 'g16df11dab') >= (1, 0, 0)   # True
(1, 0, 0, 'alpha', 'g16df11dab') >= (1, 0, 0) # Still true
```

Of course, this corner-case could be covered, but then the work to cover the corner cases might warrant shipping a new `cuda-versions` package (or just relying on packaging after all) 😜

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/_compat.py
+++ b/src/numba_cuda_mlir/numba_cuda/_compat.py
@@ -2,9 +2,8 @@
 # SPDX-License-Identifier: BSD-2-Clause
 from cuda import core
 
-# Utilize core._version attributes
-#    __version__ => '1.0.2.dev46+g16df11dab'
-#    __version_tuple__ => (1, 0, 2, 'dev46', 'g16df11dab')
-CUDA_CORE_VERSION = core._version.__version_tuple__
+# core.__version__ returns str like '1.0.1',
+# '1.0.2.dev46+g16df11dab', etc.
+CUDA_CORE_VERSION = tuple(int(i) for i in core.__version__.split(".") if i.isdigit())
 CUDA_CORE_GT_0_6 = CUDA_CORE_VERSION >= (0, 6, 0)
 CUDA_CORE_GE_1_0 = CUDA_CORE_VERSION >= (1, 0, 0)
--- a/src/numba_cuda_mlir/numba_cuda/_compat.py
+++ b/src/numba_cuda_mlir/numba_cuda/_compat.py
@@ -1,9 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
-from packaging import version
 from cuda import core
 
-
-CUDA_CORE_VERSION = version.parse(core.__version__)
-CUDA_CORE_GT_0_6 = CUDA_CORE_VERSION >= version.parse("0.6.0")
-CUDA_CORE_GE_1_0 = CUDA_CORE_VERSION >= version.parse("1.0.0")
+# Utilize core._version attributes
+#    __version__ => '1.0.2.dev46+g16df11dab'
+#    __version_tuple__ => (1, 0, 2, 'dev46', 'g16df11dab')
+CUDA_CORE_VERSION = core._version.__version_tuple__
+CUDA_CORE_GT_0_6 = CUDA_CORE_VERSION >= (0, 6, 0)
+CUDA_CORE_GE_1_0 = CUDA_CORE_VERSION >= (1, 0, 0)


gmarkall:
/ok to test https://github.com/NVIDIA/numba-cuda-mlir/pull/114/commits/65fcc597c2baad84e2016d3e1a2ebc6c3d8b9be0

src/numba_cuda_mlir/numba_cuda/_compat.py
```diff
@@ -1,9 +1,10 @@
...
 from cuda import core
 
-
-CUDA_CORE_VERSION = version.parse(core.__version__)
-CUDA_CORE_GT_0_6 = CUDA_CORE_VERSION >= version.parse("0.6.0")
-CUDA_CORE_GE_1_0 = CUDA_CORE_VERSION >= version.parse("1.0.0")
+# Utilize core._version attributes
+#    __version__ => '1.0.2.dev46+g16df11dab'
+#    __version_tuple__ => (1, 0, 2, 'dev46', 'g16df11dab')
+CUDA_CORE_VERSION = core._version.__version_tuple__
```
leofang: Is there a way to access `__version_tuple__` without reaching into the private module?

acosmicflamingo: Currently, there isn't. The closest I can get to accessing content in the private module is doing something like `from importlib.metadata import version; version('cuda.core')`, but that's silly compared to using `cuda.core.__version__`.

With that being said, there is a change I can make in `cuda.core.__init__.py` that would make it possible. I found that `platformdirs` does something like `from .version import __version_tuple__ as __version_info__` [here](https://github.com/tox-dev/platformdirs/blob/main/src/platformdirs/__init__.py#L18). I think the Python community already associates `version_info`, `__version_info__` as a module's version in tuple form (e.g. it's common to check Python version like this: `sys.version_info < (3, 0)`), so we could lean into that.

What are your thoughts? Should I create a PR in `cuda-python` repo with this change? I can always try other approaches if none of this sounds too appealing to you.

Edit: I could also take a list comprehension approach (`(tuple(int(i) for i in cuda.core.__version__.split('.') if i.isdigit())`, seems like the cleanest path if we don't want to touch `cuda.core`)

leofang: Yes I have in mind we should split the version string. This means to work with already-published cuda-core packages, so nothing to change on the other side.

acosmicflamingo: Sounds good! Thanks!

src/numba_cuda_mlir/numba_cuda/_compat.py
```diff
@@ -1,9 +1,9 @@
...
-from packaging import version
 from cuda import core
 
-
-CUDA_CORE_VERSION = version.parse(core.__version__)
-CUDA_CORE_GT_0_6 = CUDA_CORE_VERSION >= version.parse("0.6.0")
-CUDA_CORE_GE_1_0 = CUDA_CORE_VERSION >= version.parse("1.0.0")
+# core.__version__ returns str like '1.0.1',
+# '1.0.2.dev46+g16df11dab', etc.
+CUDA_CORE_VERSION = tuple(int(i) for i in core.__version__.split(".") if i.isdigit())
```
leofang, line 7: My bot told me that versions like `1.0.0rc1` and `1.0.0a2` would fail here. I think it is acceptable given that we have dropped the RC practice in the cuda-python repo.

leofang (approved):
LGTM, thanks!

leofang:
/ok to test [9b30608](https://github.com/NVIDIA/numba-cuda-mlir/pull/114/commits/9b306082f3d96df3c54791e7bfb82e3cbf5dd487)

acosmicflamingo:
😮 your bot is right! Alright well if we know that the patch number will always be a digit, then great. Otherwise, I can push a change that addresses this in 30 minutes

acosmicflamingo:
@leofang shall I push this change in the off-chance cuda-python does decide to go back to using RC string in their patch number? I verified locally that this change would ensure that versions like `1.0.2.dev46+g16df11dab` and `1.0.2rc.dev46+g16df11dab` would all be evaluated to `(1, 0, 2)`.
```diff
diff --git a/src/numba_cuda_mlir/numba_cuda/_compat.py b/src/numba_cuda_mlir/numba_cuda/_compat.py
index f122015..4c7f027 100644
--- a/src/numba_cuda_mlir/numba_cuda/_compat.py
+++ b/src/numba_cuda_mlir/numba_cuda/_compat.py
@@ -2,8 +2,11 @@
 # SPDX-License-Identifier: BSD-2-Clause
 from cuda import core
 
+import re
+
 # core.__version__ returns str like '1.0.1',
 # '1.0.2.dev46+g16df11dab', etc.
-CUDA_CORE_VERSION = tuple(int(i) for i in core.__version__.split(".") if i.isdigit())
+sanitized_version = re.search(r'[\d.]+', core.__version__).group(0)
+CUDA_CORE_VERSION = tuple(int(i) for i in sanitized_version.split(".") if i.isdigit())
 CUDA_CORE_GT_0_6 = CUDA_CORE_VERSION >= (0, 6, 0)
 CUDA_CORE_GE_1_0 = CUDA_CORE_VERSION >= (1, 0, 0)
```

PR #117, opened 2026-06-11
Replace unconditional LTOIR linker optimization disabling with a user-controlled + warning.

Jlisowskyy:
The fix introduced in #80 disabled all LTOIR optimizations when linking with LTOIR. This made LTOIR libraries practically unusable, since they are primarily intended for performance-critical use cases and the performance benefit was lost.

Example performance difference:

### Opts disabled
```
((.venv) ) [jlisowski@NV-JVQ3W94 mlir_perf]$ NUMBA_CUDA_DISABLE_LTO_OPT=1 python gemm_perf_mlir.py
GFlop/s        avg 1.21  min 1.21  max 1.21
```

### Opts enbaled
```
((.venv) ) [jlisowski@NV-JVQ3W94 mlir_perf]$ NUMBA_CUDA_DISABLE_LTO_OPT=0 python gemm_perf_mlir.py
GFlop/s        avg 57.61  min 57.47  max 57.69
```

This is roughly a 50x performance difference.

Most use cases are unaffected by the original correctness issue. The problematic path appears to be limited to float16/bfloat16 vector stores when linking with LTOIR and calling an LTOIR function around those stores. Types such as float32, float64, complex64, and complex128 appear to be unaffected.

Given this, it would be better to expose a user-controlled knob and clearly warn about the potential correctness issue, rather than unconditionally disabling all LTOIR optimizations. LTOIR libraries remain usable, and users can decide how to handle the correctness risk depending on their use case.

It should be noted that this LTOIR bug is only present on non-Blackwell architectures with f16, for which a fix will be incoming soon.

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/core/config.py
+++ b/src/numba_cuda_mlir/numba_cuda/core/config.py
@@ -459,6 +459,10 @@ class _EnvReloader:
         # Whether to generate verbose log messages when JIT linking
         CUDA_VERBOSE_JIT_LOG = _readenv("NUMBA_CUDA_VERBOSE_JIT_LOG", int, 1)
 
+        # WAR for an LTO linking bug that erases float16/bfloat16 (and their
+        # vector type) stores: when set, force opt_level=0 on LTO links
+        CUDA_DISABLE_LTO_OPT = _readenv("NUMBA_CUDA_DISABLE_LTO_OPT", int, 0)
+
         # Whether the default stream is the per-thread default stream
         CUDA_PER_THREAD_DEFAULT_STREAM = _readenv("NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM", int, 0)
 
--- a/src/numba_cuda_mlir/numba_cuda/core/config.py
+++ b/src/numba_cuda_mlir/numba_cuda/core/config.py
@@ -461,7 +461,7 @@ class _EnvReloader:
 
         # WAR for an LTO linking bug that erases float16/bfloat16 (and their
         # vector type) stores: when set, force opt_level=0 on LTO links
-        CUDA_DISABLE_LTO_OPT = _readenv("NUMBA_CUDA_DISABLE_LTO_OPT", int, 0)
+        CUDA_DISABLE_LTO_OPT = _readenv("NUMBA_CUDA_MLIR_DISABLE_LTO_OPT", int, 0)
 
         # Whether the default stream is the per-thread default stream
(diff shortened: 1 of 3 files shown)


src/numba_cuda_mlir/numba_cuda/core/config.py
```diff
@@ -588,12 +592,3 @@ def reload_config():
...
     _env_reloader.update()
-
-
-# use numba.core.config if available, otherwise use numba.cuda.core.config
-try:
-    import numba.core.config as _config
-
-    sys.modules[__name__] = _config
-except ImportError:
-    pass
```
Jlisowskyy, line 599: I wasn’t sure about this one, but when the numba package is installed, it overrides those variables. I’m also not sure whether this is the right place to parse the environment, but it made sense to me

src/numba_cuda_mlir/numba_cuda/core/config.py
```diff
@@ -459,6 +459,10 @@ def avx_default():
         # Whether to generate verbose log messages when JIT linking
         CUDA_VERBOSE_JIT_LOG = _readenv("NUMBA_CUDA_VERBOSE_JIT_LOG", int, 1)
 
+        # WAR for an LTO linking bug that erases float16/bfloat16 (and their
+        # vector type) stores: when set, force opt_level=0 on LTO links
+        CUDA_DISABLE_LTO_OPT = _readenv("NUMBA_CUDA_DISABLE_LTO_OPT", int, 0)
```
ZzEeKkAa: Is there MLIR specific settings section? We should have separate namespace if I understand correctly. (cc: @atmnp )

atmnp: `NUMBA_CUDA_MLIR` is the lengthier prefix we opted for, so that needs to be changed.

src/numba_cuda_mlir/numba_cuda/core/config.py
atmnp: Keeping this shouldn't be a problem for now, usually our CI environments and local dev environments don't have numba anyways, but I see why yours would. @gmarkall Are we good to remove these references yet?

atmnp (changes requested):

src/numba_cuda_mlir/numba_cuda/core/config.py
gmarkall: I think it's OK to remove these references. I think they're just a leftover / oversight remaining here.

src/numba_cuda_mlir/numba_cuda/core/config.py
Jlisowskyy: Sure, renamed

gmarkall (approved):

atmnp (approved):

PR #122, opened 2026-06-11
Fix f16 and vectorization

ZzEeKkAa:
## Overview

This PR addresses a silent data corruption bug caused by LLVM 7.0's `BasicAA` (Basic Alias Analysis) in the legacy `libNVVM` path. 

When Numba MLIR emits strictly typed vector shared/local memory allocations (e.g., `[0 x <4 x half>]`) and later performs stores with a different sub-vector view (e.g., `store <2 x half>`), the legacy LLVM 7.0 `BasicAA` pass incorrectly concludes they do not alias. This triggers Dead Store Elimination (DSE) to incorrectly drop the store during LTO, resulting in uninitialized memory/zeros. 

To resolve this, we adopt `nvcc`'s frontend strategy: **demoting shared and local array allocations to raw byte arrays (`i8`)** specifically for the LLVM 7.0 compilation path, which forces the compiler to treat all stores as potentially aliasing.

## Changes

- **Array Demotion Pass:** Introduced `_demote_shared_and_local_arrays_to_bytes` in `numba_cuda_mlir.optimization`. This MLIR pass intercepts `llvm.alloca` and `llvm.mlir.global` (with `addr_space=3`) and dynamically rewrites them to allocate plain byte blocks (`!llvm.array<N x i8>`).
- **Targeted Legacy Execution:** Updated `run_pre_codegen_patterns` to selectively run the array demotion pass and the integer storage workaround (`_resolve_shared_bit_storage_float_accesses`) **only** when targeting the legacy NVVM pipeline (`use_llvm70=True`). Modern architectures (Blackwell/`sm_100+`) will preserve their strong types and optimizations.
- **Robust Code Cleanup:** Grouped imports properly at the top of the file, resolved the `llvm.mul` argument mismatch by passing `IntegerOverflowFlags.none`, and ran `pre-commit` for consistent formatting.
- **Zero-Shape Edge Case Fix:** Modified `cuda_static_shared_memory` to gracefully handle `(0,)` dimensional tuples preventing downstream crashes.
- **Extensive Testing Suite:** Added comprehensive parameterized tests (`tests/test_lto_aliasing.py` and `test_float_aliasing.py`) to systematically assert proper LTO aliasing resolution for cross-views containing scalars, `vec2`, and `vec4` across `float16`, `float32`, and `float64` types.

## Impact

- **Bug Fix:** Resolves dropped local/shared memory stores when using 16-bit vectors in the legacy NVVM path.
- **Performance:** Bypassing strict types with `i8` causes no functional or performance penalty inside `llc`, as the backend will naturally re-vectorize scalar operations in the final PTX. Modern architectures skip this hack entirely.


src/numba_cuda_mlir/numba_cuda/core/config.py
```diff
@@ -339,13 +339,13 @@ def optional_str(x):
         DEBUG_INLINE_CLOSURE = _readenv("NUMBA_DEBUG_INLINE_CLOSURE", int, 0)
 
         # Force dump of LLVM IR
-        DUMP_LLVM = _readenv("NUMBA_DUMP_LLVM", int, DEBUG)
```
atmnp, line 342: Probably not necessary, same with below

ZzEeKkAa:
Converting it to DRAFT as I'm still chasing some unintended behavior locally

ZzEeKkAa:
That's conceptually wrong approach.

PR #127, opened 2026-06-12
[CI] Windows: drop test-stage devcontainer + use static CRT

leofang:
## Summary

- Tests run on the bare Windows host runner — no more `docker pull` of the devcontainer per matrix entry.
- Build LLVM, MLIR, and cext modules on Windows with `/MT` (static CRT) so the wheel has no `msvcp140` runtime dependency. No delvewheel repair needed; wheel is self-contained.
- Re-enable the 5 TCC matrix entries PR #81 deferred as untested.

## Other cleanups

- `ci/tools/env-vars` Windows path handling now matches cuda-python (`cygpath -w` `REPO_DIR` + `realpath` wrap on derived exports).
- Drop `restore_llvm_artifacts` + `LD_LIBRARY_PATH` from wheel-test jobs on both Linux and Windows; verified the wheel is self-contained (Linux pass/skip counts identical to main).
- Delete `ci/windows-run-tests-container.ps1`.

## Test plan

- [x] All 15 Windows test entries pass on the host runner.
- [x] Linux tests unchanged vs main.


leofang:
/ok to test [7769cbe](https://github.com/NVIDIA/numba-cuda-mlir/pull/127/commits/7769cbe173c1ade3dc5c1189d546b730734b989d)

atmnp (approved):
Looks great! Thank you for handling it/fixing the bugs inadvertently introduced by me. Surprised that it was just the /MT / multithreading option that I had overlooked to get it working.

leofang:
## Journey to the final solution

The simple change in this PR — drop the devcontainer at the Windows test stage — immediately exposed that the wheel wasn't actually self-contained on Windows: DLL init failures (`WinError 1114`) on the bare runner because the build's MSVC redist version didn't match the runner's system `msvcp140`.

The canonical pip-ecosystem fix for this is *wheel repair* — auditwheel on Linux, delvewheel on Windows. Repair tools walk the .pyd import tables, find the missing redistributable DLLs (msvcp140, libstdc++, etc.), bundle them into the wheel with content-hash-mangled names for cross-wheel isolation, and patch the .pyd's import table to use the mangled name. Most native-Python projects ship Windows wheels this way (cuda-bindings, cuda-core, numpy, scipy, etc.). It seemed like the obvious lever to reach for, so that's where we started.

*Ten* iterations of trying to make delvewheel work before pivoting to fix the root cause:

1. *Wire delvewheel into the workflow with `--add-path`* — failed: delvewheel wasn't even installed in cibuildwheel's env.
2. *Move delvewheel config to `pyproject.toml`* (matching cuda-python) — failed: `FileNotFoundError: nanobind-numba_cuda_mlir.dll`. delvewheel doesn't see in-wheel DLLs.
3. *Add `--add-path` pointing at the LLVM install dir* — failed: shell variable expansion didn't reach delvewheel under cmd.exe.
4. *Switch to repo-relative `--add-path`* — failed: same env-var path problem.
5. *Substitute `${{ env.VAR }}` at workflow render time (in `build-wheel.yml`)* — found paths were MSYS-style `/d/a/...`, delvewheel needs Windows-native.
6. *`cygpath -w REPO_DIR` in `env-vars`* — failed: bash silently ate backslashes from the resulting `D:\a\...` paths when passed unquoted to downstream `cygpath -w`.
7. *Revert env-vars + apply `cygpath -w` inline in delvewheel cmd* — build worked; tests crashed with access violations. delvewheel-mangled `MLIRPythonCAPI` + setup.py-staged un-mangled copy ⇒ two images in process ⇒ static-state desync.
8. *Mirror cuda-python: `cygpath -w` + `realpath` wrap (in `env-vars`)* — same access violation, same duplicate-image cause.
9. *`--exclude` project DLLs from delvewheel* — avoided duplicates, but `msvcp140` got mangled to a name the un-mangled `MLIRPythonCAPI` couldn't find.
10. *`--ignore-existing --analyze-existing --with-mangle`* — hit an actual delvewheel bug: `KeyError: 'msvcp140-<hash>.dll'` in its double-patching loop.

*The pivot*: stepped back, asked why delvewheel was needed at all. Answer: only to bundle `msvcp140` because everything was built with `/MD` (dynamic CRT). Built LLVM 7, modern LLVM/MLIR, and the cext modules with `/MT` (static CRT) instead — each binary embeds its own C++ runtime, the wheel has zero `msvcp140` dependency, delvewheel becomes a no-op, the "two competing bundling pipelines" problem (setup.py wholesale-stages the MLIR bindings tree _and_ delvewheel re-bundles) dissolves. *The first commit on the static-CRT track (`99d354c`) made every Windows test pass on the first CI run.*

Lessons:
- *Reach for the build-system fix before the wheel-repair fix when you control the build*. Static CRT was a 3-line cmake-flag change that obsoleted ~9 commits of delvewheel plumbing. delvewheel is the right answer when you can't change how upstream DLLs are linked — but we _do_ control how LLVM is built here.
- *delvewheel's design assumes the wheel doesn't pre-stage DLLs*. cuSIMT's setup.py wholesale-copies the MLIR Python bindings tree, which violates that assumption; the combination has corner cases delvewheel doesn't handle well.
- *When debugging keeps looping on the same component, suspect the component is the wrong tool for the job*.

PR #128, opened 2026-06-12
[Lowering] Convert to correct integer/float type for power operator

### LINKED ISSUE #123: The power operator `**` lowers to ops with mismatched operand types (MLIR verification fails)
## Summary

Using the Python power operator `**` in device code produces an MLIR module that
fails verification, because the generated power lowering emits an op whose
operands/results do not all share a type:

- **Integer** base: `arith.cmpi' op requires all operands to have the same type`
  (in the exponent-sign check of the integer-power expansion).
- **Floating-point** base: `math.powf' op requires the same type for all operands
  and results`.

## Environment

- numba-cuda-mlir 0.3.0
- Python 3.12, Linux

## Reproducer

Integer power:

```python

def square_i(a):
    return a ** 2

cuda.compile(square_i, T.int32(T.int32), device=True, abi="c",
             abi_info={"abi_name": "square_i"}, output="ltoir")
```

→ `InternalCompilerError: Invalid MLIR module: Verification failed: 'arith.cmpi'
op requires all operands to have the same type` (with a nearby
`cf.assert "negative exponent not supported for integer exponentiation"`).

Floating-point power:

```python

def square_f(a):
    return a ** 2

cuda.compile(square_f, T.float32(T.float32), device=True, abi="c",
             abi_info={"abi_name": "square_f"}, output="ltoir")
```

→ `InternalCompilerError: Invalid MLIR module: Verification failed: 'math.powf'
op requires the same type for all operands and results`.

## Expected vs actual

- **Expected:** `a ** 2` compiles; the power lowering coerces the exponent /
  result to the base type (or otherwise emits type-consistent ops).
- **Actual:** the lowering mixes types (e.g. an `i32`/`i64` mismatch in the
  integer path's `cmpi`, or a `f32`/`f64` mismatch in `math.powf`), so the
  module fails verification.

atmnp:
This should fix #123.


tests/numba_cuda_tests/cudapy/test_compiler.py
```diff
@@ -738,6 +738,18 @@ def div_by_2(x):
         sig = types.complex128(types.complex128)
         cuda.compile(div_by_2, sig, device=True, abi="c")
 
+    def test_compile_power_operator_c_abi(self):
+        # Reproducer from gh-123
+        # https://github.com/NVIDIA/numba-cuda-mlir/issues/123
```
VijayKandiah: Do we want to include pointers to github issues in our reproducer tests? I would avoid mentioning it, the test seems simple enough.

VijayKandiah (changes requested):
LGTM but I'd avoid the comment pointing to github issue in the test file.

tests/numba_cuda_tests/cudapy/test_compiler.py
atmnp: Yeah it was unclear to me, I seem to remember numba/numba-cuda doing it? @gmarkall I've removed it for this one.

VijayKandiah (approved):
LGTM now

PR #129, opened 2026-06-12
[Lowering] Fix storage/value distinction for complex values lowered through CPointers

### LINKED ISSUE #120: A `complex` value loaded through a `CPointer` fails to lower in a `device`/`abi="c"` function
## Summary

A device function that takes a complex value **by value** compiles fine, but one
that loads a complex value **through a `CPointer`** (`ptr[0]`) fails to compile:
the MLIR `complex<f64>` produced by the pointer load survives into the
`gpu.module` and the module fails verification / cannot be lowered for the NVVM
target.

This blocks the "C-ABI wrapper" pattern used by CCCL,
where a user operator is wrapped as `void f(void* in0, void* in1, void* out)`
and the inputs are typed as `CPointer(complex...)`.

## Environment

- numba-cuda-mlir 0.3.0
- Python 3.12, Linux

## Reproducer

Works (complex passed by value):

```python
import numpy as np

def cmp_c(lhs, rhs):
    return np.uint8(lhs.real < rhs.real)

# OK:
cuda.compile(cmp_c, (T.complex128, T.complex128), device=True, abi="c",
             abi_info={"abi_name": "cmp_c"}, output="ltoir")
```

Fails (same logic, but complex loaded through a `CPointer`):

```python
import numpy as np

def add_real(xp, yp, rp):
    # xp, yp : CPointer(complex128) ; rp : CPointer(int8)
    rp[0] = np.uint8(xp[0].real < yp[0].real)

sig = T.void(T.CPointer(T.complex128), T.CPointer(T.complex128), T.CPointer(T.int8))
cuda.compile(add_real, sig, device=True, abi="c",
             abi_info={"abi_name": "add_real"}, output="ltoir")
```

This raises `UserFacingInternalCompilerError` ("Invalid MLIR module: Verification
failed"). The emitted op contains:

```mlir
%2 = "llvm.load"(%1) : (!llvm.ptr) -> complex<f64>
%6 = "complex.re"(%2) : (complex<f64>) -> f64
```

i.e. the pointer load yields an MLIR `complex<f64>` (and a `complex` dialect op)
inside the `gpu.module`, which is not legal/lowerable for the NVVM target.

## Expected vs actual

- **Expected:** loading a `complex` value through a `CPointer` and operating on
  it lowers the same way as receiving the complex by value (which works).
- **Actual:** the `complex<f64>` from the pointer load is not lowered to a
  target-legal representation, and module verification fails.

## Note

A plain `@cuda.jit` *kernel* that does `out[i] = a[i].real < b[i].real` over
complex arrays works, so the gap is specific to the `device` / `abi="c"` (LTOIR)
path with a `CPointer(complex)` argument.

## Impact

Complex-typed user operators cannot be used through the C-ABI void* wrapper
convention (custom reduce/scan/comparators over `complex64`/`complex128`).

atmnp:
This should close #120.


atmnp:
Seems to have been resolved.

PR #130, opened 2026-06-12
Avoid duplicate optimization in public compile helpers

fallintoplace:
## Summary
- Let MLIRDispatcher.compile() remain the single owner of optimization and cache registration.
- Stop the public compile helper from re-running optimize() on dispatcher results.
- Add a regression test using a cached-style compile result with optimized artifacts but no raw MLIR module.

## Why
Public helpers such as compile_cubin() route through MLIRDispatcher.compile(), which already optimizes on cache misses and returns cached optimized artifacts on hits. Re-running optimize() from compiler._compile() can double-optimize fresh results and fails on cached results that do not carry the live mlir_module.

## Validation
- python3 -m py_compile src/numba_cuda_mlir/compiler.py tests/test_compiler_public_api.py
- uvx ruff check src/numba_cuda_mlir/compiler.py tests/test_compiler_public_api.py
- uvx ruff format --check src/numba_cuda_mlir/compiler.py tests/test_compiler_public_api.py
- git diff --check --cached

Could not run pytest locally on macOS arm64 because cuda-bindings has no compatible wheel for this platform, and importing the package from source also requires the built _mlir extension.


atmnp (approved):

fallintoplace:
Something is failing, let me have a look.

PR #131, opened 2026-06-12
Fix comparators not working with unsigned int types

### LINKED ISSUE #121: Integer comparisons ignore operand signedness (signed compared as unsigned, and vice versa)
## Summary

Integer comparison operators (`<`, `>`, `<=`, `>=`) in device code do not respect
the operand's signedness. Both directions are wrong:

- **Unsigned operands are compared as signed** — values with the high bit set
  (≥ 2^(N-1)) compare as negative.
- **Signed operands are compared as unsigned** — negative values compare as
  large positive.

## Environment

- numba-cuda-mlir 0.3.0
- Python 3.12, Linux

## Reproducers

Unsigned compared as signed:

```python
import numpy as np, cupy as cp

@cuda.jit
def lt(out, a, b):
    i = cuda.grid(1)
    if i < out.size:
        out[i] = a[i] < b[i]

a = cp.asarray(np.array([40000, 5], dtype=np.uint16))
b = cp.asarray(np.array([5, 40000], dtype=np.uint16))
out = cp.zeros(2, dtype=np.bool_)
lt[1, 2](out, a, b)
print(out.get())   # [ True False ]  -- should be [False  True]
```

`40000` (uint16) is treated as the signed `int16` value `-25536`. Same for
`uint8(200)` and `uint32(3_000_000_000)`.

Signed compared as unsigned:

```python
import numpy as np, cupy as cp

@cuda.jit
def gt0(out, a):
    i = cuda.grid(1)
    if i < out.size:
        out[i] = a[i] > 0

a = cp.asarray(np.array([-50, -1, 0, 1, 50], dtype=np.int32))
out = cp.zeros(5, dtype=np.bool_)
gt0[1, 5](out, a)
print(out.get())   # [ True  True False  True  True ]  -- should be [F F F T T]
```

The negative `int32` values `-50` and `-1` compare as `> 0`, i.e. they are
treated as large unsigned values.

## Expected vs actual

- **Expected:** comparisons use the signed predicate for signed integer types
  and the unsigned predicate for unsigned integer types.
- **Actual:** the emitted compare predicate does not match the operand
  signedness, so results are wrong for the affected value ranges.

## Impact

Any device operator that compares integers is silently wrong for the affected
ranges — sort comparators over unsigned keys, predicates like `x > 0` over
signed data, etc. Small/positive test data masks the bug.

acosmicflamingo:
Resolves #121 

It looks like the `ge/gt/le/lt` comparator operators do not work when an unsigned int type is large enough (e.g. `40_000 > 5` returns `False` as @shwina had found in the ticket). This PR addresses the issue 😃

It also fixes an import issue that currently resides in `tests/numba_cuda_tests/cudapy/test_operator.py`.


tests/numba_cuda_tests/cudapy/test_operator.py
```diff
@@ -341,13 +388,49 @@ def test_multiple_float16_comparisons_false(self):
...
+            pytest.param(simple_fp16_ne, "setp.ne.b16", id="ne"),
+        ],
+    )
+    def test_int16_comparison_ptx(self, func, opstring):
+        args = (b1[:], i2, i2)
+        compiled = cuda.jit("void(b1[:], i2, i2)", lto=True)(func)
+        ptx = compiled.inspect_lto_ptx(args)
+        assert opstring in ptx, f"{opstring} not in PTX"
+
+    @pytest.mark.xfail(True, reason="NVVM verify error")
```
atmnp: Can you explain these xfails? Do we know why we get them?

acosmicflamingo: Oh you're right! I don't need this for my tests after all; just verified that they pass as-is. I saw that lots of the others tests were using them, incorrectly thought all PTX tests currently had them, and figured that I needed it in my own test. I will remove it now; thanks!

acosmicflamingo: As to why the others have them, I have not looked into it but I could make that my next task.

acosmicflamingo:
Fascinating...locally I get successes:
```
tests/numba_cuda_tests/cudapy/test_operator.py::TestOperatorModule::test_int16_comparison_ptx[gt] PASSED [ 16%]
tests/numba_cuda_tests/cudapy/test_operator.py::TestOperatorModule::test_int16_comparison_ptx[ge] PASSED [ 33%]
tests/numba_cuda_tests/cudapy/test_operator.py::TestOperatorModule::test_int16_comparison_ptx[lt] PASSED [ 50%]
tests/numba_cuda_tests/cudapy/test_operator.py::TestOperatorModule::test_int16_comparison_ptx[le] PASSED [ 66%]
tests/numba_cuda_tests/cudapy/test_operator.py::TestOperatorModule::test_int16_comparison_ptx[eq] PASSED [ 83%]
tests/numba_cuda_tests/cudapy/test_operator.py::TestOperatorModule::test_int16_comparison_ptx[ne] PASSED [100%]
```

And yet, CI fails. I suppose this is why those expected failure markers were being used. As to why, I do not know 😝 I'll try to see what's going on.

acosmicflamingo:
I see, some of those were a false negative; I forgot to remove the `pytest.mark.xfail` from `test_uint16_comparison_ptx`. However, the question with the failing `int_comparson_ptx` tests do remain...I wonder why CI gives me signed `eq/ne` operators when I don't get them locally...

~~@atmnp can you run "/ok to test a4fbc10444335d34adba6ad36d9bcc8c6407b14c" when you have a chance so I can see the exact tests I need to debug?~~

acosmicflamingo:
I suppose b16 is interchangeable with s16 and u16 based on context: https://docs.nvidia.com/cuda/archive/12.2.1/nvvm-ir-spec/index.html#parameter-passing-and-return. Tried to find in CTK release notes something that could explain this behavior (I have a newer toolkit installed locally), but didn't have much success.

@atmnp can you run "/ok to test 4c7bf4dd94d0dfee50cc652367cff1a2939f69ed" when you have a chance? Hopefully tests pass this time. Also, do you think it'd be worth the time to try and find out why in CI we get `s16` or `u16` while I get `b16` locally?

tests/numba_cuda_tests/cudapy/test_operator.py
```diff
@@ -1,16 +1,19 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
+import itertools
+import operator
+
```
atmnp: This whole block is churn, we don't need to be doing this many line edits for no purpose other than re-arranging, makes it harder to review.

tests/numba_cuda_tests/cudapy/test_operator.py
```diff
@@ -77,12 +80,12 @@ def simple_fp16_ne(ary, a, b):
     ary[0] = a != b
```
atmnp, line 80: This is helpful tho! Cleans up some of the files.

tests/numba_cuda_tests/cudapy/test_operator.py
```diff
@@ -112,17 +115,17 @@ def multiple_hcmp_5(r, a, b, c):
     r[0] = a < b and c >= b
 
 
-class TestOperatorModule(NumbaCUDATestCase):
+class TestOperatorModule:
```
atmnp, line 116: This makes me a little nervous, why was this necessary?

atmnp:
PTX spelling varies after LTO by target/toolchain, I think it's fine how it is, but see how it's passing on the rtxpro6000? It means that the tests that failing on the others are written in a way that's specific to that generation, so those tests should probably be specialized or generalized (depending on your take) to work on both.

tests/numba_cuda_tests/cudapy/test_operator.py
acosmicflamingo: My apologies; didn't know whether rearranging imports to be grouped by built-ins first, then 3rd party, etc was welcomed. I agree that it does make it harder to navigate reviews; I'll keep this in mind going forward (and sort it to how it was before). Thanks!

tests/numba_cuda_tests/cudapy/test_operator.py
acosmicflamingo: I get the sense from tickets I've seen in CuPy, numba-cuda, etc. that there is an eventual goal to migrate tests from `unittest` to `pytest` and `NumbaCUDATestCase` is a subclass of `unittest.TestCase`. If I had subclassed `TestOperatorModule` from `NumbaCUDATestCase`, then the `python -m pytest` command I use doesn't work the way I'd like.

In addition, the tests for `TestOperatorModule` do not use any of the parent functions like `assertFileCheckMatches`, `assertPreciseEqual`, etc. It would also break a whole ton of tests trying to decouple `unittest.TestCase` from `NumbaCUDATestCase`, so this seemed like the safe approach (and I just checked again that the `super().setUp()` is calling `unittest.TestCase` and not a function within `NumbaCUDATestCase` class).

Lastly, I think what would be ideal is that functions within `NumbaCUDATestCase` eventually become pytest fixtures that other tests can import. But given how much work that involves, it's one of those things where we can just take it one step at a time instead of create some monolithic PR that someone would have to sift through hehehe

tests/numba_cuda_tests/cudapy/test_operator.py
acosmicflamingo: 🕺🕺🕺

acosmicflamingo:
@atmnp yes good point about it passing in the rtxpro6000 test. Interesting too that now only the unsigned integer PTX test fails for `ne/eq`. I just pushed a commit that (in addition to reverting the importing order) re-adds PTX to the assertion output so I can see what's going on. Depending on what I find, I might have to generalize the test. I wouldn't be worried about this because the `test_int16_comparison` and `test_uint16_comparison` functions are what actually catch the computation errors.

tests/numba_cuda_tests/cudapy/test_operator.py
atmnp: Aha, wise guy with all the references. You're right! I did forget about that eventual migration plan.

acosmicflamingo: LOL I gotta provide receipts to build some street cred around here 😜

acosmicflamingo:
Wow, so I'm getting a `setp.eq.s16` even in the PTX test that explicitly uses unsigned integers. ~~@atmnp I'd like to get your opinion :) do think the PTX tests are even necessary when all we really need are the comparator tests that actually assert correct values? It seems silly that I check to see that any of the strings (e.g. for `eq` I'd check for `["setp.eq.s16", "setp.eq.b16", "setp.eq.u16"]` or `setp.eq`) are in the PTX code.~~

acosmicflamingo:
@atmnp I just pushed a commit that simplified the PTX assertions; code could always be improved in the future if it has to. Can you run "/ok to test 6089df9a61cd30af21b27787428b0656d188ca11" when you have a chance?

acosmicflamingo:
I'm having a very difficult time reproducing the `tests/numba_cuda_tests/cudapy/test_cooperative_groups.py::TestCudaCooperativeGroups::test_max_cooperative_grid_blocks` issue locally on my Windows machine with a Blackwell GPU, and it seems like this test hasn't failed in CI for anyone else's PRs. Since I don't have the ability to execute `/ok to test`, I pushed two commits that hopefully help us narrow down what is going on the next time it runs.

 @atmnp sorry to ping you so frequently; can you run "/ok to test 6186b7df4869e494f01ab112bfe8f7144399e348" when you're free?

acosmicflamingo:
Super! The Windows issues are no longer present :)

atmnp (approved):
Looks great! Thank you so much for looking into it. Ready to merge whenever you want, sorry it took a bit.

acosmicflamingo:
Hooray! I appreciate that you took the time to even review it; no worries on time 🙇🏽 Thanks! You can merge it whenever you'd like. I can't since I don't have those privileges 🥺

PR #132, opened 2026-06-12
[CI] Strip LLVM .so files at install time

leofang:
## Summary

Linux wheels currently ship un-stripped `.so` files — `file(1)` reports `not stripped` for `libMLIRPythonCAPI.so` and friends. The build is Release without `-g` (no DWARF debug info), but the static symbol table (`.symtab` + `.strtab`) is preserved by default at install time. Stripping it saves ~76 MB uncompressed across the LLVM `.so` files in the wheel (~20-25 MB compressed).

Breakdown (`strip --strip-unneeded`):

| File | Before | After | Saved |
|---|---|---|---|
| `libMLIRPythonCAPI.so` | 231.7 MB | 162.2 MB | 69.5 MB |
| `libMLIRPythonSupport-numba_cuda_mlir.so` | 2.6 MB | 1.4 MB | 1.2 MB |
| `libLLVM-7.so` | 37.3 MB | 32.3 MB | 5.0 MB |
| `libMLIRToLLVM70.so` / `libnanobind-*.so` | < 0.1 MB savings each | | |
| **Total** | | | **~76 MB** |

## Changes

- `ci/build-llvm-modern.sh`: pass `--strip` to `cmake --install`.
- `ci/build-llvm7.sh`: this script doesn't `cmake --install` (it copies the LLVM dylib directly out of the build tree), so add an explicit `strip --strip-unneeded` on the copied `.so`.

Windows is not affected: MSVC stores debug info out-of-line in `.pdb` files, so `.dll` files don't carry an embedded symbol table to begin with — the Windows wheel is already at the minimum size.

## Test plan

- [ ] CI build succeeds on Linux.
- [ ] Linux Python tests still pass (stripping `--strip-unneeded` keeps dynamic symbols needed at runtime, so functionality should be unchanged).
- [ ] Verify wheel size shrinks by ~20-25 MB.

### DIFF
--- a/ci/build-llvm-modern.sh
+++ b/ci/build-llvm-modern.sh
@@ -80,9 +80,12 @@ cmake_args=(
 
 cmake "${cmake_args[@]}"
 
-# Build & install
+# Build & install. --strip removes the static symbol table (.symtab +
+# .strtab); LLVM is built without -g so there's no DWARF to strip, but
+# the symbol table itself accounts for ~70 MB on libMLIRPythonCAPI.so
+# alone -- ~25 MB compressed in the wheel.
 cmake --build "${LLVM_BUILD}" -j "${PARALLEL}"
-cmake --install "${LLVM_BUILD}"
+cmake --install "${LLVM_BUILD}" --strip
 
 echo "=== Modern LLVM installed to ${LLVM_INSTALL} ==="
 echo "  MLIR_DIR=${LLVM_INSTALL}/lib/cmake/mlir"
--- a/ci/build-llvm7.sh
+++ b/ci/build-llvm7.sh
@@ -39,16 +39,10 @@ command -v sccache &>/dev/null || { echo "ERROR: sccache not found"; exit 1; }
 export CMAKE_C_COMPILER_LAUNCHER="$(which sccache)"
 export CMAKE_CXX_COMPILER_LAUNCHER="$(which sccache)"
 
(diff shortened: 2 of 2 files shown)


leofang:
## Wheel-size impact

Verified the strip changes on the Linux py3.11 wheel:

- **Wheel: 82 MB → 68 MB** (-14 MB, ~17%)
- `libMLIRPythonCAPI.so`: 232 MB → 163 MB (-69 MB uncompressed)
- `libLLVM-7.so`: 38 MB → 33 MB (-5 MB uncompressed)
- `libMLIRPythonSupport-numba_cuda_mlir.so`: 2.7 MB → 1.5 MB (-1.2 MB uncompressed)

Same shrink applies to the other Python versions and to aarch64.

## Projected vs actual

Going in I projected 20–25 MB of compressed savings, based on a generic ~3× compression ratio for code. Actual compressed savings landed at **14 MB**.

The discrepancy: the ~76 MB of uncompressed savings is dominated by `.strtab` (~58 MB of mangled C++ symbol names) and `.symtab` (~10 MB). `.strtab` has massive prefix redundancy across mangled names and compresses closer to ~5×, much better than typical code. Reconciling: 58/5 + 10/2.5 ≈ 14 MB, which matches the observed delta.

The 3× rule of thumb was wrong for symbol-table-heavy content. Net result is smaller than projected but still real and worth landing.

atmnp (approved):

PR #134, opened 2026-06-13
Revert "Bump the actions-monthly group across 1 directory with 5 updates"

leofang:
Reverts NVIDIA/numba-cuda-mlir#125

The Windows build failed with the latest cibuildwheel, but we did not update the status check to include Windows pipelines, so automerge ignored the failures and merged away.


leofang:
Admin-merge to keep the main branch healthy.

leofang:
> The Windows build failed with the latest cibuildwheel, but we did not update the status check to include Windows pipelines, so automerge ignored the failures and merged away.

Turns out it's way more subtle than I originally anticipated, see RCCA: https://github.com/NVIDIA/cuda-python/issues/2208

PR #140, opened 2026-06-15
[Lowering] Add support for carray

### LINKED ISSUE #124: No way to build a device array (`carray`) from a raw pointer in device code
## Summary

There is no device-code equivalent of `numba.carray` — i.e. no way to construct
a typed device `Array` view from a raw pointer (plus shape/strides) inside a
`@cuda.jit` / device function. As a result, a `CPointer` argument cannot be used
with array operations: `cuda.atomic.*`, `len()`, `.shape`, `.ndim`, slicing,
etc. all require a real `Array`.

This blocks accelerated libraries that receive opaque `void*`/pointer arguments
(e.g. operator state captured as device arrays) from reconstructing them as
arrays for use inside user operators.

## Environment

- numba-cuda-mlir 0.3.0
- Python 3.12, Linux

## Reproducer

```python

def bump(p):
    # p : CPointer(int32) -- a raw pointer to a 1-element counter
    cuda.atomic.add(p, 0, 1)

cuda.compile(bump, T.void(T.CPointer(T.int32)), device=True, abi="c",
             abi_info={"abi_name": "bump"}, output="ltoir")
```

→ `AttributeError: 'CPointer' object has no attribute 'ndim'` (from `atomic.add`
typing). The same happens for `len(p)`, `p.shape`, etc.

`numba.carray` is not available (`numba_cuda_mlir.cuda.carray` /
`numba_cuda_mlir.carray` do not exist), so there is no way to turn `p` into an
`Array(int32, 1)` view in device code.

## Expected

A device-callable `carray(ptr, shape, dtype=...)` (or equivalent) that yields a
typed `Array` view over the pointed-to memory, usable with `atomic.*`, `len`,
`.shape`, indexing, etc. — matching `numba.cuda`'s `carray`.

## Impact

Operators that perform array operations on a pointer/captured-array argument
cannot be compiled. In our integration this affects stateful operators whose
captured device arrays are used with `cuda.atomic.add` (counting/side-effect
ops) or `len()` — element indexing (`state[i]`) works, but any array operation
does not.

atmnp:
This closes #124.

### DIFF
--- a/src/numba_cuda_mlir/__init__.py
+++ b/src/numba_cuda_mlir/__init__.py
@@ -3,8 +3,8 @@
 
 from numba_cuda_mlir._version import __version__
 from numba_cuda_mlir.mlir import make_nanobind_metaclass_inheritable
-from numba_cuda_mlir.numba_cuda.np.numpy_support import carray  # noqa: F401
+from numba_cuda_mlir.numba_cuda.np.numpy_support import carray, farray  # noqa: F401
 
 make_nanobind_metaclass_inheritable()
 
-__all__ = ["cuda", "carray", "__version__"]
+__all__ = ["cuda", "carray", "farray", "__version__"]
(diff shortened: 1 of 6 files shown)


VijayKandiah (approved):
LGTM

PR #142, opened 2026-06-16
ci: adopt CCCL check_result aggregator and add doc-only PR mode

leofang:
## Summary

Mirrors the gating-check fix in NVIDIA/cuda-python#2210 (root-cause analysis: NVIDIA/cuda-python#2208) and bundles in cuda-python's `[doc-only]` PR-mode feature, which had not been propagated to this repo.

## Changes

1. **`checks:` aggregator** — replaced the `cancelled || failure` predicate with CCCL's `check_result` pattern: per-dep `expected="success"`, `expected="skipped"` for tests in `[doc-only]` mode, early short-circuit for `[no-ci]`. The previous predicate silently passed when an upstream `build-*` failure cascaded as `'skipped'` on the downstream tests — the same bug that bit this repo in #125 (auto-merged with failing Windows builds, reverted in #134).

2. **`[doc-only]` PR mode** — new `should-skip.outputs.doc-only`, gated into all 11 `test-*` jobs and the non-linux-64 `build-*` jobs. The linux-64 build chain (`build-llvm-linux-64`, `build-linux-64`) and `build-docs` intentionally stay ungated because `build-docs` needs the linux-64 wheel.

3. **`should-skip` itself** is now in the aggregator's `needs:` and explicitly checked, so a crash in the gating job can't masquerade as a green CI run.

## Verification

Local simulation of the new aggregator across 8 scenarios (normal-green, build-fail-cascade, `[no-ci]`, `[doc-only]`-OK, `[doc-only]`-doc-fails, build-windows-fail-cascade, single-test-fail, should-skip-itself-fails) — all behave as expected. End-to-end CI evidence is in the comment thread on this PR.

## What is NOT changed

- `detect-changes` / path-based gating from cuda-python — cuSIMT's module structure differs; if wanted, that's a separate PR with its own design.
- Satellite test-workflow consolidation (separate ongoing concern).
- GitLab CI mirror.

## Reference

- Source-of-truth fix in cuda-python: NVIDIA/cuda-python#2210
- Root cause: NVIDIA/cuda-python#2208
- CCCL gate pattern: https://github.com/NVIDIA/cccl/blob/8c0e6cb1b6412d7bd0070f9ac3f55fa80231961a/.github/workflows/ci-workflow-pull-request.yml#L463-L526

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -115,7 +115,7 @@ jobs:
         uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756  # v1
 
       - name: Build numba-cuda-mlir wheel
-        uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
+        uses: pypa/cibuildwheel@294735312765b09d24a2fbec22660ce817587d55  # v4.1.0
         with:
           package-dir: .
           output-dir: dist
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -145,6 +145,11 @@ jobs:
             --exclude libLLVM-7.so &&
             "/host/${{ env.SCCACHE_PATH }}" --show-adv-stats &&
             "/host/${{ env.SCCACHE_PATH }}" --show-stats --stats-format=json > /host/${{ github.workspace }}/sccache_NUMBA_CUDA_MLIR.json
+          # Windows wheels are built with /MT (static CRT) per PR #127, so the
+          # wheel is self-contained and needs no DLL bundling. cibuildwheel v4
+          # made `delvewheel repair` the Windows default; opt out explicitly
+          # to restore the v3.x no-op behavior we rely on.
+          CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: ""
 
       - name: Report sccache stats (numba-cuda-mlir)
         if: ${{ startsWith(inputs.host-platform, 'linux') }}
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Install uv
-        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b  # v8.1.0
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39  # v8.2.0
         with:
           enable-cache: false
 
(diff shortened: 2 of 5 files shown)


.github/workflows/build-wheel.yml
```diff
@@ -115,7 +115,7 @@ jobs:
         uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756  # v1
 
       - name: Build numba-cuda-mlir wheel
-        uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
+        uses: pypa/cibuildwheel@294735312765b09d24a2fbec22660ce817587d55  # v4.1.0
```
leofang, line 118: This would trigger a build error on Windows, as `delvewheel repair` (which we don't need) becomes the default since v4.

leofang:
### What this PR validates

Two interlocking features exercised end-to-end on this branch:

**1. `[doc-only]` PR mode** — run https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/27591192624 (auto-triggered when the PR title still contained the literal `[doc-only]` substring): `build-llvm-linux-64` / `build-linux-64` / `build-docs` ran; `build-llvm-{aarch64,windows}`, `build-{aarch64,windows}`, all 11 `test-*` jobs skipped via the new `if: && !fromJSON(needs.should-skip.outputs.doc-only)` gate; aggregator's per-dep `expected="skipped"` path → `Check job status = success`.

**2. Gating-check fix (CCCL `check_result` pattern)** — run https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/27592793126 (after the title was renamed to drop the literal trigger, and dependabot's `cibuildwheel v3.4.1 → v4.1.0` bump from #125 was cherry-picked back in): 5 `Build win-64` matrix entries fail at `delvewheel repair`, `Test win-64` cascades to `skipped`, **`Check job status = failure`**.

Side-by-side with the original incident:

| | Reproducer run | `Check job status` |
|---|---|---|
| **Pre-fix** — #125 (auto-merged on green) | https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/27448129517 | success (the bug) |
| **Post-fix** — this branch | https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/27592793126 | failure (fix working) |

Same input, opposite verdict. The skipped-as-success hole on cuSIMT is closed.

The underlying cibuildwheel v4.1.0 delvewheel break will be fixed shortly.

.github/workflows/build-wheel.yml
leofang: Verified, see https://github.com/NVIDIA/numba-cuda-mlir/pull/142#issuecomment-4715014300. Fixing in commit 1cd8652.

leofang:
(Fixed the PR description. Claude seems to have a race condition when running in parallel, and the content was leaking from another session 😱 This PR is now ready for review/merge.)

atmnp (approved):

PR #145, opened 2026-06-20
ENH: Add support for inspect_llvm

### LINKED ISSUE #95: [FEA] Support `inspect_llvm()` as in Numba-CUDA
One use case of Numba-CUDA is to decorate functions with `@jit` and then call `inspect_llvm()` to be able to get the LLVM IR for feeding into another (possibly non-Python) compilation pipeline.

To support this use case it would be desirable to add the `inspect_llvm()` method to dispatchers in Numba-CUDA-MLIR.

Numba-CUDA's `inspect_llvm()` always produces IR with typed pointers. However, it is more natural in the context of Numba-CUDA-MLIR to produce IR with typed pointers for pre-Blackwell architectures and with opaque pointers for Blackwell onwards. The conclusion from discussions is that this discrepancy should be accommodated by:

- Producing IR with opaque pointers for Blackwell onwards by default.
- Allowing typed pointer IR to be produced even for Blackwell with a keyword argument on the `inspect_llvm()` method.

This provides the best balance of outcomes:

- The "natural" choice for Numba-CUDA-MLIR is used by default.
- The backward-compatible choice can be opted-in at the same time users are making another code change (to modify their imports to use Numba-CUDA-MLIR) - adding a keyword argument at the same time is the point at which there is least friction for making this change to their code. 

See also: NVIDIA/numba-cuda#890 (we would want to ensure our behaviour is correct too).

acosmicflamingo:
Resolves #95

This PR allows users to use the `inspect_llvm` function to retrieve LLVM-IR code from a JIT-compiled function. There actually wasn't too much involved because the library has already done lots of the heavy-lifting. To be specific, when an architecture's compute capable version is above 10, LLVM-IR is automatically generated during the compilation phase.

Much of the PR is actually handling the instance where the architecture pre-dates Blackwell. Actually, it turns out that for older architectures, PTX is automatically generated from the MLIR input, while more modern architectures has the MLIR -> LLVM IR -> PTX flow and for that reason, we didn't have any LLVM-IR code to cache. However, it turns out that there is a `translateToNVVMIR` function that can be leveraged in `cext/mlir-llvm70/lib/CAPI.cpp` to give us the LLVM-IR we seek.

Additionally, the older LLVM-IR will never need to be used anywhere outside of `inspect_llvm`, so there's no need to try and generate this code in the compilation state for users just so it’s cached for future use.

I also generated a basic test (very similar to an upstreamed version) that asserts that checks for a mangled name being present in the LLVM-IR code.

### DIFF
--- a/cext/mlir-llvm70/lib/CAPI.cpp
+++ b/cext/mlir-llvm70/lib/CAPI.cpp
@@ -41,7 +41,7 @@ int llvm70_translate_gpu_module_from_op(
     void *raw_op,
     const char *chip, const char *data_layout,
     const char *libllvm, const char *libnvvm, const char *libdevice,
-    int gen_lto, int opt_level, int gen_lineinfo,
+    int gen_lto, int gen_llvmir, int opt_level, int gen_lineinfo,
     char **out, size_t *out_len, char **err_out) {
 
   *out = nullptr;
@@ -87,7 +87,9 @@ int llvm70_translate_gpu_module_from_op(
   llvm::install_fatal_error_handler(fatalErrorHandler, nullptr);
 
   try {
-    auto ptxOrErr = llvm70::translateToPTX(gpuMod, opts);
+    auto ptxOrErr = gen_llvmir != 0
+      ? llvm70::translateToNVVMIR(gpuMod, opts)
+      : llvm70::translateToPTX(gpuMod, opts);
     llvm::remove_fatal_error_handler();
 
     if (!ptxOrErr) {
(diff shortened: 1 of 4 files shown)


atmnp:
Closed in favor of https://github.com/NVIDIA/numba-cuda-mlir/pull/252, since it merges this and https://github.com/NVIDIA/numba-cuda-mlir/pull/149.

acosmicflamingo:
@atmnp terrific! Glad to see @ZzEeKkAa and my changes were able to help shape the final implementation 😄

PR #149, opened 2026-06-23
Add dump nvvm env

ZzEeKkAa:
# Add `NUMBA_CUDA_MLIR_DUMP_NVVM` to dump NVVM IR before PTX/LTO generation

## Summary

Adds a new `NUMBA_CUDA_MLIR_DUMP_NVVM` environment variable that dumps the NVVM
IR (the LLVM IR actually handed to libnvvm) right before PTX/LTOIR is generated.
This is a debugging aid for inspecting exactly what the compiler feeds into
libnvvm, after MLIR lowering and the libnvvm-compatibility downgrades have been
applied.

## Motivation

We could already dump the MLIR module and the pre-downgrade LLVM IR, but there
was no way to capture the final NVVM IR that libnvvm consumes. When diagnosing
codegen/libnvvm issues, having that exact input is the most useful artifact.

## What changed

- **`config.py`**: Register the variable through the standard config machinery as
  `config.CUDA_DUMP_NVVM` (alongside the other `NUMBA_CUDA_MLIR_*` options), so it
  benefits from env reload and `.numba_config.yaml` support.
- **`mlir_optimization.py`**: Add a `_maybe_dump_nvvm()` helper and call it from
  `_prepare_llvm_ir()`. Since `_prepare_llvm_ir()` produces the bytes consumed by
  both `_compile_to_ptx()` and `_compile_to_ltoir()`, a single hook covers both
  the regular PTX path and the LTO path, as well as the `get_ptx()` / `optimize()`
  entry points.
- **`tests/test_dump_nvvm.py`**: Unit tests for the helper (no GPU required).

## Usage

| Value | Behavior |
| --- | --- |
| _(unset / empty)_ | Disabled (default, no-op) |
| `1`, `true`, `yes`, `stderr` | Print to stderr |
| `<directory>` | Write to `<dir>/nvvm-<pid>-<n>.{bc,ll}` |
| `<file>` | Write to that exact path (parent dirs created) |

The libnvvm input is usually serialized **bitcode**, so the helper detects the
bitcode magic (`BC\xC0\xDE`): on-disk dumps use a `.bc` suffix for bitcode and
`.ll` for textual IR. When printing to stderr, bitcode is summarized (byte count)
rather than dumped as raw binary, while textual IR is printed in full.

```bash
# Print to stderr
NUMBA_CUDA_MLIR_DUMP_NVVM=1 python my_kernel.py

# Write each compilation's NVVM IR into a directory
NUMBA_CUDA_MLIR_DUMP_NVVM=/tmp/nvvm-dumps python my_kernel.py

### DIFF
--- a/cext/mlir-llvm70/lib/CAPI.cpp
+++ b/cext/mlir-llvm70/lib/CAPI.cpp
@@ -116,6 +116,72 @@ int llvm70_translate_gpu_module_from_op(
   }
 }
 
+/// Translate a gpu.module Operation* to old (LLVM 7 dialect) NVVM IR text.
+/// This is the IR that would be handed to libnvvm by
+/// llvm70_translate_gpu_module_from_op; exposing it lets callers dump the
+/// NVVM IR for debugging (e.g. NUMBA_CUDA_MLIR_DUMP_NVVM) without compiling
+/// to PTX. libnvvm/libdevice are not needed because no compilation happens.
+int llvm70_translate_gpu_module_to_nvvm_ir_from_op(
+    void *raw_op,
+    const char *chip, const char *data_layout,
+    const char *libllvm, int gen_lineinfo,
+    char **out, size_t *out_len, char **err_out) {
+
+  *out = nullptr;
+  *out_len = 0;
+  *err_out = nullptr;
+
+  if (!raw_op) {
+    *err_out = copyCString("null Operation*");
+    return 1;
+  }
+
+  auto *op = static_cast<mlir::Operation *>(raw_op);
+  auto gpuMod = mlir::dyn_cast<mlir::gpu::GPUModuleOp>(op);
+  if (!gpuMod) {
+    *err_out = copyCString("Operation is not a gpu.module");
+    return 1;
+  }
+
+  llvm70::LLVM70Options opts;
+  opts.chip = chip ? chip : "sm_80";
+  opts.dataLayout = (data_layout && data_layout[0]) ? data_layout
+                                                    : kDefaultDataLayout;
+  if (libllvm && libllvm[0]) opts.libLLVMPath = libllvm;
+  opts.debugLevel = gen_lineinfo;  // 0=none, 1=lineinfo, 2=full debug
+
+  llvm::install_fatal_error_handler(fatalErrorHandler, nullptr);
+
+  try {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cext/mlir-llvm70/lib/CAPI.cpp [hunk 1]
```diff
@@ -116,6 +116,72 @@ int llvm70_translate_gpu_module_from_op(
...
+
+  if (!raw_op) {
+    *err_out = copyCString("null Operation*");
+    return 1;
+  }
+
+  auto *op = static_cast<mlir::Operation *>(raw_op);
+  auto gpuMod = mlir::dyn_cast<mlir::gpu::GPUModuleOp>(op);
+  if (!gpuMod) {
+    *err_out = copyCString("Operation is not a gpu.module");
```
atmnp, line 142: Can we reuse the logic from line 58 by refactoring it into a helper?

src/numba_cuda_mlir/mlir_optimization.py
```diff
@@ -168,6 +182,45 @@ def _get_op_ptr(op) -> ctypes.c_void_p:
...
+    if rc != 0:
+        msg = err_out.value.decode() if err_out.value else "unknown error"
+        if err_out.value:
+            lib.llvm70_free(err_out)
+        raise RuntimeError(f"llvm70 NVVM IR generation failed: {msg}")
+
+    try:
+        _maybe_dump_nvvm(ctypes.string_at(out, out_len.value))
+    finally:
+        if out.value:
```
atmnp, line 220: Might just be my misunderstanding, but isn't `out.value` the value of the string? So this is guarded on the truthiness of the string, not the actual pointer being non-null?

acosmicflamingo:
Before this PR gets merged, I want to bring up that @gmarkall had opened a [ticket](https://github.com/NVIDIA/numba-cuda-mlir/issues/95) about adding support for `inspect_llvm` in numba-cuda-mlir and I had created a PR that handles this [here](https://github.com/NVIDIA/numba-cuda-mlir/pull/145).

@ZzEeKkAa could the changes in that PR be beneficial for your use-case?

atmnp:
Closed in favor of https://github.com/NVIDIA/numba-cuda-mlir/pull/252, since it merges this and https://github.com/NVIDIA/numba-cuda-mlir/pull/145.

PR #155, opened 2026-06-26
Linker Re-Design

atmnp:
- Preserve resolved LTO mode for CUDA source links so existing LTOIR inputs do not mutate later CU link items.
  - Keep plain unlinked cuda.jit kernels on the PTX path by default while enabling LTO when external link inputs require it.
  - Make explicit lto=True and lto=False take precedence, and reject inconsistent LTOIR-with-lto=False plans.
  - Ensure lineinfo=True no longer disables LTO, and pass lineinfo/debug through NVRTC for CUDA source links.
  - Remove hidden cuda.jit(output=...); keep output= on compile APIs and track internal compile output as _compile_output.
  - Resolve link policy once after lowering, then replay collected link inputs into the selected linker.
  - Update MLIR disk-cache keys to track explicit LTO intent instead of late output mutation.
  - Preserve inspect_lto_ptx() for PTX-default kernels by lazily compiling diagnostic LTOIR when needed.
  - Fix LLVM70 lineinfo plus LTOIR linking by omitting the incompatible debug-info module flag.
  - Add focused regressions for PTX default behavior, LTO precedence, invalid LTOIR plans, callback fallback, lineinfo-preserved LTO, diagnostic LTO inspection, and CUDA source link mode preservation.


tests/numba_cuda_tests/cudadrv/test_module_callbacks.py
```diff
@@ -85,7 +85,7 @@ def setup(object_code):
         callback = CUSource("", setup_callback=setup)
         add_one = cuda.declare_device("add_one", "int32(int32)", link=[linked_function, callback])
 
-        @cuda.jit(output="ltoir")
+        @cuda.jit(lto=True)
```
gmarkall, line 88: If setting `lto=True` works here, then does it imply that setup callbacks do work with LTO? Or is the LTO silently being accepted but then modified internally?

tests/test_lto_defaults.py
```diff
@@ -0,0 +1,122 @@
...
+    assert plan.compile_new_inputs_as_ltoir is False
+    assert plan.lto_explicit is False
+
+
+def test_link_plan_keeps_implicit_ptx_for_cu_source_link_items():
+    dispatcher = cuda.jit(link=[cuda.CUSource("")])(_kernel)
+
+    plan = resolve_link_plan(dispatcher.targetoptions, dispatcher.targetoptions["link"])
+
+    assert plan.compile_new_inputs_as_ltoir is False
```
gmarkall: Shouldn't we be compiling with LTO if we just have CU source files? Otherwise we pay a performance penalty for the calls across translation units.

gmarkall (commented):
The logic is still complicated, but I feel like it's a bit easier to reason about what the linker will do here compared to #71.

However, I am still confused by the policy applied. I think the policy we want is:

- Use the PTX path if there is nothing externally linked
- Use the PTX path if at least one externally linked item is a PTX file.
- Maybe use the PTX path if there are setup / teardown callbacks. However, the test with setup callbacks and explicit LTO seems to imply that LTO works with setup callbacks, and I also think we wouldn't want to ding the performance of libraries that use setup / teardown callbacks by forcing them to use PTX instead of LTO - e.g. nvshmem.

Where I think this differs from what we want:

- It appears not to enable LTO for a kernel linked with cu sources.
- It appears to accept `lto=True` for a kernel with setup callbacks - I haven't yet worked out from reading the code if that means that LTO is enabled or if the request is ignored.

Maybe this is doing what we want and I've misunderstood - if that's the case, can you show me what might be incorrect above?

tests/numba_cuda_tests/cudadrv/test_module_callbacks.py
atmnp: Yes, setup callbacks work with LTO. lto=True is not ignored or modified internally. The generated module and CUDA sources are compiled as LTOIR. Callback metadata is carried separately and invoked after module loading. Also, I removed callbacks from implicit LTO selection. A callback-bearing CU source now uses implicit LTO unless another policy condition, such as a PTX input or debug=True, selects PTX.

tests/test_lto_defaults.py
atmnp: Yes, fixed. CUsource-only links now enable implicit LTO, allowing optimization across translation units. If any PTX input is present, newly generated inputs remain on the PTX path unless lto=True was explicitly requested. I added coverage for CU-only, CU+PTX, LTOIR+PTX, and explicit overrides.

atmnp:
So now:
- No external inputs use PTX.
- Any PTX input selects PTX for newly generated inputs during implicit selection.
- Otherwise, CU source or LTOIR inputs enable LTO.
- Setup and teardown callbacks do not affect code-generation selection.
- Explicit lto=True or lto=False takes precedence over implicit selection. Incompatible LTOIR plus lto=False remains an error.

We treat “any PTX input uses PTX” as an implicit-selection rule. Explicit lto=True still wins rather than being silently downgraded; external PTX remains PTX while newly generated inputs use LTOIR.

The callback test confirms that explicit LTO is real and that the setup callback runs after module loading. Enabling implicit LTO for CU sources also exposed a diagnostic inspect_lto_ptx() replay bug, where an already-linked CU source was added twice. I fixed that and added bookkeeping coverage.

I agree that this re-design isn't as "simple" as I was hoping, but I'm all for architectural changes to how we set up the Plans to make it simpler. I'm not sure if the solution is just better/more ironed out documentation on the exact contract we aim to provide here and then change all the verbiage in the code to align with it.

gmarkall (approved):

gmarkall (approved):

Issue #111, opened 2026-06-09
[BUG] Tuple slice inside kernel fails to compile


Jlisowskyy:
# Description

Slicing a compile-time-constant tuple inside a kernel (e.g. `shape[::-1]`) fails to
compile. The same code works on `numba-cuda`.

# Repro

```python
import numpy as np
from numba_cuda_mlir import cuda

SHAPE = (2, 4)


@cuda.jit
def kernel(out):
    rev = SHAPE[::-1]
    out[0] = rev[0]


out = cuda.to_device(np.zeros(1, np.int64))
kernel[1, 1](out)
cuda.synchronize()
print(out.copy_to_host())
```

# Actual behavior

```
  File "/home/jlisowski/repos/numba-cuda-mlir-upstream/src/numba_cuda_mlir/mlir_lowering.py", line 1295, in lower_expr_assign
    self.lower_getitem_expr_assign(target, expr.value, index)
  File "/home/jlisowski/repos/numba-cuda-mlir-upstream/src/numba_cuda_mlir/mlir_lowering.py", line 1568, in lower_getitem_expr_assign
    raise NotImplementedError(f"lowering getitem {target} = {value}[{index}]")
NotImplementedError: lowering getitem rev = $4load_global.0[$20build_slice.5]
```

# Expected

Tuple slice works: code prints `4`, as on `numba-cuda`.

PR #158, opened 2026-06-29
Support compile-time tuple slicing

atmnp:
Preserve literal indices for tuple static_getitem while retaining index-variable lowering for other containers (src/numba_cuda_mlir/mlir_lowering.py:1303-1314).

Type integer and slice constants as literals, then lower tuple slices through compile-time Python slicing (src/numba_cuda_mlir/mlir_lowering.py:1567-1583; src/numba_cuda_mlir/lowering/numpy.py:1402-1409).

Closes issue #111 and enable tuple cases that no longer ICE (tests/test_tuple.py:10-18; tests/numba_cuda_tests/cudapy/test_array_args.py:105-190).


Jlisowskyy (approved):
LGTM!

PR #160, opened 2026-06-30
Fix nested record field access lowering

### LINKED ISSUE #156: [BUG] ICE: `memref.extract_aligned_pointer_as_index` receives `!llvm.ptr` instead of memref for nested record access
I come bearing (slightly AI generated but correct) bug reports. Looks like this works nicely if everything is an array, but not when it's a single scalar.

## Reproducer

```python
import numpy as np

nested_dt = np.dtype([('inner', [('a', '<f8'), ('b', '<f8')], (1,))])

@cuda.jit
def kernel(arr):
    arr[0]['inner'][0]['a'] = 11.0
    arr[0]['inner'][0]['b'] = 22.0

buf = np.zeros(2, dtype=nested_dt)
kernel[1, 1](buf)
```

## Error

```
numba_cuda_mlir.errors.InternalCompilerError: Invalid MLIR module: Verification failed:
error: 'memref.extract_aligned_pointer_as_index' op operand #0 must be ranked or unranked
memref of any type values, but got '!llvm.ptr'
 note: see current operation:
   %12 = "memref.extract_aligned_pointer_as_index"(%10) : (!llvm.ptr) -> index
This is always a bug in the compiler, please file an issue.
```

The lowering computes a sub-record pointer via `llvm.inttoptr` + `llvm.getelementptr` (producing `!llvm.ptr`), then passes it to `memref.extract_aligned_pointer_as_index` which expects a `memref<>`.

Without the `(1,)` shape (i.e., plain nested record `('inner', [('a','<f8'),('b','<f8')])`), the same bug manifests as `CUDA_ERROR_ILLEGAL_ADDRESS` at runtime instead of a compile-time verification failure.

## Environment

- numba-cuda-mlir 0.4.0, Python 3.12.13, NumPy 2.5.0
- CUDA Toolkit 13.2, Tesla T4 (sm_75), Windows 11 x86_64

## Notes

- Flat structured dtypes (no nesting) work correctly, including array fields: `('a', '<f8', (1,)), ('b', '<f8', (1,))` accessed as `arr[0]['a'][0]` works fine.
- The bug is specifically sub-record descent (`arr['outer']['inner']`). Keeping all fields at one level of record avoids it.
- This blocks any codebase using nested NumPy record arrays as kernel arguments.

atmnp:
Dispatch NestedArray access before record-array lowering and preserve nested Record addresses instead of loading them as pointer values (src/numba_cuda_mlir/lowering/numpy.py:1033, src/numba_cuda_mlir/lowering/record.py:109).

Cover shaped and direct nested structured-dtype field writes (tests/test_record_dtype.py:263).

This closes #156.

### DIFF
--- a/tests/test_record_dtype.py
+++ b/tests/test_record_dtype.py
@@ -28,6 +28,8 @@ recordwithvaluestorage = np.dtype([("h", np.float16), ("flag", np.bool_)], align
 
 nestedrecordwitharray = np.dtype([("inner", [("a", np.float64), ("b", np.float64)], (1,))])
 
+nestedrecordwith2darray = np.dtype([("inner", [("a", np.float64), ("b", np.float64)], (2, 2))])
+
 nestedrecord = np.dtype([("inner", [("a", np.float64), ("b", np.float64)])])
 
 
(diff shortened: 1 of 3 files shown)


tests/test_record_dtype.py
```diff
@@ -253,6 +257,25 @@ def set_fields(ary):
...
+        result = arr.copy_to_host()
+        scalar_result = scalar_arr.copy_to_host()
+        np.testing.assert_equal(result[0]["inner"][0]["a"], 11.0)
+        np.testing.assert_equal(result[0]["inner"][0]["b"], 22.0)
+        np.testing.assert_equal(scalar_result[0]["inner"]["a"], 33.0)
+        np.testing.assert_equal(scalar_result[0]["inner"]["b"], 44.0)
+
 
 @pytest.mark.skip(reason="Causes memory errors")
 class TestScalarRecordArgument:
```
VijayKandiah, line 296: Can you also add a test for tuple-indexed nested record access to invoke the `lower_nested_array_getitem_tuple` Record branch?

VijayKandiah (approved):
LGTM after adding another test.

PR #161, opened 2026-06-30
[Debug Info] Fix bitwise logic ops result types

jiel-nv:
When a CUDA kernel applies a bitwise logic operator (`&`, `|`, `^`) to two `int32` operands, the result variable is annotated as `int64` in the DWARF debug information. This widening is part of Numba's integer typing NBEP, intentional for arithmetic operators, but it is unnecessary for bitwise operators: AND, OR, and XOR. The result of a bitwise operation always fits in the wider of the two operand types, so no promotion is needed.


atmnp (approved):
Thanks Jie!

PR #162, opened 2026-06-30
Merge LTO error-code globals with common linkage

### LINKED ISSUE #119: "symbol multiply defined" when linking multiple device functions
## Summary

Every device function compiled by numba-cuda-mlir emits a global
`__numba_cuda_mlir_error_code` with **`external` (defining) linkage**. When two
or more such device functions are linked into a single program (e.g. with
nvJitLink in LTO mode, as accelerated libraries do), the link fails with:

```
link error: error: Linking globals named '__numba_cuda_mlir_error_code': symbol multiply defined!
ERROR NVVM_ERROR_COMPILATION (9) in nvvmCompileProgram callback
```

This blocks any use case that compiles several user operators separately and
links them together (custom comparators/predicates, transform+reduce pipelines,
multi-operator algorithms, etc.).

## Environment

- numba-cuda-mlir 0.3.0
- CUDA Toolkit 13.3 (nvcc V13.3.33), cuda.bindings 13.3.1
- Python 3.12, Linux

## Minimal reproducer (no external library required)

```python
from cuda.bindings import nvjitlink

def op_a(x): return x + T.int32(1)
def op_b(x): return x + T.int32(2)

lto_a, _ = cuda.compile(op_a, T.int32(T.int32), device=True, abi="c",
                        abi_info={"abi_name": "op_a"}, output="ltoir")
lto_b, _ = cuda.compile(op_b, T.int32(T.int32), device=True, abi="c",
                        abi_info={"abi_name": "op_b"}, output="ltoir")

def link(items, label):
    h = nvjitlink.create(2, ["-lto", "-arch=sm_89"])
    try:
        for data, name in items:
            nvjitlink.add_data(h, nvjitlink.InputType.LTOIR, data, len(data), name)
        nvjitlink.complete(h)
        print(label, "-> OK")
    except Exception:
        n = nvjitlink.get_error_log_size(h); buf = bytearray(n)
        nvjitlink.get_error_log(h, buf)
        print(label, "-> FAIL:", buf.decode(errors="replace").strip())

link([(lto_a, "op_a")], "one op")                       # OK
link([(lto_a, "op_a"), (lto_b, "op_b")], "two ops")     # FAIL
```

Output:

```
one op -> OK
two ops -> FAIL: op_b: link error: error: Linking globals named '__numba_cuda_mlir_error_code': symbol multiply defined!
ERROR NVVM_ERROR_COMPILATION (9) in nvvmCompileProgram callback
```

## Expected vs actual

- **Expected:** linking N independently-compiled device functions succeeds; the
  `__numba_cuda_mlir_error_code` global is defined at most once in the linked
  program (or is mergeable).
- **Actual:** each object carries its own *defining* `external` global, so the
  linker reports a duplicate symbol and aborts.

## Root cause

`numba_cuda_mlir/mlir_lowering.py`, `MLIRLower._ensure_error_global` (~line 382)
creates the global with `#llvm.linkage<external>` and an initializer in **every**
compilation unit:

```python
linkage = ir.Attribute.parse("#llvm.linkage<external>")
llvm.GlobalOp(T.i32(), ERROR_CODE_GLOBAL_NAME, linkage,
              value=ir.IntegerAttr.get(T.i32(), 0), addr_space=0)
```

The docstring there notes "When device function MLIR is later linked into a
kernel module, duplicate globals are skipped by the linker" — but that
assumption does not hold for nvJitLink/LTO linking of independently compiled
LTOIR objects.

## Notes on attempted workaround

Changing the linkage to `linkonce_odr` *does* propagate to the emitted LLVM IR
(verified: `@__numba_cuda_mlir_error_code = linkonce_odr global i32 0`), but
nvJitLink's LTO step **still** reports "symbol multiply defined". So a mergeable
LLVM linkage alone may be insufficient — the global may need to be emitted as a
declaration (resolved once at final link) or otherwise made LTO-mergeable.
Guidance on the intended cross-object story for this symbol would help.

## Impact

High for accelerated-library use (compiling user operators to LTOIR and linking
them via nvJitLink). In our integration this blocks ~107 test cases across
multi-operator algorithms (partition with two predicates, reduce/scan over a
transform iterator, segmented reduce, …).

atmnp:
Use common linkage for the mutable zero-initialized error-code global so nvJitLink can merge definitions from independently compiled device-function LTOIR objects. The global remains available for host error reporting (src/numba_cuda_mlir/mlir_lowering.py:391).

Assert the emitted common PTX global and link two separately compiled LTOIR device functions in one nvJitLink invocation (tests/test_kernel_exceptions.py:52, tests/test_kernel_exceptions.py:64).

This closes #119.

### DIFF
--- a/src/numba_cuda_mlir/mlir_lowering.py
+++ b/src/numba_cuda_mlir/mlir_lowering.py
@@ -388,7 +388,7 @@ class MLIRLower(object):
             )
             if not already_exists:
                 with ir.InsertionPoint.at_block_begin(gpu_block):
-                    linkage = ir.Attribute.parse("#llvm.linkage<external>")
+                    linkage = ir.Attribute.parse("#llvm.linkage<common>")
                     llvm.GlobalOp(
                         T.i32(),
                         ERROR_CODE_GLOBAL_NAME,
--- a/tests/test_kernel_exceptions.py
+++ b/tests/test_kernel_exceptions.py
@@ -6,6 +6,9 @@
 import numpy as np
 import pytest
 from numba_cuda_mlir import cuda
+from numba_cuda_mlir.numba_cuda import require_context, types
+from numba_cuda_mlir.numba_cuda.testing import skip_if_nvjitlink_missing
+from numba_cuda_mlir.tools import get_gpu_compute_capability
 
 
 @cuda.jit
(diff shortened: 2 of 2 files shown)


acosmicflamingo:
Nice!

VijayKandiah (approved):
LGTM

PR #166, opened 2026-07-01
Add native complex32 type support

### LINKED ISSUE #163: [FEA] Add native complex32 type
numba-cuda-mlir currently provides only complex64 and complex128 as complex scalar types. There is no native complex32 support.

This proposal is to add complex32 as a first-class scalar type, on par with complex64 and complex128, usable in function signatures, kernel locals, arrays, and arithmetic operations.

Today, users need to emulate complex32 using the corresponding vector types. Providing native complex32 support would simplify those workflows and make certain use cases easier to express.

atmnp:
Initial support for complex numbers in numba-cuda-mlir was limited to the NumPy types. Issue #163 asks for a native complex32 type (not in NumPy), this PR introduces the basic typing rules and a basic test to support this.

This closes #163.

### DIFF
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -118,6 +118,28 @@ def test_shared_memory_complex_real_imag(complex_dtype, float_dtype):
     np.testing.assert_allclose(out_imag, inp.imag)
 
 
+def test_shared_memory_complex32_real_imag():
+    @cuda.jit
+    def kernel(inp_real, inp_imag, out_real, out_imag):
+        tid = cuda.threadIdx.x
+        sm = cuda.shared.array(N_SHARED, dtype=types.complex32)
+        sm[tid] = types.complex32(inp_real[tid], inp_imag[tid])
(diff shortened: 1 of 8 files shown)


tests/test_complex.py
Jlisowskyy, line 1: Should we add `complex32` to all existing complex tests? It might help us catch issues earlier

Jlisowskyy (approved):
Looks good to me. I ran the tests locally, and everything passed. I also manually checked the usages, and they look fine. I’m not an expert in this area, but it looks good based on my understanding.

ZzEeKkAa:
Thank you! Should we add native ml_dtype support for c32 as well? https://github.com/jax-ml/ml_dtypes/issues/110

atmnp:
I think we could and perhaps should, but that should be a rounded out ml_dtypes PR. I know we have some support now, but we should one-shot it as best as we can to complete/verify the coverage.

PR #167, opened 2026-07-01
Improve register-pressure launch diagnostics

atmnp:
CUDA reports excessive kernel register usage as a generic CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES. This does not explain which resource caused the failure or how users can make the launch viable.

When a launch fails with this error, query the kernel's registers per thread and the device's register limit per block. If the requested block provably requires more registers than the device supports, report the calculation and recommend an appropriate cuda.jit(max_registers=N) value. Preserve the original CUDA error when register pressure cannot be confirmed.

- cext/launcher/cuda_shim.h: add the CUDA result and attribute constants needed by the native launcher.
- cext/launcher/kernel.cpp: diagnose impossible register requirements after failed launches without adding overhead to successful launches.
- tests/numba_cuda_tests/cudadrv/test_linker.py: construct an architecture-aware excessive-register launch and verify the detailed CUDAError and max_registers recommendation.


cext/launcher/kernel.cpp
```diff
@@ -1698,6 +1698,43 @@ bool try_clarify_invalid_value_error(const Grid& grid) {
     return false;
 }
 
+bool try_clarify_launch_out_of_resources_error(CUkernel kernel, const Grid& block) {
+    CUfunction function;
+    if (g_cuKernelGetFunction(&function, kernel) != CUDA_SUCCESS) return false;
```
ZzEeKkAa, line 1705: Should we add a comment here that we prefer to not do anything here if we fail to get preconditions to avoid hiding root problem?

cext/launcher/kernel.cpp
```diff
@@ -1698,6 +1698,43 @@ bool try_clarify_invalid_value_error(const Grid& grid) {
...
+    if (g_cuFuncGetAttribute(&registers_per_thread, CU_FUNC_ATTRIBUTE_NUM_REGS, function)
+            != CUDA_SUCCESS)
+        return false;
+
+    CUdevice device;
+    if (g_cuCtxGetDevice(&device) != CUDA_SUCCESS) return false;
+
+    int max_registers_per_block;
+    if (g_cuDeviceGetAttribute(&max_registers_per_block,
+            CU_DEVICE_ATTRIBUTE_MAX_REGISTERS_PER_BLOCK, device) != CUDA_SUCCESS)
```
ZzEeKkAa, line 1717: Should we also consider `CU_DEVICE_ATTRIBUTE_MAX_REGISTERS_PER_MULTIPROCESSOR`? Apparently, there could be multiple blocks run per MP. That's what second arg of `launch_bounds` control: https://docs.nvidia.com/cuda/cuda-c-programming-guide/#launch-bounds

ZzEeKkAa: Is there a way to get blocks per MP? I only see the `CU_DEVICE_ATTRIBUTE_MAX_BLOCKS_PER_MULTIPROCESSOR`. Probably it should be somewhere at launch config

ZzEeKkAa: There is also `CU_DEVICE_ATTRIBUTE_MAX_THREADS_PER_MULTIPROCESSOR`, but it is still device attribute, have nothing to do with the particular kernel

atmnp: My understanding is that `minctasm` (what that arg sets) is a hint in that ptxas uses it to choose a register budget, but it is still compile-time metadata, not a launch-time precondition. The launch config gives grid/block/sharedmem/cluster, not requested blocks-per-SM. We can also add a diagnostic for errors where the `maxActiveBlocksPerMultiprocessor` and the `minBlocksPerMultiprocessor` are not satisfiable, but it's a different case than the one we're trying to tackle here.

cext/launcher/kernel.cpp
atmnp: added!

ZzEeKkAa (approved):

Issue #165, opened 2026-07-01
[BUG] Slow string compile times


Jlisowskyy:
# Description

A device-side `x == "const"` comparison compiles significantly slower than the equivalent
bool-constant branch.

# Repro

```python
import os
import time
import warnings

import numpy as np

from numba_cuda_mlir import cuda
from numba_cuda_mlir.extending import overload, typing_registry

warnings.filterwarnings("ignore", message="Grid size")
warnings.filterwarnings("ignore", message="Linking LTOIR")

USE_COMPARE = os.environ.get("USE_COMPARE", "1") == "1"


def arrangement():
    raise NotImplementedError


@overload(arrangement, strict=False, inline="always", typing_registry=typing_registry)
def _ol_arrangement():
    value = "col_major" if USE_COMPARE else True
    return lambda: value


if USE_COMPARE:

    @cuda.jit
    def kernel(out):
        out[0] = 1 if arrangement() == "col_major" else 0

else:

    @cuda.jit
    def kernel(out):
        out[0] = 1 if arrangement() else 0


out_d = cuda.to_device(np.zeros(1, dtype=np.int64))

t0 = time.perf_counter()
kernel[1, 1](out_d)
cuda.synchronize()
compile_s = time.perf_counter() - t0

print(f"USE_COMPARE={int(USE_COMPARE)} compile={compile_s:.2f}s result={out_d.copy_to_host()}")
```

# Actual behavior

The string path takes significantly longer to compile than the bool path, for a single comparison
site, both return the same result.

```sh
USE_COMPARE=0 python repro.py ; USE_COMPARE=1 python repro.py
```

returns:

```
USE_COMPARE=0 compile=0.24s result=[1]
USE_COMPARE=1 compile=2.32s result=[1]
```

# Expected

String-path compile time should match the bool-constant path.

PR #168, opened 2026-07-01
Prioritize MLIR unicode comparison typing

atmnp:
Mark UnicodeComparisonTemplate as CUDA-target-specific so it resolves before generic Numba-CUDA unicode and charseq overloads. This avoids compiling unused comparison implementations during literal equality typing (src/numba_cuda_mlir/typing/unicode.py:24-32).

Add an inline-overload regression matching issue #165 and verify both CUDA template priority and comparison behavior
(tests/test_string_types.py:19-40).

This closes issue #165.


Jlisowskyy (approved):
Works great: over a 2x speedup with string comparison in our tests. Such a small change for a big win. Thanks!

PR #171, opened 2026-07-02
Isolate bundled LLVM/MLIR from differently-versioned LLVM in-process

### LINKED ISSUE #170: libMLIRPythonCAPI.so is loaded with RTLD_GLOBAL — symbol collision risk with other LLVM-embedding extensions
`libMLIRPythonCAPI.so` is opened with `RTLD_NOW | RTLD_GLOBAL`, which pushes every `LLVM*` / `MLIR*` C symbol into the process-global namespace:

https://github.com/NVIDIA/numba-cuda-mlir/blob/ef2009e70d39379fd98939489d14c88a8e00d14a/cext/launcher/llvm_downgrade.cpp#L196

Any other Python extension in the same process that also embeds LLVM (Triton, Halide, Mojo, JAX/XLA in some configurations, a future llvmlite regression, ...) may then resolve to numba-cuda-mlir's symbols instead of its own — the same class of bug that forced llvmlite to statically link LLVM with hidden visibility (conda-forge/llvmlite-feedstock#100).

The LLVM 7 path already gets this right:

https://github.com/NVIDIA/numba-cuda-mlir/blob/ef2009e70d39379fd98939489d14c88a8e00d14a/cext/mlir-llvm70/include/llvm70/CAPILoader.h#L53

`RTLD_LOCAL` keeps LLVM 7 symbols out of the global namespace, and `RTLD_DEEPBIND` also insulates it from external pollution.

Build config: `libMLIRPythonCAPI.so` is built without `-fvisibility=hidden` or an `LLVM_EXPORTED_SYMBOL_FILE`, so all `LLVM*` / `MLIR*` symbols end up exported. The nanobind domain (`MLIR_BINDINGS_PYTHON_NB_DOMAIN=numba_cuda_mlir`) at `ci/build-llvm-modern.sh:70` handles MLIR-Python coexistence but not the LLVM C surface.

### Suggested fix

Minimum: change to `RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND`. The launcher only uses `dlsym(handle, ...)` on the returned handle, which works fine with `RTLD_LOCAL`.

Longer-term: build with `-fvisibility=hidden -fvisibility-inlines-hidden` and an `LLVM_EXPORTED_SYMBOL_FILE` limiting exports to the C API surface actually needed, matching the llvmlite playbook.

leofang:
Fixes #170.

Deviation from the issue's sketch worth noting: DEEPBIND is applied
via a ctypes preload in `numba_cuda_mlir/__init__.py`, not at the
launcher's dlopen. By the time the launcher runs, libMLIRPythonCAPI.so
is already loaded via DT_NEEDED and DEEPBIND on a re-open is a no-op.

Regression tests in `tests/test_symbol_isolation.py` (Linux-only,
subprocess-based) were each verified to fail with the corresponding
fix reverted.


leofang:
/ok to test [60e31ec](https://github.com/NVIDIA/numba-cuda-mlir/pull/171/commits/60e31ecd427689c2f965cdb7c08012eefa1d4b7f)

atmnp (approved):
LGTM, thanks Leo!

Issue #178, opened 2026-07-06
float32 stores to dynamic shared memory (cuda.shared.array(0, ...)) deleted at opt>0


ccam80:
Stores to a dynamic shared memory array (`cuda.shared.array(0, dtype)`) are deleted by the optimizer (we see it with float32).

## Repro (0.4.0 wheel, RTX 4070 SUPER, CUDA 12)

```python
import numpy as np
from numba_cuda_mlir import cuda
from numba_cuda_mlir.types import float32

@cuda.jit
def k(inp, idx_arr, out):
    shared = cuda.shared.array(0, dtype=float32)
    for i in range(inp.shape[0]):
        shared[0] = inp[i, 0]
        out[i, 0] = shared[idx_arr[0]]

inp = np.array([[1.5], [2.5], [3.5]], dtype=np.float32)
idx = np.zeros(1, dtype=np.int32)
out = cuda.to_device(np.zeros((3, 1), dtype=np.float32))
k[1, 1, 0, 4](cuda.to_device(inp), cuda.to_device(idx), out)
np.testing.assert_allclose(out.copy_to_host(), inp)  # fails
```

## Expected behaviour

The stores land and `out == inp`.

## Observed behaviour

`out` comes back all zeros. The load index has to be runtime-opaque to show the bug; with a constant index the compiler folds the load and store to the same address and the reordering is hidden.

This looks like #80/#117, but those conclude the deletion is limited to fp16/bf16 vector stores through the nvjitlink LTO path, with f32/f64 unaffected. That doesn't match what I see: the stores here are float32, and with the LTO optimization disabled the libnvvm -O3 compile deletes them as well, so nvjitlink isn't required.

## Cause

`gpu.dynamic_shared_memory` lowers to `llvm.mlir.global internal @__dynamic_shmem__N : !llvm.array<0 x i8>`. I think the optimizer reads this as the complete definition, and it deletes the stores. numba-cuda declares it external, so the optimizer expects a size outside the module. `shared.array(0, dtype)` also takes the static-shape lowering branch, so it gets its own private zero-length `memref.global` rather than a view of the dynamic region.

I've patched this locally by lowering `shared.array(0, dtype)` to a `memref.view` over `gpu.dynamic_shared_memory`, sized at runtime, and rewriting the `__dynamic_shmem__*` globals to external linkage before codegen — PR linked. That fixes the repro and our production use, but if it's treating a symptom and the intended fix lives elsewhere, happy to rework it.

[...]

PR #179, opened 2026-07-06
Emit shared memory views at the current insertion point

ccam80:
Follow-up to #178: the store deletion reported there is fixed on main by the dynamic shared memory views in #155 (verified with lto=True as well). This PR is now rescoped to the crash that remains.

`_request_dynamic_shared_memory` and `_request_shared_memory` append their view ops to the entry block. Once the entry block has a terminator, the append raises `IndexError: Cannot insert operation at the end of a block that already has a terminator`, so any `cuda.shared.array` request placed after control flow crashes the compiler. For the dynamic region this is new surface from #155 (`shared.array(0)` previously took the static-global path); for runtime-shaped arrays it predates it.

The change emits both views at the current insertion point instead. The shared-memory base is still created at the entry block's start by `_get_shared_memory_base`, and for runtime-shaped requests the size operands are computed at the request site anyway, so this is also where they dominate the view.

Tests: two regressions in `tests/test_dynamic_shared_memory.py`, one per request path, each placing the array after a branch. Both crash on main and pass with this change. Full test suite run against a main baseline with no new failures, on an RTX 4070 SUPER, CUDA 12.

---


atmnp (changes requested):
This fixes the single-allocation-after-control-flow case, but can we add a regression with two shared allocations where the first request is inside/after control flow and the second request occurs later? _total_shared_memory_bytes now becomes an SSA value emitted at the current insertion point, so I want to make sure a later shared-memory request cannot use a non-dominating offset value after a join/sibling branch.

Other than that, looks great! Thank you for catching this regression this quickly.

ccam80:
@atmnp Thanks for finding that gap — my PR broke under that regression. I got it working with a cached memref.dim(shm_base, 0) at the start of the entry block, and replaced the kernel-path zero-init and the request-site fallback with a single lazy init emitted alongside it, so device functions and kernels share one path. That covers everything except a runtime-shaped allocation inside a branch arm, as the offset increment uses request-site size values that can't be hoisted, so a later request after one of those now raises a LoweringError instead of tripping the MLIR verifier. Added three regressions: runtime-shaped then dynamic after control flow, dynamic-in-branch then a later request, and the error case. 

Is there any IM or chat platform set up for numba-cuda for quick dev queries? I'd like to check in on general good-citizen rules for contributing, as I'm currently spamming the project with issues and PRs and am unsure if that noise is wanted, as well as minors like whether I should be merging main into PR branches. I've got a couple more bug-fixes/patches in the queue, currently blocked by empathy when viewing the list of open PRs.

Jlisowskyy:
Hi @ccam80, thanks for the PR and for working on this bug! We ended up merging a similar fix as part of a combined MR with a few small related changes, so this PR is now superseded. I’ll close this one to avoid duplication. Thanks again for the contribution!

Issue #173, opened 2026-07-02
Windows on ARM notes


gmarkall:
In order to make Numba-CUDA-MLIR work on Windows on ARM, a few changes are necessary. This issue documents the fixes I had to make in order to run the test suite on a Windows on ARM machine.

## Building with Visual Studio 2022

I initially tried building with Visual Studio 2026, thinking that was most likely to be reliable for a Windows on ARM build of LLVM and related tools. However, that runs into an ICE when compiling the modern LLVM, with the message:

```
OpDefinitionsGen.cpp(4039) : fatal error C1001: Internal compiler error
```

Compiling with Visual Studio 2022 seems to avoid the issue.

## Linking for `ARM64` target when building `llvm-c-stub` for LLVM 7

The patch that I applied for this to build was:

```diff
diff --git a/ci/build-windows.sh b/ci/build-windows.sh
index 6574beb..4904778 100755
--- a/ci/build-windows.sh
+++ b/ci/build-windows.sh
@@ -229,7 +229,7 @@ build_llvm7() {
   {
     printf '/NOLOGO\n'
     printf '/DLL\n'
-    printf '/MACHINE:X64\n'
+    printf '/MACHINE:ARM64\n'
     printf '/OUT:%s\n' "$(cmake_path "${llvm_c_dll}")"
     printf '/IMPLIB:%s\n' "$(cmake_path "${llvm_c_import_lib}")"
     printf '/DEF:%s\n' "$(cmake_path "${def_file}")"
```

However, a better situation would be not to build LLVM 7 for Windows on ARM - there is no probably no hardware configuration that would require it on Windows on ARM.

# Working around nanobind issues

I think that nanobind 2.13 has some issues that affected Windows on ARM - I came across this in https://github.com/llvm/llvm-project/issues/205329 and I saw that there seems to be some attempt to fix issues related to nanobind 2.13 upstream, in https://github.com/llvm/llvm-project/commit/657f1ccfbfc0dfd0abadc7477275404f01699598. I was running into an issue the the MLIR Python bindings that would manifest as an uncaught C++ exception being thrown in `nanobind::detail::raise_python_error()`.

To work around this, I built the MLIR bindings with nanobind 2.10 installed. However, this runs into another issue during the CMake configuration, https://github.com/wjakob/nanobind/issues/1237/. To work around this I applied:

```diff
diff --git a/llvm/CMakeLists.txt b/llvm/CMakeLists.txt
[...]

PR #183, opened 2026-07-06
Bump LLVM Commit

atmnp:
This is a test to resolve the regression introduced by the latest version of nanobind (2.13) interacting with our modern llvm commit hash which this now updates.

This was originally reported in #173.


lahwaacz:
@atmnp Is there a bug report for this nanobind issue? Patching it will not work when it is installed system-wide (which is admittedly not a problem for your CI, but others may need to build from source too...)

atmnp:
There is no separate issue yet but it is time we create one. It's mentioned in the WoA issue, and of course, the patch is not a real solution - just the fastest way for me to triage the failure/confirm in the Windows CI. The real fix needs to happen within the MLIR bindings/our usage of them, which now expect a behavior no longer supported by nanobind (https://github.com/wjakob/nanobind/pull/1364).

I will close this when I submit the real fix.

lahwaacz:
I'm asking because I've got the same problem with nanobind 2.13 on Arch Linux.

FWIW, I had an AI agent working on this for a few hours and it suggested this patch:

```patch
--- a/src/numba_cuda_mlir/mlir/dialect_exts/arith.py
+++ b/src/numba_cuda_mlir/mlir/dialect_exts/arith.py
@@ -132,92 +132,6 @@
     return get_op_result_or_op_results(IndexCastOp(res_type, in_, loc=loc, ip=ip))


-class ArithValueMeta(type(Value)):
-    """Metaclass that orchestrates the Python object protocol
-    (i.e., calling __new__ and __init__) for Indexing dialect extension values
-    (created using `mlir_value_subclass`).
-
-    The purpose/benefit of handling the `__new__` and `__init__` calls
-    explicitly/manually is we can then wrap arbitrary Python objects; e.g.
-    all three of the following wrappers are equivalent:
-
-    ```
-    s1 = ScalarValue(arith.ConstantOp(f64, 0.0).result)
-    s2 = ScalarValue(arith.ConstantOp(f64, 0.0))
-    s3 = ScalarValue(0.0)
-    ```
-
-    In general the Python object protocol for an object instance is determined
-    by `__call__` of the object class's metaclass, thus here we overload
-    `__call__` and branch on what we're wrapping there.
-
-    Why not just overload __new__ and be done with it? Because then we can't
-    choose what get's passed to __init__: by default (i.e., without overloading
-    __call__ here) the same arguments are passed to both __new__ and __init__.
-
-    Note, this class inherits from `type(Value)` (i.e., the metaclass of
-    `ir.Value`) rather than `type` or `abc.ABCMeta` or something like this because
-    the metaclass of a derived class must be a (non-strict) subclass of the
-    metaclasses of all its bases and so all the extension classes
-    (`ScalarValue`, `TensorValue`), which are derived classes of `ir.Value` must
-    have metaclasses that inherit from the metaclass of `ir.Value`. Without this
-    hierarchy Python will throw `TypeError: metaclass conflict`.
-    """
-
-    def __call__(cls, *args, **kwargs):
-        """Orchestrate the Python object protocol for mlir
-        values in order to handle wrapper arbitrary Python objects.
-
-        Args:
-          *args: Position arguments to the class constructor. Note, currently,
-            only one positional arg is supported (so constructing something like a
-            tuple type from element objects isn't supported).
-          **kwargs: Keyword arguments to the class constructor. Note, currently,
-            we only look for `dtype` (an `ir.Type`).
-
-        Returns:
-          A fully constructed and initialized instance of the class.
-        """
-        if len(args) != 1:
-            raise ValueError("Only one non-kw arg supported.")
-        arg = args[0]
-        fold = None
-        if isinstance(arg, (OpView, Operation, Value)):
-            # wrap an already created Value (or op the produces a Value)
-            if isinstance(arg, (Operation, OpView)):
-                assert len(arg.results) == 1
-            val = get_op_result_or_value(arg)
-        elif isinstance(arg, (int, float, bool, np.ndarray)):
-            # wrap a Python value, effectively a scalar or tensor literal
-            dtype = kwargs.get("dtype")
-            if dtype is not None and not isinstance(dtype, Type):
-                raise ValueError(f"{dtype=} is expected to be an ir.Type.")
-            fold = kwargs.get("fold")
-            if fold is not None and not isinstance(fold, bool):
-                raise ValueError(f"{fold=} is expected to be a bool.")
-            loc = kwargs.get("loc")
-            ip = kwargs.get("ip")
-            # If we're wrapping a numpy array (effectively a tensor literal),
-            # then we want to make sure no one else has access to that memory.
-            # Otherwise, the array will get funneled down to DenseElementsAttr.get,
-            # which by default (through the Python buffer protocol) does not copy;
-            # see mlir/lib/Bindings/Python/IRAttributes.cpp#L556
-            val = constant(deepcopy(arg), dtype, loc=loc, ip=ip)
-        else:
-            raise NotImplementedError(f"{cls.__name__} doesn't support wrapping {arg}.")
-
-        # The mlir_value_subclass mechanism works through __new__
-        # (see mlir/Bindings/Python/PybindAdaptors.h#L502)
-        # So we have to pass the wrapped Value to the __new__ of the subclass
-        cls_obj = cls.__new__(cls, val)
-        # We also have to pass it to __init__ because that is required by
-        # the Python object protocol; first an object is new'ed and then
-        # it is init'ed. Note we pass arg_copy here in case a subclass wants to
-        # inspect the literal.
-        cls.__init__(cls_obj, val, fold=fold)
-        return cls_obj
-
-
 @register_attribute_builder("Arith_CmpIPredicateAttr", replace=True)
 def _arith_CmpIPredicateAttr(predicate: Union[str, Attribute], context: Context):
     predicates = {
@@ -447,22 +361,37 @@
     return _pow_op(lhs, rhs, loc=loc)


-class ArithValue(Value, metaclass=ArithValueMeta):
+class ArithValue(Value):
     """Class for functionality shared by Value subclasses that support
     arithmetic operations.

-    Note, since we bind the ArithValueMeta here, it is here that the __new__ and
-    __init__ must be defined. To be precise, the callchain, starting from
-    ArithValueMeta is:
-
-    ArithValueMeta.__call__ -> mlir_value_subclass.__new__ ->
-                          (mlir_value_subclass.__init__ == ArithValue.__init__) ->
-                          Value.__init__
+    Wrapping logic is implemented in ``__new__`` so that nanobind's own
+    instance-creation path (``nb_type.__call__``) is used. That path knows how
+    to initialize a Python subclass of a nanobind-bound ``Value`` correctly;
+    calling ``cls.__init__`` explicitly from a Python metaclass does not.
     """

-    def __init__(self, val, *, fold: Optional[bool] = None):
+    def __new__(cls, arg, *, dtype=None, fold=None, loc=None, ip=None):
+        if isinstance(arg, (OpView, Operation, Value)):
+            if isinstance(arg, (Operation, OpView)):
+                assert len(arg.results) == 1
+            val = get_op_result_or_value(arg)
+        elif isinstance(arg, (int, float, bool, np.ndarray)):
+            if dtype is not None and not isinstance(dtype, Type):
+                raise ValueError(f"{dtype=} is expected to be an ir.Type.")
+            if fold is not None and not isinstance(fold, bool):
+                raise ValueError(f"{fold=} is expected to be a bool.")
+            val = constant(deepcopy(arg), dtype, loc=loc, ip=ip)
+        else:
+            raise NotImplementedError(f"{cls.__name__} doesn't support wrapping {arg}.")
+        instance = Value.__new__(cls, val)
+        instance._arith_value = val
+        return instance
+
+    def __init__(self, arg, *, fold: Optional[bool] = None):
+        super().__init__(self._arith_value)
+        del self._arith_value
         self._fold = fold if fold is not None else False
-        super().__init__(val)

     def is_constant(self) -> bool:
         return isinstance(self.owner, OpView) and isinstance(self.owner, ConstantOp)
```

atmnp:
Yep, that's about the same shape as the patch I was expecting to make. Assuming it's green, pretty much exactly what we need. Would you be open to creating a PR with this?

lahwaacz:
Sure, I've opened https://github.com/NVIDIA/numba-cuda-mlir/pull/213 for this.

PR #191, opened 2026-07-06
Support dynamic getitem on tuples with tuple-typed elements

### LINKED ISSUE #190: Dynamic indexing into a tuple of tuples fails to compile
Indexing a tuple with a runtime value works for scalar elements but fails to compile when the elements are themselves tuples.

## Repro (main 7c5df73, RTX 4070 SUPER, CUDA 12)

```python
import numpy as np

ROWS = ((1.0, 2.0), (3.0, 4.0), (5.0, 6.0))

@cuda.jit
def k(out):
    i = cuda.grid(1)
    if i < 3:
        row = ROWS[i]
        out[i] = row[0] + row[1]

out = cuda.to_device(np.zeros(3))
k[1, 3](out)
```

## Expected behaviour

Compiles; `out` is `[3.0, 7.0, 11.0]`. numba-cuda supports this pattern (it's how we index a tuple of Butcher tableau rows with the loop variable).

## Observed behaviour

`ValueError: Result 0 of operation "scf.index_switch" must be a Type (bad cast)` at compile time.

## Cause

Dynamic tuple indexing emits a single-result `scf.index_switch` typed by `get_mlir_type(target_type)`. When the selected element is a tuple, `get_mlir_type` returns a tuple of types rather than one type, and `IndexSwitchOp` construction fails.

I've patched this by decomposing tuple-element selection into one scalar `index_switch` per leaf position, recursing through nested tuple types, and storing the result as a Python tuple of the switch results, which is how tuples are represented in the varmap elsewhere. Scalar element selection is unchanged — PR linked. Happy to rework or abandon if this collides with in-flight work.

ccam80:
Fixes #190.

Dynamic indexing into a tuple of tuples fails to compile; see the issue for the repro.

Changes in `lowering/numpy.py`:

- Tuple-element selection decomposes into one scalar `scf.index_switch` per leaf position, recursing through nested tuple types, and stores the result as a Python tuple of the switch results — the varmap convention for tuples.
- Scalar element selection is unchanged.

Tests: four additions to `tests/test_tuple.py` — runtime indexing into a tuple of tuples, a tuple of heterogeneous tuples, a deeply nested tuple, and a scalar-tuple regression guard for the unchanged path. The first three fail on main; all four pass with this change. Full test suite run against a main baseline with no new failures, on an RTX 4070 SUPER, CUDA 12.

---


src/numba_cuda_mlir/lowering/numpy.py
```diff
@@ -1376,28 +1381,42 @@ def lower_uni_tuple_getitem(builder, target, args, kwargs):
...
             tup = builder.lower_literal_if_needed(tup)
             index = index_of(index)
-            result_type = builder.get_mlir_type(target_type)
             error_memref = builder._get_or_create_error_global()
+            cases = ir.DenseI64ArrayAttr.get(range(len(tup)))
+
+            def select(candidates, element_type):
+                # candidates[i] is the value this selection yields when the
+                # runtime index equals i.
+                if isinstance(element_type, types.BaseTuple):
```
atmnp, line 1409: Empty tuple result emits no scf.index_switch, so no default arm, so no IndexError. We used to catch these, can we get that behavior back here?

tests/test_tuple.py
```diff
@@ -314,3 +314,59 @@ def kernel(out, flag):
 
 if __name__ == "__main__":
     test_tuple_concat((1, 2), (3, 4, 5))
+
```
atmnp: Can we just move these to above the conditional above? Just a minor style thing.

atmnp (commented):
Thanks so much for flagging this and taking the time to submit a PR! Just these two minor things and I'll approve.

src/numba_cuda_mlir/lowering/numpy.py
ccam80: Good catch, thanks for the check. Hoisting the bounds-check out of select and into an early zero-result index_switch protects the empty-tuple case, and as a bonus  it swaps two branches for `selp`s in the resulting PTX by removing the error write out of the default arms. Updated in e361281

tests/test_tuple.py
ccam80: Easy one, done.

atmnp (approved):

atmnp:
@ccam80 The test you introduced is failing, can we get that fixed so this can be merged?

ccam80:
@atmnp hoist by my own petard - the test relied on a raise but didn't run with debug=True, broken since #232/#225. Fixed now.

PR #195, opened 2026-07-06
Lower frozen array literals to constant globals instead of device malloc

### LINKED ISSUE #194: Kernels capturing a frozen numpy array call device malloc per thread and fault on large grids
A numpy array captured in a kernel closure is materialized with a per-thread device `malloc` that is never freed. The default device heap is 8 MB, so kernels fault with `CUDA_ERROR_ILLEGAL_ADDRESS` once the grid reaches tens of thousands of threads, and smaller grids leak an allocation per thread per launch.

## Repro (main 7c5df73, RTX 4070 SUPER, CUDA 12)

```python
import numpy as np

FVALS = np.array([1.5, -2.5, 4.0, 8.0], dtype=np.float32)

n = 1 << 18

@cuda.jit
def k(inp, out):
    i = cuda.grid(1)
    if i < out.size:
        out[i] = inp[i] + FVALS[i % 4]

inp = np.ones(n, dtype=np.float32)
out = cuda.to_device(np.zeros(n, dtype=np.float32))
threads = 256
k[(n + threads - 1) // threads, threads](cuda.to_device(inp), out)
print(out.copy_to_host()[:4])
```

## Expected behaviour

The launch completes; `FVALS` is read-only shared data and costs no per-thread work.

## Observed behaviour

`RuntimeError: Failed to synchronize CUDA context: an illegal memory access was encountered` at 256k threads. At smaller grids the kernel completes, but the PTX contains a `malloc` call per thread with no matching `free`.

## Cause

`lower_array_literal` builds the array element-by-element with `tensor.from_elements`, and bufferization materializes that as a heap `memref.alloc`, which lowers to device `malloc` in every thread. Nothing frees it.

I've patched this by lowering array literals whose storage width matches the numpy itemsize to an internal constant `llvm.mlir.global` carrying the array's raw bytes wrapped in a memref descriptor with the identity layout. One global is emitted per literal with a process-unique name so separately lowered functions link cleanly. Dtypes whose storage width differs from the numpy itemsize (and rank-0 arrays) keep the element-wise lowering — PR linked. Happy to rework or abandon if this collides with in-flight work.

ccam80:
Fixes #194.

Kernels capturing a frozen numpy array allocate it with a per-thread device `malloc` that is never freed, and fault with `CUDA_ERROR_ILLEGAL_ADDRESS` on large grids; see the issue for the repro.

Changes in `mlir_lowering.py`:

- Array literals whose storage width matches the numpy itemsize lower to an internal constant `llvm.mlir.global` carrying the array's raw bytes, wrapped in a memref descriptor with the identity layout. The byte payload is the same encoding string constants use; a dense-attribute initializer compiles but the global reads back zeroed after the LLVM70 translation.
- One global per literal with a process-unique name, so separately lowered functions link cleanly.
- Dtypes whose storage width differs from the numpy itemsize, and rank-0 arrays, keep the element-wise lowering.

Tests: `tests/test_array_literal_lowering.py` covers gather correctness through frozen index/value/2-D arrays, asserts the PTX contains the constant global and no malloc, and launches a 256k- thread grid. The PTX assertion and the large-grid launch fail on main; all five pass with this change. Full test suite run against a main baseline with no new failures, on an RTX 4070 SUPER, CUDA 12.

---


atmnp:
Can we also confirm that `complex64`/`complex128` work via tests? Those storage sizes should match the numpy item sizes right?

ccam80:
@atmnp I had omitted these from the fix completely - caught them now and added 2x regressions to confirm in 979f172

ccam80:
Looks like this is handled more cleanly in #251, closing as a duplicate. @atmnp don't hesitate to push back on any of my other PRs if I've missed the mark or if they're a bit muddy, I'm happy to rework or abandon them.

Issue #196, opened 2026-07-07
Perf: 6x compile time improvement for deep trees of inline="always" device functions


ccam80:
Compiling a kernel assembled from a deep tree of `inline="always"` device functions is dominated by frontend passes that scale superlinearly with the flattened function's size. This is present in base numba-cuda, but while migrating a GPU ODE integrator (kernels flatten to ~2000 blocks / ~90k SSA names) onto this backend I took a run at some compile time improvements. There's five independent patches, each take or leave - based on my view of the compiler frontend and the passing test suite, I think they are correctness-neutral. The toy repro goes from 132s to 22s, my production kernels go from ~3 hours to 40s.

## Repro

https://gist.github.com/ccam80/43118fd1d1cf1f11e2348f410378de70 — a tree of small `inline="always"` device functions, each called at two sites by its parent, so each unique callee is inlined at many transitive call sites. Self-validating against a NumPy evaluation of the same tree. `python stress_inline_chain.py 12 4` builds depth 12, width 4.

## Measurements

Compile plus first launch, 0.4.0 wheel, RTX 4070 SUPER, CUDA 12. Patches were applied as monkeypatches in selectable groups; leave-one-out marginals are against the all-patches configuration.

| configuration | seconds | vs all |
|---|---|---|
| all five patches | 21.76 | - |
| without lazy-postproc-liveness | 25.13 | 1.15x |
| without liveness-bitsets | 26.28 | 1.21x |
| without lazy-error-markup | 27.91 | 1.28x |
| without ssa-restricted-sweeps | 38.08 | 1.75x |
| without inline-callee-ir-cache | 123.55 | 5.7x |
| stock | 132.42 | 6.1x |

Which pass dominates depends on the CFG shape. On our kernel the ordering inverts: the liveness marginal grows to 7.2x and SSA to 2.0x, and any configuration without the inline cache did not finish inside 2.5 hours. The 3 hour blowup is due to the process running out of memory and spilling to virtual memory on disk, so indirectly due to pipeline-scaling.

The patches don't negatively impact low-inline workloads. The CI suite gets an unimpressive boost - it runs about 8% faster with all five patches on my machine: 334.9s stock, 307.2s patched.

## The patches

Each is a separate PR, independently vetoable, with tests where behaviour could regress:

- Inline callee-IR cache (#197): changes per-call-site
[...]

PR #197, opened 2026-07-07
Cache and structurally clone callee IR for inline="always" inlining

ccam80:
Part of #196. Largest single lever there: 123.55s to 21.76s on the repro (5.7x), and the difference between our production kernel not compiling inside 2.5 hours and compiling in 46s.

`InlineWorker.inline_function` re-runs the full untyped compiler pipeline for every call site of every inlinable callee. The pipeline itself contains `InlineInlinables`, so a tree of nested `inline="always"` functions recompiles its whole subtree at every transitive call site — the cost grows with total inlined sites, not unique functions.

Two commits, in `numba_cuda/core/inline_closurecall.py`:

- `inline_ir` gains a `preserve_ir` flag (default True) so callers that build a fresh callee IR per call can skip the defensive deep copy, and the redundant second copy used only to return an unmutated IR is removed. The same change is submitted to numba-cuda as numba/numba-cuda#689.
- Canonical callee IR is cached per (function, flags, enable_ssa) in a module-level weak dictionary, and each call site receives a structural clone. `copy.deepcopy` of numba IR is slower than re-running the pipeline, so the clone rebuilds scope, vars, statements and containers while sharing immutable leaves. This is performance-negative without the first `preserve_ir=False` commit - this splice regresses compile time by 2x if applied alone.

Behaviour is covered by the existing inlining tests. Full test suite run against a main baseline with no new failures, on an RTX 4070 SUPER, CUDA 12.

---


atmnp:
It's not clear to me that this avoids the fact that changing a global or closure cell used by an inline="always" callee, then recompiling its caller, might silently reuse old value. Can we add a test for this? Also, is it just me or does the the weak-key cache still leak because each cached FunctionIR strongly references its function, preventing the weak key from ever being collected.

ccam80:
The weak-key approach was an attempt to make the cache live at module/process scope, which no other numba-cuda caches do. Instead of fixing that approach it looks cleaner to shift ownership down to one cache per compile call. This catches global/closure changes and drops the leaky weak-key requirement as the whole thing gets GC'd between compiles. It gives up reuse across specializations, but ends up performing faster than the weak-key approach on the repro in #196. 

I've added some tests to cover the cases mentioned, please point me at any use cases I've missed that might break this approach.

PR #198, opened 2026-07-07
Run liveness fix points on bitsets in reverse sweep order

ccam80:
Part of #196: 26.28s to 21.76s on the repro (1.21x). This one is shape-dependent — on our production kernel (~2000 blocks, ~90k SSA names, dense reachability) it is the second-largest lever at 7.2x: per `compute_live_map` call, 0.8s on bitsets against 98s on sets, three calls per compile.

`compute_live_map` in `numba_cuda/core/analysis.py` assigns every variable a bit index and runs both dataflow fix points on arbitrary-size integers, replacing hash-set element traversal with bignum bitwise operations. Successor and predecessor lists are hoisted out of the loops, and the backward liveness sweep visits blocks in reverse label order — numba block labels ascend roughly topologically, so the reverse order converges in a couple of sweeps where ascending order needed one sweep per block of CFG depth. Results are identical sets; only the fix-point representation changes.

Tests: `tests/test_liveness_bitsets.py` checks equivalence against a direct transcription of the set-based fix-point algorithm kept in the test as the oracle: every 2-block universe over 2 variables (all edge subsets x all use/def assignments), every 3-block digraph, variable counts straddling bitset byte boundaries, and a deep chain with a backedge. Full test suite run against a main baseline with no new failures, on an RTX 4070 SUPER, CUDA 12.

---


atmnp (approved):
Thanks for fixing this! It was inherited from the numba analysis pass, and is obviously against the standard of using bitsets to do fix point analysis.

PR #201, opened 2026-07-07
Build error markup without re-wrapping the terminal streams

ccam80:
Part of #196: 27.91s to 21.76s on the repro (1.28x).

Typing speculatively instantiates thousands of `NumbaError` objects per successful compile, and `HighlightColorScheme._markup` produced each message via a colorama ColorShell/reset_terminal round-trip that inits and deinits the standard-stream wrappers on every call. The markup is now assembled directly with an explicit `RESET_ALL` terminator, in `numba_cuda/core/errors.py`, matching the previous output byte for byte.

Output is pinned by the existing error-formatting tests. Full test suite run against a main baseline with no new failures, on an RTX 4070 SUPER, CUDA 12.

---


src/numba_cuda_mlir/numba_cuda/core/errors.py
```diff
@@ -323,11 +323,7 @@ def _markup(self, msg, color=None, style=Style.BRIGHT):
                 features += color
             if style:
                 features += style
-            with ColorShell():
```
atmnp, line 326: Can we remove these dead imports? There should be a few now right.

atmnp (approved):

src/numba_cuda_mlir/numba_cuda/core/errors.py
ccam80: You're right, that was a bit sloppy - removed them in 3c2e26d

PR #202, opened 2026-07-07
Add post-inline whole-function planners

### LINKED ISSUE #62: Rerun before-inference rewrites after device-function inlining
## Problem

Registered `before-inference` rewrites currently run before inline-always device functions have been merged into the caller IR. If a supported rewrite target is inside a device helper, the initial rewrite pass sees only the outer call. After the inliner merges the helper body into the kernel, the newly exposed IR is not passed back through the same rewrite registry.

This blocks extension and cooperative-code patterns where users put generated primitive calls behind small device helpers for readability or reuse.

Depends on #59 for the launch-context metadata used by launch-sensitive rewrites.

## Example

```python
import numpy as np

# Pseudocode: this marker stands in for a call transformed by a registered
# before-inference rewrite.
def rewrite_marker(x):
    return x

@cuda.jit(device=True, inline="always")
def helper(x):
    return rewrite_marker(x)

@cuda.jit
def kernel(inp, out):
    out[0] = helper(inp[0])

inp = np.asarray([7], dtype=np.int32)
out = np.asarray([0], dtype=np.int32)
kernel[1, 1](inp, out)
```

Expected behavior is that after `helper()` is inlined, the rewritten marker call in the caller IR is visible to registered `before-inference` rewrites. Without rerunning rewrites after inlining, the marker remains in the IR and later typing/lowering can fail.

## Expected behavior

When `InlineInlinables` modifies the caller IR, it should clean up the control-flow graph as it already does, then reapply registered `before-inference` rewrites to the modified state before type inference proceeds.

This should only happen when inlining actually changed the IR.

## Implementation notes

The local fix has already been implemented and a PR will be forthcoming.

The implementation invokes `rewrites.rewrite_registry.apply("before-inference", state)` after inline cleanup when `InlineInlinables` reports modifications. The local implementation is currently represented by commit `93167c3` (`Rerun rewrites after device inlining`) on the cooperative CUDA work branch.

Tests should cover:

- an inline-always device function containing a rewrite target;
- the target is rewritten after the helper body is merged into the kernel;
- the rewrite can observe launch-context metadata when appropriate;
- no behavior change when no inlining occurs.

Issue created with GPT-5.5 xhigh, Codex 0.131.0.

tpn:
## Overview

Add an opt-in whole-function planner extension point immediately after device-function inlining and before type inference.

- Export `WholeFunctionPlanner` and `register_planner` from `numba_cuda_mlir.extending`.
- Run registered planners in deterministic order using a thread-safe per-pass snapshot.
- Repair post-inline CFG and analysis state before the first planner, then repair again after each planner that mutates IR so every planner sees a coherent function.
- Expose kernel-versus-device-function classification through `is_device_function`.
- Bypass persistent cache load/save while planners are registered so cached code cannot skip planner transformations.
- Serialize persistent-cache hit publication and save decisions with planner registration, while leaving compilation itself unlocked so planners can register for the next compilation.
- Document the public contract and API.

This provides the post-inline visibility needed by extensions without rerunning every before-inference rewrite.

## Validation

- Fresh-head planner/dispatcher suite: 60 passed on SM120.
- Fresh-head changed-file pre-commit and git diff checks passed.
- Exact-head GitHub Actions run [30037569855](https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/30037569855) completed green at `b3449050d23c9b0238e496b4ad164f6cee63181a`: 124 successful jobs, 4 skipped, zero failed or pending.
- Focused inline, extending, LTO, and compile-device regressions: 48 passed.
- Full Sphinx HTML build succeeded without warnings.

Closes #62.

Co-authored-by: Codex 5.6 Sol Ultra Fast

### DIFF
--- a/docs/source/extending/low-level.rst
+++ b/docs/source/extending/low-level.rst
@@ -68,6 +68,10 @@ planner, start with the next compilation. Planners operate on untyped Numba
 IR; typing and MLIR lowering extensions remain responsible for the later
 compilation phases.
 
+Persistent disk-cache load and save are bypassed while planners are
+registered because planner identity and implementation are not part of the
+cache key. In-memory overload reuse is unchanged.
+
 .. autoclass:: numba_cuda_mlir.extending.WholeFunctionPlanner
    :members: is_device_function, run
 
--- a/docs/source/extending/low-level.rst
+++ b/docs/source/extending/low-level.rst
@@ -42,11 +42,11 @@ caller and before type inference. This makes calls introduced by device
 helpers visible without rerunning unrelated rewrite registries.
 
 Implement ``run()`` and return ``True`` only when the planner changed
-``state.func_ir``. After each change, Numba-CUDA-MLIR simplifies the control
-flow graph, refreshes postprocessing state and definitions, and verifies every
-block before invoking the next planner. Planners run for both kernels and
-device-function compilations; a planner that only applies to kernels should
-check ``self.is_device_function`` and decline device functions.
+``state.func_ir``. Before the first planner and after each change,
+Numba-CUDA-MLIR simplifies the control flow graph, refreshes postprocessing
+state and definitions, and verifies every block. Planners run for both kernels
+and device-function compilations; a planner that only applies to kernels
+should check ``self.is_device_function`` and decline device functions.
 
 .. code-block:: python
 
--- a/docs/source/extending/low-level.rst
+++ b/docs/source/extending/low-level.rst
@@ -68,6 +68,11 @@ planner, start with the next compilation. Planners operate on untyped Numba
 IR; typing and MLIR lowering extensions remain responsible for the later
 compilation phases.
 
+Register every planner before compiling a dispatcher that needs it.
+Registration does not invalidate an overload that the dispatcher has already
+compiled and cached in memory; a planner registered later applies only when a
+new overload is compiled.
+
 Persistent disk-cache load and save are bypassed while planners are
 registered because planner identity and implementation are not part of the
 cache key. In-memory overload reuse is unchanged.
--- a/docs/source/extending/low-level.rst
+++ b/docs/source/extending/low-level.rst
@@ -7,8 +7,11 @@
 Low-level extension API
 =======================
 
-The low-level API exposes Numba-CUDA-MLIR's two compilation phases directly:
+The low-level API exposes Numba-CUDA-MLIR's extension phases directly:
 
+* **Planning** — after device functions have been inlined and before type
+  inference, whole-function planners can inspect and transform the complete
+  untyped Numba IR.
 * **Typing** — running before code generation, type inference resolves each
   variable, attribute, and call to a type. Extension authors register new
   typing rules with the :py:mod:`~numba_cuda_mlir.numba_cuda.typing` template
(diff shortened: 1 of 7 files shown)


src/numba_cuda_mlir/_whole_function_planners.py
```diff
@@ -0,0 +1,87 @@
...
+        for planner_cls in planners:
+            result = planner_cls(state).run()
+            if not isinstance(result, bool):
+                raise TypeError(
+                    f"{planner_cls.__name__}.run() must return bool, got {type(result)!r}"
+                )
+            if result:
+                self._repair_ir(state.func_ir)
+                modified = True
+        return modified
```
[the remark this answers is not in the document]
tpn: Fixed in 27b60db. The registry now performs the full CFG/postprocessing/definition repair before the first planner, not only after a planner mutates IR. The compile-only regression checks the first inspection-only planner against a freshly rebuilt definition map by object identity and verifies block-entry analysis matches the live blocks.

docs/source/extending/low-level.rst
```diff
@@ -17,18 +20,63 @@ The low-level API exposes Numba-CUDA-MLIR's two compilation phases directly:
...
+           return bool(modified)
+
+Registration order is preserved, and registering the same planner class more
+than once has no effect. The planner pass snapshots its registry before
+running, so planners registered after that snapshot, including from a running
+planner, start with the next compilation. Planners operate on untyped Numba
+IR; typing and MLIR lowering extensions remain responsible for the later
+compilation phases.
+
+Persistent disk-cache load and save are bypassed while planners are
```
atmnp, line 76: Is the contract: "register all planners before any relevant dispatcher compiles."? Otherwise, I worry that a user could register a planner *after* compiling, and it'd keep re-using the same overload from the first compile before the planner was registered.

tpn: Yes. Commit 3cc951c makes this contract explicit in both the public guide and register_planner docstring: every planner must be registered before a dispatcher that needs it is compiled. Registration does not retroactively invalidate an already-compiled in-memory overload; a later registration applies when a new overload is compiled. The refreshed branch is rebased onto current main (80a61b6), and the planner/inlining/extending/LTO focused set passes 48/48, including the real-GPU inline-device planner test.

atmnp (approved):
Looks great! Clear, would love to see how we plan on using this feature.

tpn:
@atmnp Thanks! The immediate consumer is CCCL's `cuda.coop.numba_mlir`. It registers its planner when the module is imported—before relevant dispatchers compile—and uses the hook to lower cooperative operations after device helpers have been inlined.

A concrete user shape from the current prototype is:

```python
import cuda.coop.numba_mlir as coop

@cuda.jit(device=True, inline="always")
def block_sum(value):
    return coop.reduce(coop.this_block(), value)

@cuda.jit
def kernel(values, totals):
    tid = cuda.threadIdx.x
    totals[tid] = block_sum(values[tid])
```

The extension side is conceptually:

```python
from numba_cuda_mlir.extending import (
    WholeFunctionPlanner,
    register_planner,
)

@register_planner
class CoopPlanner(WholeFunctionPlanner):
    def run(self) -> bool:
        modified = rewrite_coop_calls(self.state.func_ir)
        return bool(modified)
```

By the time `CoopPlanner` runs, `block_sum()` has been inlined, so the planner sees `this_block()` and `reduce()` in the complete kernel IR. It can resolve the group/provider, replace those markers with the selected device implementation, and let the existing link path carry that provider's LTO-IR into compilation.

That is the main motivation for the hook: users can keep cooperative logic factored into Python device helpers while we still do whole-kernel planning, preserve inlining, and avoid runtime dispatch or rerunning unrelated rewrite passes. The follow-on work adds launch-qualified planning, dynamic shared-memory requirements, and literal specialization for the richer group-first API.

PR #204, opened 2026-07-07
Add llvm.intrin.assume to downgrade tool

atmnp:


atmnp:
Closing in favor of #206

PR #212, opened 2026-07-10
Preserve numpy integer scalar argument width on lazy-JIT launch path

jiel-nv:
Scalar `int8`/`int16`/`int32` (and their unsigned counterparts) kernel arguments were reported as `int64` in CUBIN DWARF debug info when the kernel was compiled via the implicit / lazy-JIT path (a bare `@cuda.jit` with no explicit signature and no type annotations). 

The problem: the dispatcher was widening every numpy integer scalar to a 64-bit Python `int` before type inference, so `typeof()` saw `int64`, and the compiled parameter, together with its debug info, became `int64`.

After this change, numpy integer scalar arguments keep their native width through the lazy-JIT path, mirroring how `np.float16`/`np.float32` are already preserved. The DWARF now reports the correct `int8`/`int16`/`int32`/`int64` (and unsigned) base types. Python `int` arguments continue to map to `int64`, unchanged.

### DIFF
--- a/cext/launcher/kernel.cpp
+++ b/cext/launcher/kernel.cpp
@@ -53,6 +53,11 @@ PyTypeObject* g_cuda_core_Stream_type;
 
 PyTypeObject* g_enum_Enum_type;
 
+// numpy integer abstract base type; matches every signed/unsigned numpy
+// integer scalar (int8/16/32/64, uint8/16/32/64, including aliases such as
+// numpy.longlong).
+PyTypeObject* g_numpy_integer_type;
+
 constexpr uint8_t BYTE_BITWIDTH = 8;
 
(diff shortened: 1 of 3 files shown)


jiel-nv:
Beyond fixing the DWARF correctness issue, this PR might give a per-launch dispatch speedup as a side benefit: it restores a fast path that numpy integer arguments previously disabled.

The dispatcher caches results keyed by the argument type tuple. A cache entry is reusable only when every argument is handed to the launcher unchanged.

Before the fix, a numpy integer scalar was widened and produced a new object, which marked the argument as changed and left the cache entry non-reusable.

After the fix, the numpy integer scalar passes through untouched, so repeated launches with the same argument types reuse the cache.

atmnp (approved):

PR #213, opened 2026-07-10
Fix nanobind 2.13 compatibility for ArithValue wrappers

lahwaacz:
nanobind 2.13 tightened `nb_type_check()` so that a Python subclass of a nanobind-bound type must have the exact nanobind-generated metaclass (`type(Value)` for MLIR's `Value` wrapper). Derived metaclasses such as `ArithValueMeta` are no longer accepted, which breaks construction of `ScalarValue`/`TensorValue` and any code that goes through the arith dialect extensions on systems using nanobind 2.13.

Remove the custom ArithValueMeta and instead move the wrapping logic into `ArithValue.__new__` so that nanobind's own instance-creation path is used; this path knows how to initialize a Python subclass of a nanobind-bound `Value` correctly, while calling `cls.__init__` explicitly from a Python metaclass does not.

Update `ArithValue.__init__` to accept the same keyword arguments that the old metaclass forwarded (dtype, fold, loc, ip).

This keeps the same public behavior (wrapping Values, Operations, OpViews, and Python scalars/arrays) while restoring compatibility with nanobind 2.13.

Related to #173 and #183

### DIFF
--- a/ci/windows-llvm-container-build.ps1
+++ b/ci/windows-llvm-container-build.ps1
@@ -34,8 +34,7 @@ Write-Host "Using Python: $pythonExe"
 & $pythonExe -m pip install --upgrade pip
 
 if ($Mode -eq 'modern') {
-    # nanobind 2.13 breaks the pinned MLIR Python bindings; see issue #173.
-    $pkgs = @('pybind11', 'nanobind==2.12.0', 'numpy', 'ninja', 'cmake', 'awscli')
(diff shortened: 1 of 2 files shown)


atmnp:
Could you please also remove the version pin for nanobind in `ci/windows-llvm-container-build.ps1:38`?

lahwaacz:
@atmnp Done :+1:

atmnp (approved):
LGTM, Thanks so much for submitting this!

PR #217, opened 2026-07-12
Fix llvm70 typos and lit-checks

ccam80:
The compile-unit's producer string output was one character short, storing "llvm7" instead of "llvm70". It now matches the "llvm70" expected by debuginfo.mlir.

debuginfo.mlir expected `emissionKind: DebugDirectviesOnly`. LLVM 7 has no name for emission kind 3 (the name was added in LLVM 8) and prints it as empty, so this check could never pass. The check now expects the empty form.

matmul.mlir used `nvvm.barrier0`, which the bundled MLIR no longer parses. It now uses `nvvm.barrier`, which lowers to the same instruction.

**Verification:** (llvmorg-7.1.0, Windows) the 2 failing `--dump-llvm` RUN lines in `cext/mlir-llvm70/test` go red->green; the other 78 are unaffected.


atmnp (approved):
LGTM (assuming green). Thanks @ccam80!

PR #221, opened 2026-07-12
Restore selective fastmath support via per-op MLIR fastmath attributes

### LINKED ISSUE #220: fastmath has almost no effect on generated code, and selective flag sets are rejected
**Describe the bug**

`fastmath=True` compiles most operations to the same instructions as precise code: no approximate division, no approximate transcendentals. The selective form numba-cuda accepts (`fastmath={"arcp"}`) raises `TypeError`.

```python
from numba_cuda_mlir import compiler, types

def kernel(r, x, y):
    r[0] = x / y

sig = types.void(types.float32[::1], types.float32, types.float32)
ptx, _ = compiler.compile_ptx(kernel, sig, fastmath=True)
print([l for l in ptx.splitlines() if "div" in l])
compiler.compile_ptx(kernel, sig, fastmath={"arcp"})  # TypeError
```

| op | expected |
|---|---|
| `a / b` (f32) | `div.approx.ftz.f32` |
| `sin/log/exp` (f32) | `sin.approx` etc. |
| `tanh` (f32, sm_75+) | `tanh.approx.f32` |
| `fastmath={"arcp"}` | accepted, approximate division |

**Observed behavior** (sm_89)

`div.rn.f32`, precise libdevice calls for all transcendentals, and `TypeError` for the set form.

**Root cause**

`fastmath` is typed bool and consumed only as module-wide options (the target's `fast` flag and the libnvvm/ptxas knobs). No operation ever carries a per-operation fastmath attribute, so every effect LLVM gates per-instruction is lost. The module knobs cannot recover approximate division; libnvvm's `-prec-div=0` produces `div.full`, never `div.approx`. On GPUs below sm_100 the module-level options are not forwarded to libnvvm at all.

**Environment details**

Built from source on Windows; observed on sm_89.

ccam80:
Fixes #220.

`fastmath` accepted only a bool; it now accepts `bool | set | dict | FastMathOptions` over the eight LLVM flags, validated at decoration time (new module `fastmath.py`).

Each lowered function is walked once and every fastmath-capable `arith`/`math` op is stamped with `#arith.fastmath<...>`. The capable-op set is discovered from the generated bindings, so it tracks the bundled MLIR. Device callees compile under their own options and are cloned in already stamped, so flags scope per function.

Two rewrites cover instructions the flags alone cannot produce, both the same substitutions numba-cuda makes: f32 division under `arcp` or `fast` becomes `__nv_fast_fdividef`, and f32 `tanh` under `afn` or `fast` on sm_75+ becomes `tanh.approx.f32`. Constant-zero powers are folded to 1.0 before lowering; `__nv_fast_powf` returns NaN for negative, NaN, zero, and infinite bases.

The module-level libnvvm/ptxas knobs were all-or-nothing off the bool. Each is now derived from the flag that requests it, and they now reach libnvvm on the sm<100 path through a new trailing CAPI parameter.

The LLVM 7 C API cannot set per-instruction fast-math flags. On the sm<100 path, flagged instructions get marker value names, the flag keywords are inserted into the printed IR, and the text is passed to libnvvm, whose parser accepts it directly. It also names the compile unit's emission kind 3, which LLVM 7 prints empty and cannot re-parse; any substitute kind it can parse makes libnvvm emit full DWARF, which ptxas refuses to link above -O0. The tests check that debug output, including `.target`, is unchanged, and that the PTX and LTO links succeed under `lineinfo`.

**Tests**

- `test_fastmath.py`: `{"arcp"}` gives `div.approx` on both compile paths, `{"nsz"}` keeps precise division, `{"afn"}` gives `sin.approx` without approximate division, invalid flags are rejected, per-op attributes appear in the MLIR, the knob mapping is unit-tested, and fastmath with `lineinfo`/`debug` keeps `div.approx` with identical debug directives.
- `fastmath_flags.mlir` (new lit test): full, selective, and unflagged instructions survive the sm<100 text path.
- `test_math.py`: `x ** (0.0)` and `x ** (-0.0)` return 1.0 for NaN, +-0.0, +-inf, and -2.0 bases under `fastmath=False`, `{"afn"}`, and `True`; the fold is unit-tested on a parsed module.

**Verification:** (sm_89, Windows) 24/26 xfailed fastmath tests go red->green, 2 stay xfailed (fastmath doesn't propagate into device functions from decorated kernel, no ZeroDivisionError); GPU fast-vs-precise results agree to relative error 7e-8.

### DIFF
--- a/cext/mlir-llvm70/include/llvm70/LLVM70IRBuilder.h
+++ b/cext/mlir-llvm70/include/llvm70/LLVM70IRBuilder.h
@@ -248,12 +248,6 @@ public:
   size_t getBufferSize(LLVMMemoryBufferRef buf);
   void disposeMemoryBuffer(LLVMMemoryBufferRef buf);
 
-  /// Replace the current module with one parsed from textual IR (in the same
-  /// context). Used to round-trip fast-math flag injection, which the old C
-  /// API cannot express directly. Only valid once construction is complete;
-  /// existing LLVMValueRefs into the old module become invalid.
-  llvm::Error replaceModuleWithParsedIR(llvm::StringRef ir);
-
 private:
   LLVM70IRBuilder() = default;
   llvm::Error resolveSymbols();
@@ -307,10 +301,6 @@ private:
   LLVM_FN(unsigned, fnCountParams, LLVMValueRef)
   LLVM_FN(void, fnSetValueName2, LLVMValueRef, const char *, size_t)
   LLVM_FN(LLVMValueRef, fnIsAInstruction, LLVMValueRef)
-  LLVM_FN(LLVMMemoryBufferRef, fnCreateMemoryBufferWithMemoryRangeCopy,
-          const char *, size_t, const char *)
-  LLVM_FN(LLVMBool, fnParseIRInContext, LLVMContextRef, LLVMMemoryBufferRef,
-          LLVMModuleRef *, char **)
 
   // Basic blocks
   LLVM_FN(LLVMBasicBlockRef, fnAppendBB, LLVMContextRef, LLVMValueRef,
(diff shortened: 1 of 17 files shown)


ccam80:
Superseded by #255, which is a more complete and slightly cleaner implementation that pipes these changes through the LTO path.

PR #223, opened 2026-07-13
Make the SSA reaching-definition search iterative

### LINKED ISSUE #222: RecursionError compiling kernels with large control-flow graphs
**Describe the bug**

Kernels whose control-flow graphs contain long dominator chains fail to compile with `RecursionError` raised inside the SSA pass. Large fully-inlined device functions hit this easily, because inlining flattens the callee's blocks into the caller's CFG before SSA runs.

```python
from numba_cuda_mlir.numba_cuda.compiler import run_frontend
from numba_cuda_mlir.numba_cuda.core.ssa import reconstruct_ssa

# `a` has two definitions; its use sits at the far end of a long
# immediate-dominator chain built from unrelated conditionals.
n = 500
src = ["def deep_cfg(x):", "    a = 0", "    if x < 0:", "        a = 1"]
for i in range(n):
    src += [f"    if x > {i}:", f"        b{i} = x"]
src.append("    return a")
ns = {}
exec("\n".join(src), ns)

func_ir = run_frontend(ns["deep_cfg"])   # ~1000 blocks
reconstruct_ssa(func_ir)                 # RecursionError
```

**Expected behavior**

Compilation succeeds for any CFG that fits in memory. The SSA pass does not depend on the interpreter recursion limit.

**Observed behavior**

`RecursionError` at CPython's default 1000-frame limit once a use sits about 500 blocks from its reaching definition. The workaround is `sys.setrecursionlimit` plus compiling inside a thread with an enlarged stack.

**Root cause**

`reconstruct_ssa` resolves each variable's reaching definition with a mutual recursion between `_find_def_from_top` and `_find_def_from_bottom`, costing two Python frames per CFG block walked.

**Environment details**

Built from source on Windows.

ccam80:
Fixes #222.

The reaching-definition search recursed once per CFG block, through a mutual recursion between `_find_def_from_top` and `_find_def_from_bottom`, so large control-flow graphs overflowed the interpreter recursion limit. `_find_def_from_top` now drives the search as a loop: `_walk_def_chain` walks one dominator chain, and phi incoming values are deferred onto an explicit stack and resolved in the same depth-first order as before, so phi creation order and fresh-variable numbering are unchanged. `_find_def_from_bottom` had no callers left and is removed.

**Tests**

- `test_ssa.py` gains a deep-CFG test: a 500-conditional dominator chain reconstructs under a recursion limit a few hundred frames above the test's own stack depth and yields the expected single two-input phi.

**Verification:** (sm_89, Windows) the new test goes red->green; the previous implementation raises `RecursionError` on the same CFG at the default limit. Frame depth no longer grows with CFG size (a 4000-block CFG reconstructs at the current stack depth plus 100). With the old implementation patched back in, old and new produce identical IR over if/else, loops, nested loops, while/break/continue, and uninitialized-variable shapes. `test_ssa.py` passes and SSA reconstruction time is unchanged within noise.

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/core/ssa.py
+++ b/src/numba_cuda_mlir/numba_cuda/core/ssa.py
@@ -415,58 +415,101 @@ class _FixSSAVars(_BaseHandler):
         This method would look at all dominance frontiers.
         Insert phi node if necessary.
         """
-        _logger.debug("find_def_from_top label %r", label)
+        return self._find_def_iteratively(states, label, loc, from_top=True)
+
+    def _find_def_from_bottom(self, states, label, loc):
+        """Find definition from within the block at ``label``."""
+        return self._find_def_iteratively(states, label, loc, from_top=False)
+
+    def _find_def_iteratively(self, states, label, loc, from_top):
+        """Iterative driver for the reaching-definition search.
+
+        Equivalent to a mutual recursion between the *from_top* and
+        *from_bottom* searches, but implemented with an explicit worklist
+        so that the search depth is bounded by available memory rather
+        than the interpreter recursion limit.  The dominator chains
+        walked here grow with the size of the function's CFG, so large
+        functions (e.g. kernels with big fully-inlined callees) would
+        otherwise overflow the recursion limit.
+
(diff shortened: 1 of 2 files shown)


atmnp:
@coderabbitai review

PR #226, opened 2026-07-13
Allow importing `cext` on CPU only machines

brandon-b-miller:
This PR fixes an issue exposed by import changes between `0.3.0` and `0.4.1` where the package can't import due to a missing driver on CPU only machines. It seems to come down to `cext` not being importable, I think the best change might be to fix it at the c++ level so that another quiet import change that hits `cext` doesn't reintroduce this later. That is what is proposed here.

### DIFF
--- a/cext/launcher/cuda_helper.cpp
+++ b/cext/launcher/cuda_helper.cpp
@@ -15,6 +15,9 @@ const char* get_cuda_error(CUresult res) {
 }
 
 PyObject* get_max_grid_size(PyObject*, PyObject *args) {
+    if (!ensure_cuda_available())
+        return NULL;
+
     int device_id;
     if (!PyArg_ParseTuple(args, "i", &device_id))
         return NULL;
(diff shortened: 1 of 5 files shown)


atmnp (approved):

isVoid:
Thanks for fixing and merging this. We hit this exact issue in NVSHMEM's
`create_source_package` CI: Numbast's static MLIR binding generator runs on a
CPU-only worker with no NVIDIA runtime, and the current `0.4.1` wheel raises
while importing `_cext`.

I validated the wheel artifact from this PR with the CUDA toolkit stub forced
in place of a working driver. In that environment:

- `numba_cuda_mlir._cext`, `MLIRLower`, and Numbast's static generator import
  successfully;
- `_cext.cuda_available` is `0` and a device query raises the expected
  controlled `RuntimeError`; and
- the full NVSHMEM static binding generation completes successfully (2,777
  declarations, formatted output, exit code 0).

Would it be possible to cut a patch release (for example, `0.4.2`) containing
#226? That would let NVSHMEM consume the fix and keep source-package creation on
CPU-only workers.

PR #228, opened 2026-07-14
Add regression test for nested scalar sub-record field read (#214)

### LINKED ISSUE #214: [BUG] Record types still yield illegal memory access
We're seeing crashes when you try to access a field within a record. Lmk if you'd prefer me reopen 156 instead. Reproducer below. Gives: `Failed to synchronize CUDA context: an illegal memory access was encountered`

This monstrosity, with a call added, would be a pretty good torture test of this machinery: https://github.com/NVIDIA/numba-cuda/issues/876

```
"""Minimal reproducer: numba_cuda_mlir crashes on nested record field access.

numba_cuda_mlir 0.4.1, numba 0.65.0
Accessing a field of a sub-record (one level of nesting) causes:
  RuntimeError: Failed to synchronize CUDA context: an illegal memory access was encountered

Flat records work fine. The bug is specifically in nested struct lowering.
"""
import numba
import numpy as np

# One level of nesting: outer.inner.a
INNER_T = [('a', np.float64), ('b', np.float64)]
OUTER_T = np.dtype([('inner', INNER_T), ('scale', np.float64)])
OUTER_NB = numba.from_dtype(OUTER_T)

sig = [numba.void(OUTER_NB[::1], numba.types.Array(numba.float64, 1, 'C'))]

@cuda.jit(signature=sig)
def kernel(rec, out):
    i = cuda.grid(1)
    if i < rec.shape[0]:
        out[i] = rec[i]['inner']['a'] * rec[i]['scale']

rec = np.zeros(1, dtype=OUTER_T)
rec['inner']['a'] = 5.0
rec['scale'] = 2.0
out = np.zeros(1, dtype=np.float64)

kernel[1, 1](rec, out)
cuda.synchronize()

print(f"Expected: 10.0, Got: {out[0]}")
assert np.allclose(out, 10.0)

```

CodersAcademy006:
## Summary

Revised: this PR is now **test-only**.

I originally proposed a code change (`_is_record_type`) believing #214 was still unfixed on `main`. On verification that is not the case — **#160 (d83d2f2) already fixes the reporter's exact path.** I confirmed it on an NVIDIA P100 with the issue's exact `numba.from_dtype` reproducer under `compute-sanitizer`:

| Build | Result |
|---|---|
| `numba-cuda-mlir==0.4.1` (pre-#160, what the reporter ran) | illegal `__global__` read of size 8 @ `0x4014000000000000` (the IEEE-754 bits of `5.0`) |
| `main` + #160 (no other change) | returns `10.0`, **0 sanitizer errors** |

So the runtime is already correct. What was missing is coverage: the existing record tests exercise the nested **array** read (`test_read_tuple_indexed_nested_record_field`) and the nested **write** (`test_set_nested_record_fields`), but not the **scalar sub-record read at a runtime index** (`rec[i]["inner"]["a"]`) — the precise shape from #214. That gap is what let the pre-#160 regression go unnoticed.

This PR adds one device test for exactly that path so it can't silently regress. No production code changes.

Closes #214 (fixed by #160; this locks it with coverage).


CodersAcademy006:
The four red `cudf source` jobs are not from this change. All four (linux-64 and linux-aarch64, CUDA 12.9.1 and 13.2.1) die identically in setup, before any test runs: the toolkit fetch reports `digest-mismatch: error` and the job ends with `No files were found with the provided path: cudf-source-junit-results.xml`, so no test ever executed. A uniform failure across both architectures and both CUDA versions at the fetch step is infrastructure, and this PR only adds a regression test file plus its entry, with no runtime or build change that cudf could see.

@rparolin could you re-run the `cudf source` matrix when convenient? Everything else on the PR is green (117 passing).

atmnp (approved):

PR #230, opened 2026-07-14
Use maxnum semantics for float min and max

### LINKED ISSUE #229: Built-in float min and max return NaN when either operand is NaN
**Describe the bug**

Python built-in `max` and `min` on device float operands return NaN whenever either operand is NaN. numba-cuda returns the non-NaN operand.

```python
import numpy as np

@cuda.jit
def kernel(a, b, c):
    i = cuda.grid(1)
    c[i] = max(a[i], b[i])

a = np.array([1.0, np.nan, 1.0, np.nan], dtype=np.float32)
b = np.array([2.0, 2.0, np.nan, np.nan], dtype=np.float32)
c = np.zeros(4, dtype=np.float32)
kernel[1, 4](a, b, c)
print(c)
print(np.fmax(a, b))
ptx = next(iter(kernel.inspect_asm().values()))
print([line.strip() for line in ptx.splitlines() if "max" in line])
```

`[2. 2. 1. nan]` and `max.f32` in the PTX. numba-cuda lowers float `max` and `min` to libdevice `__nv_fmax`/`__nv_fmin` (`cudaimpl.py`), which return the non-NaN operand.

**Observed behavior** (sm_89)

`[ 2. nan nan nan]`. The PTX contains `max.f32` followed by `setp.nan.f32`, `add.f32`, and `selp.f32`, which replace the result with NaN when either operand is NaN.

**Root cause**

`lower_max` and `lower_min` use `arith.maximumf` and `arith.minimumf`, the IEEE-754 maximum and minimum operations, which propagate NaN. On sm_89 they compile to `max.f32` plus the compare-and-select above; on sm_100 and sm_120 they compile to `max.NaN.f32`. The operations with numba-cuda's semantics are `arith.maxnumf` and `arith.minnumf`. The vendored numba-cuda test checks for `max.f32` in the PTX, but the numeric inputs contain no NaN and the instruction check accepts the `max.NaN.f32` variant, so the difference was invisible to it.

ccam80:
Fixes #229.

Float `min` and `max` lowered to `arith.minimumf` and `arith.maximumf`, the IEEE-754 operations that propagate NaN. They now lower to `arith.minnumf` and `arith.maxnumf`, which return the non-NaN operand, matching numba-cuda's libdevice `__nv_fmin`/`__nv_fmax` lowering, and compile to the plain min/max PTX instructions without the NaN compare-and-select expansion.

The vendored min/max tests compared against `np.maximum`/`np.minimum` over inputs with no NaN and accepted `max.NaN.f32` variants in the PTX check. They now compare against `np.fmax`/`np.fmin` over inputs covering either operand NaN, both NaN, and signed zeros, check the sign of zero results, and reject the `.NaN.` instruction variants.

**Tests**

- `tests/numba_cuda_tests/cudapy/test_minmax.py`: all eight float min/max tests now cover NaN and signed-zero operands for f32/f64 and mixed widths.

**Verification:** (sm_89, Windows) the 8 tests go red->green; on main `max(1.0, nan)` returns nan.


atmnp (changes requested):
Numpys fmax and fmin, and PTX max/min have different signed zero behaviors, could you address this in the tests please? This is the cause of the current failures. Thanks for the PR!

atmnp (approved):

PR #233, opened 2026-07-15
Gate the post-launch error-code readback on the debug flag

### LINKED ISSUE #232: All kernels launch synchronously
**Describe the bug**

Every kernel launch blocks until the kernel completes, so host/device overlap and multi-stream concurrency are impossible. Streams, events, and async copies behave normally; only launches are synchronous.

```python
import time

import numpy as np

@cuda.jit
def busy(out):
    x = out[0]
    for _ in range(20_000_000):
        x = x * 0.9999999 + 1.0
    out[0] = x

stream = cuda.stream()
busy[1, 1, stream](out)   # compile + first run
cuda.synchronize()

event = cuda.event()
t0 = time.perf_counter()
busy[1, 1, stream](out)
print(f"launch call took {time.perf_counter() - t0:.3f}s")
event.record(stream)
print("event pending:", not event.query())
```

**Expected behavior**

The launch call returns in microseconds and the event reads as pending, matching numba-cuda.

**Observed behavior**

The launch call blocks for the kernel's full duration (0.43 s here) and the event is already complete when recorded.

**Root cause**

To surface device-side exceptions, `check_kernel_error_code` runs after every successful launch (`cext/launcher/kernel.cpp`) and calls `cuCtxSynchronize` plus a blocking `cuMemcpyDtoH` to read `__numba_cuda_mlir_error_code`. The global is defined in every compilation unit (`mlir_lowering._ensure_error_global`), so the check never takes its symbol-not-found early exit; the synchronization happens on every launch of every kernel. The documentation (`docs/source/user/cudapysupported.rst`) says `raise` and `assert` only have an effect when `debug=True`, and numba-cuda performs the readback only for debug kernels (`numba.cuda.dispatcher._Kernel.launch`), but `tests/test_kernel_exceptions.py` asserts that raises surface with `debug=False`, so the launcher pays the synchronization on every launch to honor that.

ccam80:
Fixes #232.

Every launch read the device error-code global back to surface device-side exceptions, and the readback synchronizes the context, so every launch blocked until the kernel finished. The readback now runs only for kernels compiled with `debug=True`: the compile callback passes `check_error_code` derived from the debug option, and the launcher skips the check otherwise, so launches of non-debug kernels are asynchronous. The compile-result parser accepts 3- to 5-tuples, so older callback shapes keep working.

This is the contract the documentation states (`raise` and `assert` only have an effect when `debug=True`, `docs/source/user/cudapysupported.rst`) and matches numba-cuda, which reads the error code back only for debug kernels. This modifies the behaviour that `test_kernel_exceptions.py` asserted, so those tests have been updated to pass `debug=True` where a raise is expected.

**Tests** (`tests/test_async_launch.py`, new)

- A default kernel's launch is asynchronous: the kernel spins until the host writes a release flag, and an event recorded immediately after the launch call reads as pending, then completes after the release. A count assertion proves the kernel exited on the flag, not its cap.
- A `debug=True` kernel raises `IndexError` at the launch call; the same kernel without `debug` does not raise.

**Verification:** (sm_89, Windows) the launch-asynchrony and no-raise-by-default tests fail on main; the full tests/ suite passes.

### DIFF
--- a/cext/launcher/kernel.cpp
+++ b/cext/launcher/kernel.cpp
@@ -300,10 +300,8 @@ struct CudaKernelHandle {
     CudaKernel cukernel;
     PyPtr post_load_callback;
     bool cooperative = false;
-    // When false, the compiler proved the kernel contains no use of the
-    // error-code global, so the post-launch device error check (which
-    // synchronizes the context) is skipped and launches stay
-    // asynchronous.
+    // False when the compiler proved the kernel never sets the error-code
+    // global; the post-launch check (which synchronizes) is then skipped.
     bool check_error_code = true;
 };
 
@@ -2100,8 +2098,7 @@ Status launch(KernelDispatcher& dispatcher, Grid grid, Grid block, std::optional
     }
 
     // Check for kernel error codes (set by device-side assertion
-    // replacements). Skipped when the compiler proved the kernel never
-    // sets the error code, keeping the launch asynchronous.
+    // replacements), unless the compiler proved the kernel never sets one
     if (kernel_iter->second.check_error_code
             && !check_kernel_error_code(kernel_iter->second.cukernel.lib))
         return ErrorRaised;
--- a/cext/launcher/kernel.cpp
+++ b/cext/launcher/kernel.cpp
@@ -300,6 +300,11 @@ struct CudaKernelHandle {
     CudaKernel cukernel;
     PyPtr post_load_callback;
     bool cooperative = false;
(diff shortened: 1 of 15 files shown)


atmnp (dismissed):
LGTM! Thanks for catching this

leofang:
@ccam80 thank you for your contribution. I see PR #235 from Rob on our team has been merged. Let's close this PR!

PR #234, opened 2026-07-15
nvmath-python CI

Jlisowskyy:


atmnp (approved):

ci/tools/run-nvmath-python-mlir-tests
```diff
@@ -0,0 +1,160 @@
...
+
+# --- Install nvmath test env ---
+
+python -m pip install --upgrade pip
+
+
+NVMATH_REF="${NVMATH_REF:-main}"
+
+echo "Cloning nvmath-python (${NVMATH_REF})"
+git clone --filter=blob:none "${NVMATH_REPO}" "${NVMATH_CHECKOUT}"
```
ZzEeKkAa, line 38: Why do we need to build nvmath here? We can just install one from pypi.

ci/tools/run-nvmath-python-mlir-tests
```diff
@@ -0,0 +1,160 @@
...
+# --- Run nvmath tests ---
+
+cd "${NVMATH_CHECKOUT}"
+export PYTHONSAFEPATH=1
+
+BATCH_SIZE=500
+
+echo "Running nvmath-python numba-cuda-mlir example tests"
+python -m pytest ./tests/example_tests/device_tests/test_device_samples.py \
+  -k "numba_cuda_mlir" \
```
ZzEeKkAa: I think we should skip perf examples/tests

ci/tools/run-nvmath-python-mlir-tests
Jlisowskyy: That was my initial approach, but then I remembered we may have gh/pypi drift, as we did in this release. To avoid CI breaking unexpectedly, I thought having a single source of truth could be a good solution here

ci/tools/run-nvmath-python-mlir-tests
Jlisowskyy: Sure, good catch

PR #235, opened 2026-07-16
Only read back kernel error code when debug=True

### LINKED ISSUE #232: All kernels launch synchronously

rparolin:
## Summary

Every kernel launch was **synchronous**, eliminating host/device overlap and multi-stream concurrency (issue #232).

Root cause: `check_kernel_error_code` runs after every successful launch (`cext/launcher/kernel.cpp`) and always performs a blocking `cuCtxSynchronize` + 4-byte `cuMemcpyDtoH` to read the device-side `__numba_cuda_mlir_error_code` global. Because `_ensure_error_global` defines that global in **every** compilation unit, the check never takes its symbol-not-found early exit, so the synchronization happens on every launch of every kernel.

## Fix

Gate the error readback on `debug`, mirroring numba-cuda (`_Kernel.launch` reads the error code only for debug kernels) and the documented contract that `raise`/`assert` only take effect with `debug=True`:

- **`cext/launcher/kernel.cpp`** — `check_kernel_error_code` (the sole `cuCtxSynchronize` + status D2H in the launch path) now runs only when `dispatcher.debug`. The `debug` flag is parsed in `KernelDispatcher_init` (optional 4th arg, backward-compatible).
- **`src/numba_cuda_mlir/descriptor.py`** — passes `debug=` through `_new_kernel_dispatcher` (the sole dispatcher construction site).
- The eager scalar-record (`record_copies`) writeback previously relied on that unconditional sync for kernel-completion ordering; a local `cuCtxSynchronize` preserves it when the debug check didn't already run (`ctx_guard` holds the launch context through this path, so it targets the correct context).

## Behavior change

With `debug=False`, kernel `raise`/`assert`/bounds-checks no longer surface at launch (they surface with `debug=True`). This matches numba-cuda, the docs, and both linked issues. Genuine async launch errors now surface at the next synchronization point — correct CUDA async semantics.

## Verification (on RTX 5880 Ada, CUDA 13.x, numba-cuda-mlir 0.4.1 + this change)

Issue #232 reproducer:

| | launch call | event after record |
|---|---|---|
| Before | **0.4383 s** (synchronous) | `pending: False` (already complete) |
| After | **0.0001 s** (async) | `pending: True` |

Exception gating preserved:
- `debug=True` out-of-bounds → `IndexError: Index out of bounds in kernel` ✅
- `debug=True` `raise` → `RuntimeError: Runtime error in kernel` ✅
- `debug=False` out-of-bounds → no raise (async) ✅

## Tests

`tests/test_kernel_exceptions.py` updated to the debug-gated contract (raising cases use `debug=True`; a new case asserts no-raise for `debug=False`). The canonical `tests/numba_cuda_tests/cudapy/test_exception.py` was already written debug-gated and passes unchanged. Audited repo-wide: no other test expects a device-side kernel exception from a `debug=False` launch, and `descriptor.py` is the sole `_cext.KernelDispatcher` construction site.

Fixes #232

---

### PLC Local Security Evidence (Advisory)

Branch: `fix/async-kernel-launch-232` · HEAD: `6fc58dd`

Local advisory checks only. Authoritative release gates remain Pulse, SonarQube, Coverity, BlackDuck, nSpect, ScanSpect, OSRB, and Anchore — none of the rows below are release-gate results.

| Check | Status | Tool | Details | Artifact |
|-------|--------|------|---------|----------|
| Secrets | SKIP | Pulse Secret Scanner | Pulse registry unreachable (not `docker login`'d locally; advisory) | `.plc/security/pulse-secret-scan.json` |
| SAST | PASS | Semgrep (bundled rules) | 0 high, 0 medium, 0 low on changed files | `.plc/security/semgrep.json` |
| Dependencies | WARN | pip-audit / osv-scanner | Source changed but no manifest/lockfile in scope — no dependencies modified in this change | `.plc/security/` |
| License | N/A | — | No dependencies added | — |
| Container | N/A | — | No container in this change | — |
| Security Review | PASS | inline | 0 high findings — launcher sync-gating + C++ arg parse; no auth/crypto/input-validation/secret surface, no new allocations | inline in PR diff |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cext/launcher/kernel.cpp
+++ b/cext/launcher/kernel.cpp
@@ -1462,6 +1462,11 @@ struct KernelDispatcher {
     std::vector<bool> constant_arg_flags;
     ProfileMap arg_profiles;
     FamilyMap kernel_families;
+    // When false, launches stay asynchronous: the post-launch device-side error
+    // readback (cuCtxSynchronize + cuMemcpyDtoH of __numba_cuda_mlir_error_code)
+    // is skipped. Mirrors numba-cuda, which only surfaces kernel exceptions
+    // (raise/assert/bounds checks) when the kernel is compiled with debug=True.
+    bool debug = false;
 };
 
 void get_pyarg_types(PyObject* const* pyargs, Py_ssize_t num_pyargs,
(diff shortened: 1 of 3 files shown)


atmnp:
@ccam80 Actually, this is a bit cleaner than #233, I'm going to opt for this one.

atmnp (approved):

cext/launcher/kernel.cpp
```diff
@@ -2090,16 +2095,36 @@ Status launch(KernelDispatcher& dispatcher, Grid grid, Grid block, std::optional
...
+            if (sync_res != CUDA_SUCCESS)
+                return raise(PyExc_RuntimeError, "Failed to synchronize CUDA context: %s",
+                             get_cuda_error(sync_res));
+        }
+        for (const auto& info : helper->record_copies) {
+            CUresult copy_res = g_cuMemcpyDtoH(info.host_ptr, info.device_ptr, info.size);
+            if (copy_res != CUDA_SUCCESS) {
+                return raise(PyExc_RuntimeError, "Failed to copy record back from device: %s",
+                             get_cuda_error(copy_res));
+            }
```
leofang, line 2127: Q: @rparolin Can we do async copy on a stream + stream sync, instead of a device sync here?

PR #236, opened 2026-07-20
[CI] Fix cudDF by pinning pylibcudf to 26.08.*

atmnp:


leofang (approved):
Not entirely sure how this is supposed to work, but if this fixes the issue let's merge for now

atmnp:
It's not a long term solution, we wait for @brandon-b-miller to rebase his branch and then we can drop this pin. For now, since his branch is likely working with this pinned release, we maintain it pinned for now.

Issue #60, opened 2026-05-20
Support extension-prepared StructModel launch arguments


tpn:
## Problem

Argument handler extensions can transform launch values through `prepare_args()`, but extension-defined frontend types need to remain visible to type inference and lowering even when the runtime ABI values are flattened. This matters for custom Python wrapper objects that should type as a Numba extension type in the kernel while launching as a tuple of scalar fields.

Without this, a kernel that accepts an extension struct argument can fail because the launch-time value transformation changes the value shape without preserving the frontend type used by the compiled signature.

Depends on #59, because the same active-launch marshalling path should carry temporary launch metadata and original launch arguments without leaking state between launches.

## Example

```python
import numpy as np
from numba_cuda_mlir import cuda, types
from numba_cuda_mlir.numba_cuda.extending import typeof_impl, make_attribute_wrapper
from numba_cuda_mlir.numba_cuda.extending import models as cuda_models
from numba_cuda_mlir.numba_cuda.extending import register_model as register_cuda_model
from numba_cuda_mlir.models import StructModel as MLIRStructModel
from numba_cuda_mlir.models import register_model

class Pair:
    def __init__(self, x, y):
        self.x = x
        self.y = y

    def prepare_args(self, ty, val, stream=None, retr=None):
        if isinstance(val, Pair):
            # Runtime ABI should receive field values, but the kernel should
            # still compile as if the argument has PairType.
            return ty, (val.x, val.y)
        return ty, val

class PairType(types.Type):
    def __init__(self):
        super().__init__(name="PairType")

pair_type = PairType()

@typeof_impl.register(Pair)
def typeof_pair(val, c):
    return pair_type

@register_cuda_model(PairType)
class PairCudaModel(cuda_models.StructModel):
    def __init__(self, dmm, fe_type):
        super().__init__(dmm, fe_type, [("x", types.int32), ("y", types.int32)])

@register_model(PairType)
class PairMLIRModel(MLIRStructModel):
    def __init__(self, dmm, fe_type):
        super().__init__(dmm, fe_type, [("x", types.int32), ("y", types.int32)])

make_attribute_wrapper(PairType, "x", "x")
make_attribute_wrapper(PairType, "y", "y")

@cuda.jit
def kernel(out, pair):
    out[0] = pair.x + 10 * pair.y

out = np.zeros(1, dtype=np.int32)
kernel[1, 1](out, Pair(np.int32(4), np.int32(5)))
# Expected: out[0] == 54
[...]

PR #239, opened 2026-07-23
Retry kernel compilation with scalar literals

tpn:
## Summary

- Retry `ForceLiteralArg` requests by recompiling with exact Numba literal
  types for requested top-level Python `int` and `bool` arguments.
- Allow planners to discover literal requirements in stages, retrying only when
  each request adds at least one new argument position.
- Keep top-level launch values separate from flattened native ABI arguments so
  extension-facing positions continue to refer to the Python kernel signature.
- Compose literal specialization with planner-requested launch metadata and
  dynamic shared-memory requirements while keeping launch and literal progress
  accounting independent.
- Preserve exact overload reuse, serialization, `recompile()`, native
  dispatcher lifetime, and distinct cache identities for `True` / `1` and
  `False` / `0`.
- Reject literal requests for flattened extension arguments until #60 provides
  an unambiguous top-level-to-ABI mapping.

## Why this is needed

Whole-function extensions sometimes need a scalar value, not merely its runtime
type, to finish a compile-time plan. CUDA.coop group hierarchy is a concrete
case: `group_by(count, exhaustive=...)` needs both values while resolving the
static subgroup and selecting its provider, and the planner also needs the
configured block dimensions.

Before this change, `ForceLiteralArg` could reach `MLIRDispatcher`, but the
dispatcher had no path that rebuilt its native dispatcher and retried using the
actual top-level launch value as a literal. An extension therefore had to reject
a natural Python API, encode the choice in an artificial wrapper type, or leave
provider selection as a runtime branch.

Literal requirements may also emerge in stages. The group planner first asks
for `count`; only after that value is literal can it continue far enough to ask
for `exhaustive`. A fixed one-retry budget rejects that valid sequence.

This PR retries while the recorded literal-position set grows. Progress is
monotonic and bounded by the kernel's top-level parameters; a request that adds
no position is diagnosed instead of looping. Provider-oriented extensions can
therefore select LTO IR that is subsequently inlined while the user-facing
kernel arguments remain ordinary Python scalars.

## Concrete CUDA.coop use

With the small consumer-side hook shown below, CUDA.coop group operations can
accept both mapping choices as top-level kernel arguments:

```python
import numpy as np
import cuda.coop.numba_mlir as coop


@cuda.jit
def segmented_warp_sum(
    values,
    sums,
    lanes_per_group,
    exhaustive,
):
    tid = cuda.threadIdx.x
    group = coop.this_warp().group_by(
        lanes_per_group,
        exhaustive=exhaustive,
    )
    total = coop.reduce(group, values[tid])

    if group.rank() == 0:
        sums[tid // lanes_per_group] = total


values = np.arange(1, 129, dtype=np.int32)
sums8 = np.empty(16, dtype=np.int32)
sums16 = np.empty(8, dtype=np.int32)

segmented_warp_sum[1, 128](values, sums8, 8, True)
segmented_warp_sum[1, 128](values, sums16, 16, True)

np.testing.assert_array_equal(sums8, values.reshape(-1, 8).sum(axis=1))
np.testing.assert_array_equal(sums16, values.reshape(-1, 16).sum(axis=1))
```

`this_warp()`, `group_by()`, `reduce()`, and `rank()` are CUDA.coop
compile-time markers. The downstream group planner connects a top-level
argument to literal retry with this targeted change to its constant resolver:

```python
from numba_cuda_mlir import types
from numba_cuda_mlir.errors import ForceLiteralArg
from numba_cuda_mlir.numbair_transforms import ir


def _constant(self, value):
    definition = self._definition(value)
    if isinstance(definition, ir.Arg):
        position = definition.index
        argtype = self.state.args[position]
        if not isinstance(argtype, types.Literal):
            raise ForceLiteralArg({position})
        return argtype.literal_value

    # Existing handling for globals, free variables, constants, and groups.
    ...
```

On the first launch with `lanes_per_group=8` and `exhaustive=True`:

1. Generic typing promotes the exact block `(128, 1, 1)`, then the group
   planner requests top-level argument 2 (`lanes_per_group`).
2. The next attempt has the launch facts and `Literal[int](8)`. The planner
   consumes the width, advances to the keyword, and requests argument 3
   (`exhaustive`).
3. The final attempt has both `Literal[int](8)` and `Literal[bool](True)`, so
   the planner resolves the subgroup, selects its provider and scratch
   requirement, and erases the markers.

A launch with width 16 receives a separate exact overload. Launch-generation
retry accounting remains independent throughout the staged literal sequence.
Provider choice stays static, so its LTO IR can inline into the kernel instead
of leaving a runtime selection branch.

The resolver change is a CUDA.coop consumer follow-up. This PR supplies the
dispatcher retry and exact-overload machinery that it needs.

## Specialization contract

- A `ForceLiteralArg` request names zero-based positions in the top-level
  Python kernel signature and may request multiple positions at once or in
  stages.
- Each retry must add at least one previously unseen position. A no-progress
  request is diagnosed, and the finite argument list bounds the sequence.
- Supported runtime values are Python `int` and `bool`.
- Learned literal positions and launch sensitivity survive serialization and
  `recompile()`.
- Numerically equal Python booleans and integers retain distinct literal types
  and native constant-cache entries.
- A literal request may override a partial parameter annotation during lazy
  compilation; non-literal annotation precedence is unchanged.
- Planner-requested launch metadata remains active for each literal-qualified
  attempt, and launch-generation conflicts use independent retry accounting.
- The final exact compile result retains planner-required dynamic shared memory.
- Arguments that flatten into multiple native ABI fields are rejected until
  #60 provides an unambiguous top-level-to-ABI mapping.

## Stack

The prerequisite planner, launch-metadata, and dynamic-shared-memory changes
from #202, #203, and #238 are merged. This branch is based on current `main` at
`a9c1711a849128c7bf44090cee2fc18a73c9d83c` and contains only:

- `c42cdb5` Retry kernel compilation with scalar literals
- `7a06720` Retain generic dispatcher during literal rebuild
- `b3ca615` Fix scalar literal retry edge cases
- `65fd43a` Allow staged scalar literal retries

Exact range: `a9c1711..65fd43a`.

Related: #60.

Co-authored-by: Codex 5.6 Sol Ultra Fast

### DIFF
--- a/cext/launcher/kernel.cpp
+++ b/cext/launcher/kernel.cpp
@@ -238,6 +238,7 @@ Status check_kernel_error_code(const CudaLibrary& lib) {
 
 enum class ConstantArgType {
     INT64,
+    UINT64,
     BOOL,
     FLOAT64,
     STRING,
@@ -247,6 +248,7 @@ struct ConstantArg {
     ConstantArgType type;
     union {
         int64_t i64;
+        uint64_t u64;
         double f64;
     } value;
     std::string str;
@@ -261,6 +263,13 @@ struct ConstantArg {
         } else {
             value.i64 = static_cast<int64_t>(val);
         }
+    }
+
+    ConstantArg(uint64_t val) :
+        type(ConstantArgType::UINT64),
+        value{}
+    {
+        value.u64 = val;
     }
 
     ConstantArg(bool val) :
@@ -282,6 +291,8 @@ struct ConstantArg {
         case ConstantArgType::INT64:
         case ConstantArgType::BOOL:
             return value.i64 == other.value.i64;
+        case ConstantArgType::UINT64:
+            return value.u64 == other.value.u64;
         case ConstantArgType::FLOAT64:
             // compare bits of two floats directly
             uint64_t this_float_bits, other_float_bits;
@@ -645,16 +656,37 @@ ScalarExtractor get_scalar_extractor(PythonArgKind kind) {
     }
 }
 
+std::vector<bool> effective_constant_arg_flags(
+        const std::vector<bool>& constant_arg_flags,
+        const std::vector<bool>& literal_arg_flags,
+        const std::vector<PyTypeObject*>& arg_types) {
+    assert(constant_arg_flags.size() == literal_arg_flags.size());
+
+    // Tuple/list arguments are flattened before the native launch, while the
+    // flags describe top-level Python parameters. Literal retry rejects that
+    // ABI today, but every profile still needs one safe flag per native arg.
+    std::vector<bool> effective_flags(arg_types.size(), false);
+    const size_t represented_arg_count = std::min(arg_types.size(), constant_arg_flags.size());
+    for (size_t i = 0; i < represented_arg_count; ++i) {
+        // Numba literal typing accepts exact built-in int/bool values, not
+        // int subclasses, IntEnum members, or NumPy integer scalars.
+        bool literal_eligible = arg_types[i] == &PyLong_Type || arg_types[i] == &PyBool_Type;
+        effective_flags[i] = constant_arg_flags[i] || (literal_arg_flags[i] && literal_eligible);
+    }
+    return effective_flags;
+}
+
 struct PythonArgProfile {
     RefPtr<KernelFamily> family;
     std::vector<PythonArgKind> arg_kinds;
+    std::vector<bool> constant_arg_flags;
 
     // Populated on first use: per-arg fast extractors for all-scalar, no-constant profiles.
     // Empty if the profile contains arrays, TMA descriptors, or unsupported scalar types.
     std::vector<ScalarExtractor> fast_extractors;
     bool fast_path_checked = false;
 
-    void maybe_init_fast_path(const std::vector<bool>& constant_flags) {
+    void maybe_init_fast_path() {
         if (fast_path_checked) return;
         fast_path_checked = true;
 
@@ -662,7 +694,7 @@ struct PythonArgProfile {
         std::vector<ScalarExtractor> extractors;
         extractors.reserve(arg_kinds.size());
         for (size_t i = 0; i < arg_kinds.size(); ++i) {
-            if (i < constant_flags.size() && constant_flags[i])
+            if (constant_arg_flags[i])
                 return;  // constant args need the full path
             ScalarExtractor ex = get_scalar_extractor(arg_kinds[i]);
             if (!ex) return;  // unsupported type (array, complex, etc.)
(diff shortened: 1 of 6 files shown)


cext/launcher/kernel.cpp
```diff
@@ -2475,6 +2488,9 @@ struct hash<ConstantArg> {
         switch (arg.type) {
         case ConstantArgType::INT64:
             return std::hash<int64_t>{}(arg.value.i64);
+        case ConstantArgType::BOOL:
+            return std::hash<int64_t>{}(arg.value.i64)
+                 ^ (static_cast<size_t>(ConstantArgType::BOOL) << 1);
```
tpn, line 2567: @codex this seems a little magic, can you add a small comment clarifying the xor & << 1 logic here as it pertains to BOOL?

tpn:
Following up here on the retry-budget concern we discussed offline, @atmnp: you were right. A fixed `literal_retry_budget = 1` is not composable. Planner A can request one literal, then Planner B can request another only after the first request has been satisfied. The second request used to arrive after the budget was spent.

`_record_literal_arg_positions()` now records each request, and the dispatcher compares the resulting position set with the snapshot from the start of that compile attempt:

```text
generic arguments
  request {2}

argument 2 literalized
  request {3}

arguments 2 and 3 literalized
  compile succeeds
```

If the set grows, compilation retries. If it does not grow, the dispatcher reports a repeated/no-progress literal request. Since the set is monotonic and bounded by the top-level argument count, there is no separate literal retry budget. Launch-generation retry accounting remains independent.

atmnp:
Definitely looks a little bit cleaner, I have another question though. Would there be a case where we call `kernel(7)` , where `7` needs to go through the retried compilation, and then we'd use the same function with a float argument let's say, `kernel(1.5)`. In this case, it looks like the pre-launch would reject this case, so let's either document this behavior that there is a permanent narrowing in valid signatures after a planner is run, or, fix cases like this.

tpn:
Thanks, this was a good catch. I went back through upstream Numba's `ForceLiteralArg` behavior and the actual `cuda.coop` consumer before changing it.

The retry is needed for a narrower case than the PR text originally suggested. `cuda.coop` can already resolve a fixed `group_by()` width from a global or free variable. #239 is needed when one reusable Numba dispatcher takes the width as a kernel argument:

```python
@cuda.jit
def mapped_reduce(values, lane_group_totals, lanes_per_group):
    tid = cuda.threadIdx.x
    lanes = coop.this_warp().group_by(lanes_per_group)
    total = coop.reduce(lanes, values[tid])

    if lanes.rank() == 0:
        lane_group_totals[tid // lanes_per_group] = total


mapped_reduce[1, 32](values, totals_for_8, 8)
mapped_reduce[1, 32](values, totals_for_16, 16)
```

Each integer selects a distinct static `cuda::experimental::group_by<N, ...>` type and group-specific LTO-IR provider. CuTe expresses the same distinction with `Constexpr`; `ForceLiteralArg` is Numba's existing mechanism for requesting that specialization.

Your example exposed a bug: the dispatcher permanently narrowed a position once a planner requested a literal. Recorded positions now act as conditional routing hints. Exact built-in Python `int` and `bool` values take the literal/value-keyed path. Other runtime types keep their normally inferred type and can use a generic overload. If the compilation for that type explicitly requests an unsupported literal, the dispatcher still reports the error.

For a planner that requires integer selectors to be literal, the resulting behavior is:

```python
kernel[1, 1](out, 7)    # Literal[int](7)
kernel[1, 1](out, 1.5)  # generic float64 overload
kernel[1, 1](out, 2.5)  # reuses the same float64 native kernel
kernel[1, 1](out, 9)    # Literal[int](9)
```

For the actual `cuda.coop` `group_by(lanes_per_group)` consumer, passing `1.5` still fails because the group count must be a static integer.

The native launcher applies learned flags per exact Python type profile, so the float overload keeps the scalar fast path and does not compile one kernel per float value. While tracing that path, I also fixed two cache correctness issues: flattened tuple/list profiles now size their flag vector safely, and signed and unsigned literals have distinct native keys, so `-1` cannot alias `2**64 - 1`.

The rebuilt extension passes the dispatcher unit suite, the planner/literal/tuple suite on both local GPUs, and the `cuda.coop` `group_by(8)`/`group_by(16)` consumer tests on both GPUs.

atmnp (approved):
This looks closer to what I'd expect at this stage, thanks @tpn! I still think these changes could be cleaned up a bit by changing the design of the descriptor/launcher but we can get to that later once we get some stability.

PR #242, opened 2026-07-24
Fix out-of-bounds read in extract_cuda_args for flattened tuple args

lahwaacz:
Tuple/list arguments are flattened by `_flatten_arg` in `descriptor.py` before reaching the C++ launcher, so the number of flat args can exceed the number of kernel parameters. `constant_arg_flags` is sized to the original parameter count, so indexing it with the flat arg index goes out of bounds on `std::vector<bool>`.

The fast path (`PythonArgProfile::maybe_init_fast_path`) already guards with a bounds check. Add the same guard to `extract_cuda_args`.

Without this fix, any kernel launch that passes a tuple argument alongside a device array aborts when the extension is built with `_GLIBCXX_ASSERTIONS` (as Linux distribution packages typically are). The pip wheel is unaffected because it is built without assertions, so the out-of-bounds read silently returns garbage.

### DIFF
--- a/cext/launcher/kernel.cpp
+++ b/cext/launcher/kernel.cpp
@@ -1415,7 +1415,13 @@ Status extract_cuda_args(PyObject* const* pyargs, size_t num_pyargs,
     helper.cuda_context = nullptr;
     for (size_t i = 0; i < num_pyargs; ++i) {
         PyObject* pyobj = pyargs[i];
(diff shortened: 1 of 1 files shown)


atmnp:
@coderabbitai review

atmnp (approved):

PR #243, opened 2026-07-29
Avoid locking prepared planner launches

tpn:
## Why

PR #203 prepares planner-aware kernels before launch so a planner can request a launch-config specialization. Its first implementation put all of `_prepare_for_launch()` under Numba's global compiler lock.

`_ArgMarshaller._launch()` calls that method whenever a planner is registered. As a result, a hot launch took the global compiler lock even when its signature was already compiled and no new launch-config specialization was needed. Unrelated launches could serialize on a compiler lock without doing any compilation.

## Change

Look up the matching compile result before acquiring the lock. A prepared signature now continues directly to launch.

On a miss, acquire the compiler lock, refresh the active launch config, and repeat the lookup before compiling. The second lookup matters because another thread may publish the signature while this launch waits for the lock; the planner may also mark the signature as launch-sensitive during that interval.

The existing launcher and generation repair remains unchanged.

This is a follow-up to #203 and addresses [the review nit](https://github.com/NVIDIA/numba-cuda-mlir/pull/203#discussion_r3670117876).

## Validation

- `pre-commit run --files src/numba_cuda_mlir/descriptor.py tests/test_descriptor_launch_config.py`
- `pytest -q -o addopts= -p no:cacheprovider tests/test_descriptor_launch_config.py tests/test_post_inline_planners.py`
  - 80 passed on an NVIDIA RTX PRO 6000 Blackwell (SM 120)
  - 80 passed on a Quadro RTX 4000 (SM 75)

### DIFF
--- a/src/numba_cuda_mlir/descriptor.py
+++ b/src/numba_cuda_mlir/descriptor.py
@@ -434,8 +434,7 @@ class _ArgMarshaller:
         self._launch_config_generation = launch_config_generation
         self._stream_ref = stream_ref
         self._launch_stream = launch_stream
-        self._kernel_dispatcher_registered = False
-        self._kernel_dispatcher_registration_lock = threading.Lock()
+        self._launcher_state_lock = threading.Lock()
         self._sig_cache = {}  # {type_key: (argtypes, fast_ok)}
         self._array_sig_cache = {}  # {type_key: [(argtypes, ((idx, array_key), ...))]}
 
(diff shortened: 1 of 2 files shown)


atmnp:
Maybe a cleaner design:
  1. At _ArgMarshaller._launch(), perform a conservative readiness check.
  2. If matching overload, native dispatcher, and generation are current, use existing launcher directly.
  3. Only call _prepare_for_launch() for:
      - unseen signature/configuration;
      - missing launch-specialized overload;
      - stale dispatcher generation or registration.

  4. Keep _prepare_for_launch() fully under global_compiler_lock.

  A small dispatcher helper could return either the ready (dispatcher, generation, config) tuple or None.
  Caller tries it first; locked preparation tries it again before compiling. One readiness definition, no
  unlocked compiler-state iteration, no scattered state decisions.

atmnp:
This still seems overly complex for the case we're trying to tackle, it's extremely defensive. Do we need to be this defensive?

tpn:
@atmnp pushed a simpler version, is that what you had in mind?

atmnp (approved):
Yes perfect!

PR #244, opened 2026-07-29
Add support for Python 3.10

atmnp:


src/numba_cuda_mlir/lowering/math.py
```diff
@@ -673,12 +673,14 @@ def math_isnan_cg(mlir_lower, target, args, kwargs):
     mlir_lower.store_var(target, result)
 
 
-@lower(math.exp2, types.Number)
-def math_exp2_cg(mlir_lower, target, args, kwargs):
-    assert not kwargs, "math_exp2 does not accept any keyword arguments"
-    value = _ensure_float(mlir_lower.load_var(args[0]))
-    result = math_dialect.exp2(value)
-    mlir_lower.store_var(target, result)
+if hasattr(math, "exp2"):
```
VijayKandiah, line 677: Might make sense to add a comment at places like this where you're gating for py3.10.

src/numba_cuda_mlir/typing/math.py
```diff
@@ -90,6 +89,9 @@ def _float_to_integer_return_type(arg_type):
 ]:
     _make_unary_math_template(func)
 
+if hasattr(math, "exp2"):
```
VijayKandiah, line 93: > Might make sense to add a comment at places like this where you're gating for py3.10. 

+1

VijayKandiah (approved):
LGTM after you add py3.10 specific gating comments

PR #245, opened 2026-07-29
Introduce support for WoA

atmnp:
Notes:
- On Windows ARM64, use the complete NVPTX data-layout string required by the WoA LLVM/MLIR toolchain. It explicitly defines 64-bit default pointers, 32-bit pointers in address space 6, scalar and vector alignments, native integer widths, and aggregate alignment. The target remains nvptx64-nvidia-cuda; this does not apply the host ARM ABI to device code. Other platforms retain the existing compact NVPTX layout.
- The LLVM 7 Bridge is not supported for this platform at this time.


isVoid (approved):

tests/test_post_inline_planners.py
```diff
@@ -32,6 +33,10 @@
 
 
 _RECOMPILE_VALUE = 1
+IS_WINDOWS_ARM64 = platform.system() == "Windows" and platform.machine() in (
+    "ARM64",
+    "AMD64",
+)
```
isVoid: Duh, "AMD64" is not WoA here 🤣

PR #246, opened 2026-07-30
[Lowering] Prioritize unsigned lowering is one operand is unsigned

### LINKED ISSUE #121: Integer comparisons ignore operand signedness (signed compared as unsigned, and vice versa)

atmnp:
This should resolve the missing fix required to close #121.

### DIFF
--- a/src/numba_cuda_mlir/lowering/math.py
+++ b/src/numba_cuda_mlir/lowering/math.py
@@ -5,7 +5,6 @@ import functools
 from typing import Callable
 from numba_cuda_mlir import lowering_utilities
 from numba_cuda_mlir.lowering_utilities import (
-    coerce_numpy_scalars_for_binary_op,
     numpy_implicit_type_promotion,
     convert,
     get_or_insert_function,
@@ -312,6 +311,16 @@ def pow_cg(builder, target, args, kwargs):
     builder.store_var(target, res)
 
 
+def _convert_integer_operand(value, source_type, target_type):
+    if isinstance(target_type, ir.FloatType):
+        return (
+            arith.sitofp(out=target_type, in_=value)
+            if source_type.signed
+            else arith.uitofp(out=target_type, in_=value)
+        )
+    return lowering_utilities.convert(value, target_type, signed=source_type.signed)
+
+
 def _bin_op_cg(op, builder, target, args, kwargs):
     assert not kwargs, "add_cg does not accept any keyword arguments"
     assert len(args) == 2, "add_cg expects 2 arguments"
@@ -335,6 +344,13 @@ def _bin_op_cg(op, builder, target, args, kwargs):
 
     is_unsigned = _is_unsigned_operand(lhs_type) and _is_unsigned_operand(rhs_type)
 
+    promoted_numba_type = None
+    unified_type = None
+    if isinstance(lhs_type, types.Integer) and isinstance(rhs_type, types.Integer):
+        promoted_numba_type = lhs_type.unify(None, rhs_type)
+        assert promoted_numba_type is not None
+        unified_type = builder.get_mlir_type(promoted_numba_type)
+
     lhs, rhs = builder.load_var(lhs), builder.load_var(rhs)
 
     # Handle cases where load_var returns Python/numpy scalars instead of MLIR values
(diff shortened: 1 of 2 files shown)


acosmicflamingo:
@atmnp thanks for looking into it!

VijayKandiah (commented):
I ran a matrix of 20 comparison cases each with its own dispatcher and an explicit signature with this PR and `main` branch. The `and` → `or` change in this PR fixes five cases that were wrong in `main` but breaks five others, and both `main` and this PR land at 13/20 for correctness in this small set.

Below are the cases that are wrong in either `main` or this PR.

| case | expected | `main` returns | this PR returns |
|---|:--:|:--:|:--:|
| `uint32(2**31) > 0` | True | False | True |
| `uint32(2**32-1) > 0` | True | False | True |
| `uint32(2**31) >= 1` | True | False | True |
| `uint32(2**32-1) > 3000000000` | True | False | True |
| `uint64(2**63) > 0` | True | False | True |
| `uint32(2**31) > 3000000000` | False | False | True |
| `int32(-1) > uint32(5)` | False | False | True |
| `int32(-1) < uint32(5)` | True | True | False |
| `int32(-50) > uint8(0)` | False | False | True |
| `int64(-1) > uint32(5)` | False | False | True |
| `uint32(2**31) > uint64(2**40)` | False | True | True |
| `int8(-1) > uint8(200)` | False | True | True |

The five newly wrong rows are the "signed compared as unsigned" direction from this issue's title: `or` forces an unsigned predicate even when the other operand is signed and negative, though numba has already unified the pair to a signed type (`unify(int32, uint32) == int64`). `test_operator.py` passes on the branch, so CI covers neither direction. We need to update the test to cover this in CI.

I think the predicate is just the symptom, and the real problem is that we promote operands over signless MLIR types, after signedness has already been erased:

1. `coerce_numpy_scalars_for_binary_op` reads signedness from the MLIR type, where `width > 1` (there so `i1` zero-extends) sign-extends every wider integer. Widening `uint32(2**31)` to `i64` yields `0xFFFFFFFF80000000`. The unsigned predicate hides that against `0`, hence `uint32(2**31) > 3000000000` goes wrong in the above table.
2. MLIR-level promotion can only take the wider of the two, so it can't express `uint8 + int8 -> int16` or `uint64 + int64 -> float64`. That's `int8(-1) > uint8(200)`, wrong in both columns. 

Numba's own unification already encodes exactly these rules, and `_bin_op_cg` still has the numba types in scope. Maybe we should try promoting there instead, and letting the operand conversions state their signedness explicitly. I believe leaving this in `math.py` as `and` is the correct predicate rule, since "both operands unsigned → unsigned, else signed" is what numba's unification does.

src/numba_cuda_mlir/lowering/math.py
```diff
@@ -335,6 +344,12 @@ def _is_unsigned_operand(operand_type):
 
     is_unsigned = _is_unsigned_operand(lhs_type) and _is_unsigned_operand(rhs_type)
 
+    promoted_numba_type = None
+    if isinstance(lhs_type, types.Integer) and isinstance(rhs_type, types.Integer):
+        promoted_numba_type = lhs_type.unify(None, rhs_type)
+        assert promoted_numba_type is not None
+        unified_type = builder.get_mlir_type(promoted_numba_type)
```
VijayKandiah, line 352: `unified_type` is only bound here when both operands are integers, and otherwise not until line 373, so pyright flags it as possibly unbound at lines 362, 369, 376, 383 and 384. Could you initialize it alongside `promoted_numba_type`
```suggestion
    promoted_numba_type = None
    unified_type = None
    if isinstance(lhs_type, types.Integer) and isinstance(rhs_type, types.Integer):
        promoted_numba_type = lhs_type.unify(None, rhs_type)
        assert promoted_numba_type is not None
        unified_type = builder.get_mlir_type(promoted_numba_type)
```

VijayKandiah (approved):
LGTM now with minor change.

PR #251, opened 2026-08-04
Lower captured NumPy arrays to CUDA constant memory

### LINKED ISSUE #247: [BUG] A closure-captured array is copied to the device heap per thread, per launch, and never freed — kernels die with CUDA_ERROR_ILLEGAL_ADDRESS after N launches
# Summary

When a `@cuda.jit` kernel closes over a NumPy array, `numba-cuda-mlir` lowers the capture into a
device-side `malloc` that copies the array's data onto the CUDA device heap. The allocation happens
**once per thread, on every launch**, and is never freed. Repeated launches therefore exhaust the
8 MB device heap; `malloc` then returns `NULL`, the returned pointer is dereferenced, and the launch
fails with `CUDA_ERROR_ILLEGAL_ADDRESS`.

`numba-cuda` places the same captured array in **constant memory** (`_cudapy_cmem`), which costs
nothing per launch and does not accumulate.

The failure is silent until it happens, and it is invisible to the usual memory diagnostics: the
device heap is pre-allocated by the driver, so `cudaMemGetInfo` free memory stays **flat to the
byte** while the budget drains.

# Environment

| | |
|---|---|
| `numba-cuda-mlir` | 0.4.2 (`[cu12]` and `[cu13]`) |
| control | `numba-cuda` 0.30.4 + `numba` 0.66.0 |
| Python | 3.13 |
| reproduced on | GTX 1650 Ti (cc 7.5, CUDA 12.9); Tesla T4 (cc 7.5), A10G (cc 8.6), A100-SXM4-40GB (cc 8.0), all **CUDA 13.3** |

Not specific to a GPU, a compute capability, or a CUDA major. The script above was run unchanged on
all four: every one reports the same `malloc` size (576 bytes) and fails at the **same point** — 100
launches, 12 800 cumulative threads — while `numba-cuda` completes 100 000 launches (12.8 M threads)
on every one. Three distinct compute capabilities and both CUDA majors give an identical failure
point, which is what you would expect from a fixed-size heap being consumed at a fixed rate.

# Reproducer

Self-contained; `numba-cuda` is the in-file control.

```python
import sys
import numpy as np

if sys.argv[1] == "mlir":
else:
    from numba import cuda

CAPTURED_LEN = 64
THREADS = 128

table = np.arange(CAPTURED_LEN, dtype=np.int64)   # <-- the only unusual feature


@cuda.jit(cache=False)
def kernel(out):
    i = cuda.grid(1)
    if i < out.shape[0]:
        out[i] = table[i % table.shape[0]]


out = cuda.to_device(np.zeros(THREADS))
for n in range(1, 100_001):
    try:
        kernel[1, THREADS](out)
        if n % 25 == 0:
            cuda.synchronize()
    except Exception as exc:
        print(f"FAILED after {n} launches ({n * THREADS} cumulative threads): {exc}")
        raise SystemExit(1)
cuda.synchronize()
print(f"completed 100000 launches ({100_000 * THREADS} cumulative threads)")
```

## Observed

```
$ python repro.py mlir
FAILED after 100 launches (12800 cumulative threads):
[CUresult.CUDA_ERROR_ILLEGAL_ADDRESS] Call to cuCtxSynchronize results in CUDA_ERROR_ILLEGAL_ADDRESS

$ python repro.py numba
completed 100000 launches (12800000 cumulative threads)
```

`numba-cuda` survives 12.8 M cumulative threads — three orders of magnitude past where
`numba-cuda-mlir` dies.

On Modal the host log recorded the corresponding hardware fault:

```
Xid 31 ... MMU Fault: ENGINE GRAPHICS GPC0 ... faulted @ 0x0_00000000.
Fault is of type FAULT_PDE ACCESS_TYPE_VIRT_WRITE
```

— a write through a null/unmapped pointer, consistent with a failed `malloc` being used unchecked.

# Evidence

**1. The PTX contains the allocation.** `kernel.inspect_asm()` under `numba-cuda-mlir`:

```
.extern .func  (.param .b64 func_retval0) malloc
...
{ // callseq 0, 0
  st.param.b64  [param0+0], 576;
  call.uni (retval0), malloc, (param0);
  ...
}
```

The size is exactly `8 * len(table) + 64`, i.e. the array's **data** is copied, not just a
descriptor:

| `len(table)` | malloc bytes |
|---|---|
| 4 | 96 |
| 16 | 192 |
| 64 | 576 |
| 256 | 2112 |

Under `numba-cuda` the same kernel emits **no** `malloc` and instead contains
`_cudapy_cmem` in constant memory.

**2. The failure point matches the heap arithmetic.** 8 MB ÷ 576 B ≈ 14 563 threads; the reproducer
dies at 12 800 (the first synchronize past the limit, checked every 25 launches).

**3. Raising the heap moves the failure point proportionally** — the signature of heap exhaustion
rather than a bad access. With `cudaLimitMallocHeapSize` set via
`cupy.cuda.runtime.deviceSetLimit(0x02, bytes)`:

| heap | threads × launches | result |
|---|---|---|
| 8 MB (default) | 25 600 | fails |
| 16 MB | 25 600 | passes |
| 16 MB | 51 200 | fails |
| 32 MB | 51 200 | passes |
| 64 MB | 102 400 | passes |

Exactly doubling the heap exactly doubles the budget.

**4. It is process-wide, not per kernel.** Replacing the single kernel above with two distinct
captured-array kernels and alternating the launches fails at launch 100 / 12 800 cumulative threads —
exactly where one kernel alone fails. The budget is shared across compiled kernels.

**5. The heap size is the documented default.** `cudaLimitMallocHeapSize` reads back as 8 388 608
bytes (8 MiB) on the machines above, which is what the arithmetic in (2) and (3) assumes.

# Why this is easy to miss

- `cudaMemGetInfo` free memory is **constant to the byte** right up to the fault, because the device
  heap is pre-allocated. Any leak check based on free VRAM reports "no leak".
- The error surfaces as `CUDA_ERROR_ILLEGAL_ADDRESS` at an arbitrary later launch or synchronize,
  not at the allocation, so it looks like an out-of-bounds access in user code.
- Nothing fails until enough cumulative threads have run, so short tests and small examples pass
  while longer runs of the same code do not. The threshold moves with the launch geometry: a wider
  grid spends the budget faster, so the same program fails after fewer iterations.

# Expected behaviour

A closure-captured array should not consume a per-thread, per-launch device-heap allocation that is
never released. `numba-cuda` compiles the same source without any device-heap traffic, so a kernel
that runs there indefinitely should also run indefinitely here.

# Related

We are filing a second issue separately: **compile time grows much faster than `numba-cuda` when a
kernel composes many non-trivial device functions**. The two are independent defects but interact in
practice, so they are worth reading together.

atmnp:
Materialize host array captures as aligned addrspace-4 LLVM globals and reconstruct their memref descriptors without device-heap allocation.

Reuse the descriptor builder for host and device array captures. Add repeated launch and noncontiguous-capture regressions, and remove stale constant-memory xfail markers.

This closes #247.

### DIFF
--- a/src/numba_cuda_mlir/mlir_lowering.py
+++ b/src/numba_cuda_mlir/mlir_lowering.py
@@ -1260,7 +1260,7 @@ extern "C" __global__ void
                     self.incref(target_type, value_op)
                     self.store_var(target, value_op)
 
-    def lower_array_literal(self, value: np.ndarray) -> ir.Value:
+    def lower_array_literal(self, value: np.ndarray, numba_type) -> ir.Value:
         global constant_array_id
 
         array = np.ascontiguousarray(value)
@@ -1287,17 +1287,26 @@ extern "C" __global__ void
         ptr = llvm.addrspacecast(llvm.PointerType.get(), ptr)
         strides = [stride // array.itemsize for stride in array.strides]
         desc = self._build_memref_descriptor(ptr, array.shape, strides)
-        mr_type = T.memref(*array.shape, element_type=dtype)
+        if numba_type is not None:
+            dynamic_size = ir.ShapedType.get_dynamic_size()
+            mr_type = T.memref(*([dynamic_size] * array.ndim), element_type=dtype)
+        else:
+            mr_type = T.memref(*array.shape, element_type=dtype)
         return builtin.unrealized_conversion_cast([mr_type], [desc])
 
     def lower_literal_if_needed(self, value: ir.Value | np.ndarray, numba_type=None) -> ir.Value:
         match value:
             case types.Type() if isinstance(numba_type, types.DTypeSpec):
                 return self._materialize_type_token(numba_type)
+            case tuple() if isinstance(numba_type, types.BaseTuple):
+                return tuple(
+                    self.lower_literal_if_needed(element, element_type)
+                    for element, element_type in zip(value, numba_type.types)
+                )
             case tuple():
                 return tuple(map(self.lower_literal_if_needed, value))
             case np.ndarray():
-                return self.lower_array_literal(value)
+                return self.lower_array_literal(value, numba_type)
             case np.number() | np.bool_():
                 # np.bool_ is not an np.number but lowers the same way.
                 mlir_type = to_mlir_type(value.dtype)
--- a/src/numba_cuda_mlir/mlir_lowering.py
+++ b/src/numba_cuda_mlir/mlir_lowering.py
@@ -94,6 +94,7 @@ KERNEL_ERROR_CODES = {
 }
 # MLIR LLVM dialect dynamic GEP sentinel, LLVM::GEPOp::kDynamicIndex / INT32_MIN.
 _GEP_DYNAMIC_INDEX = -2147483648
+constant_array_id = 0
 
 
 def get_mlir_module_str(metadata):
@@ -165,6 +166,7 @@ class MLIRLower(object):
         )
         self._seen_mlir_libraries = set()
         self._cloned_device_funcs: set[str] = set()
+        self._constant_array_globals = {}
         self._linked_external_items = set()
         self._linked_external_link_items = []
         self._linked_ltoirs = []
@@ -982,37 +984,14 @@ extern "C" __global__ void
             else:
                 self.store_var(target, free_var.value)
 
-    def lower_captured_array_to_memref(self, target, pyval):
-        """Build an MLIR memref value from __cuda_array_interface__ metadata.
-
-        Bakes the captured device array's pointer, shape, and strides as constants
-        into a memref descriptor struct, then casts to the target memref type.
-        Registers ``pyval`` on the active code library to keep the underlying
-        GPU memory alive while the compiled kernel exists.
-        """
-        self.context.active_code_library.referenced_objects[id(pyval)] = pyval
-        array = pyval.__cuda_array_interface__
-        memref_type = self.get_mlir_type(self.get_numba_type(target.name))
-        shape = array["shape"]
+    def _build_memref_descriptor(self, ptr, shape, strides):
         ndim = len(shape)
-        itemsize = np.dtype(array["typestr"]).itemsize
-
-        byte_strides = array.get("strides")
-        if byte_strides:
-            strides = [s // itemsize for s in byte_strides]
-        else:
-            strides, s = [], 1
-            for d in reversed(shape):
-                strides.insert(0, s)
-                s *= d
-
         if ndim > 0:
             struct_type = ir.Type.parse(
                 f"!llvm.struct<(ptr, ptr, i64, array<{ndim} x i64>, array<{ndim} x i64>)>"
             )
         else:
             struct_type = ir.Type.parse("!llvm.struct<(ptr, ptr, i64)>")
-        ptr = llvm.inttoptr(llvm.PointerType.get(), arith.constant(T.i64(), array["data"][0]))
         i64c = lambda v: arith.constant(T.i64(), v)
         ins = lambda d, v, *p: llvm.insertvalue(
             container=d, value=v, position=ir.DenseI64ArrayAttr.get(list(p))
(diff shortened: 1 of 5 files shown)


src/numba_cuda_mlir/mlir_lowering.py
```diff
@@ -1254,21 +1261,34 @@ def lower_var_assign(self, target, var):
...
+                    addr_space=4,
+                    constant=True,
+                    alignment=array.dtype.alignment,
+                    value=ir.Attribute.parse(f'"{data}"'),
+                )
+        ptr = llvm.mlir_addressof(llvm.PointerType.get(4), name)
+        ptr = llvm.addrspacecast(llvm.PointerType.get(), ptr)
+        strides = [stride // array.itemsize for stride in array.strides]
+        desc = self._build_memref_descriptor(ptr, array.shape, strides)
+        mr_type = T.memref(*array.shape, element_type=dtype)
```
VijayKandiah: This builds the memref type from the payload instead of the variable's declared type, which regresses slicing of captured arrays. Repro:

```python
C2D = np.ascontiguousarray(np.arange(12, dtype=np.int64).reshape(3, 4))

@cuda.jit(cache=False)
def k(out):
    col = C2D[:, 2]          # row slice C2D[1] is fine, column slice is not
    for i in range(col.shape[0]):
        out[i] = col[i]

k[1, 1](cuda.device_array(3, dtype=np.int64))
```
```
InternalCompilerError: Invalid MLIR module: Verification failed:
error: 'memref.collapse_shape' op expected collapsed type to be
'memref<?xi64, strided<[4], offset: ?>>' but found 'memref<?xi64, strided<[?], offset: ?>>'
```
`T.memref(*array.shape, ...)` yields a static identity layout (`memref<3x4xi64>`), so `collapse_shape` infers a static stride `[4]`, while the numba type maps to the fully dynamic `memref<?x?xi64, strided<[?, ?], offset: ?>>`.  Kernel-argument arrays are fine because `lower_captured_array_to_memref` already uses `self.get_mlir_type(self.get_numba_type(target.name))`

`lower_literal_if_needed` has `numba_type` in hand and drops it at line 1280. Threading it through fixes C, F, transposed and strided captures:

```suggestion
        if numba_type is not None:
            mr_type = self.get_mlir_type(numba_type)
        else:
            mr_type = T.memref(*array.shape, element_type=dtype)
```
I think it's worth adding a column-slice case to the new tests — the existing non-contiguous test uses a 1-D slice, where there's no C/F distinction to get wrong.

VijayKandiah (changes requested):

src/numba_cuda_mlir/mlir_lowering.py
```diff
@@ -1254,21 +1261,34 @@ def lower_var_assign(self, target, var):
...
+
+        array = np.ascontiguousarray(value)
+        dtype_numba = to_numba_type(array.dtype)
+        dtype = self.get_storage_type(dtype_numba)
+        name = self._constant_array_globals.get(id(value))
+        if name is None:
+            name = f"constant_array_{constant_array_id}"
+            constant_array_id += 1
+            self._constant_array_globals[id(value)] = name
+            array_type = ir.Type.parse(f"!llvm.array<{array.nbytes} x i8>")
```
VijayKandiah, line 1274: Captures above the 64 KB constant budget now fail to compile, where the old heap path had no such ceiling. Repro:

```python
import numpy as np

table = np.arange(9 * 1024, dtype=np.int64)   # 72 KB > 64 KB

@cuda.jit(cache=False)
def kernel(out):
    i = cuda.grid(1)
    out[i] = table[i % table.shape[0]]

kernel[1, 8](cuda.device_array(8, dtype=np.int64))
```
```
nvJitLinkError: ERROR_PTX_COMPILE (4)
nvJitLink error log: ptxas error : File uses too much global constant data (0x12000 bytes, 0x10000 max)
```

64 KB works, 72 KB doesn't. ptxas does name the cause, so this isn't a mystery to debug — but it is a behavior change worth being deliberate about. Inheriting numba-cuda's 64 KB ceiling seems like the right call; is it worth a release note, or falling back to global memory when `array.nbytes` won't fit?

atmnp: Let's just make it a release note, I don't think *all* of our users would appreciate it this kind of behavior changed out from underneath them when they're not keeping track of the size of their data.

VijayKandiah (approved):
LGTM after CI passes.

raphaelchinchilla:
Thanks everyone for addressing these issues!

PR #252, opened 2026-08-05
ENH: expose LLVM IR inspection and NVVM input dumping

atmnp:
Add architecture-native `inspect_llvm()` support and a `NUMBA_CUDA_MLIR_DUMP_NVVM` debugging hook.

Generate readable LLVM IR lazily for inspection, using the LLVM70 path for pre-Blackwell targets and the modern bridge elsewhere. Keep it separate from the libnvvm preparation path, whose output may be serialized bitcode.

Capture NVVM input at the actual libnvvm boundary for both PTX and LTO, including failed LLVM70 translations. The dump preserves raw bitcode when applicable and only renders textual IR for inspection.

Cache inspected IR per compiled overload and retain launch-specialized inspection keys.

Why this matches the two PR designs:
- #145 (https://github.com/NVIDIA/numba-cuda-mlir/pull/145) established inspect_llvm() as a lazy, public inspection API: it should return readable, target-native LLVM IR without adding LLVM IR generation to the normal compilation state just for later inspection.
- #149 (https://github.com/NVIDIA/numba-cuda-mlir/pull/149) established that dumping belongs at the libnvvm input boundary, covering both PTX and LTO and preserving the actual artifact—even when it is binary bitcode.
- The staged patch keeps those concerns separate: inspect_llvm() explicitly requests text, while the dump captures the bytes used for libnvvm. That avoids decoding BC\xc0\xde bitcode as UTF-8 and makes the LLVM70 translation setup reusable for either result.

### DIFF
--- a/src/numba_cuda_mlir/mlir_optimization.py
+++ b/src/numba_cuda_mlir/mlir_optimization.py
@@ -479,7 +479,12 @@ def get_llvmir(cres, target_options=None) -> str:
         cc = chip.replace("sm_", "")
 
         if _needs_llvm70_path(cc):
-            return _call_llvm70_capi(module, target_options, gen_llvmir=True).decode("utf-8")
+            return _call_llvm70_capi(
+                module,
+                target_options,
+                gen_lto=target_options.get("lto", False),
+                gen_llvmir=True,
+            ).decode("utf-8")
 
         gpu_mod = _get_single_gpu_module(module)
         gpu_mod.operation.attributes["llvm.data_layout"] = ir.StringAttr.get(NVPTX64_DATALAYOUT)
@@ -490,7 +495,11 @@ def get_llvmir(cres, target_options=None) -> str:
 
             ctk_major, ctk_minor = get_cuda_runtime_version()
             llvm_ir = translate_gpu_module_to_libnvvm_ir(
-                _operation_to_text(gpu_mod.operation),
+                _operation_to_text(
+                    gpu_mod.operation,
+                    preserve_debug_info=target_options.get("debug", False)
+                    or target_options.get("lineinfo", False),
+                ),
                 ctk_major,
                 ctk_minor,
                 _get_nvvm_ir_version(),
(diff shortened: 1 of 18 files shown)


src/numba_cuda_mlir/mlir_optimization.py
```diff
@@ -255,18 +272,69 @@ def _call_llvm70_capi(module, target_options, gen_lto=False) -> bytes:
...
+def _maybe_dump_nvvm(nvvm_ir: bytes) -> None:
+    """Dump the NVVM IR fed to libnvvm when configured."""
+    from numba_cuda_mlir.numba_cuda import config
+
+    target = config.CUDA_DUMP_NVVM
+    if not target:
+        return
+
+    is_bitcode = nvvm_ir[:4] == _BITCODE_MAGIC
+    if target.lower() in {"1", "true", "yes", "stderr"}:
```
VijayKandiah, line 319: Only `{"1", "true", "yes", "stderr"}` are recognized, so the negations fall through to the path branch. `CUDA_DUMP_NVVM` is read as a `str` (config.py:469), so `"0"` gets treated as a filename:
```console
$ NUMBA_CUDA_MLIR_DUMP_NVVM=0 python -c "
from numba_cuda_mlir import cuda, types
@cuda.jit(device=True, chip='sm_90')
def foo(v): return v + 1
foo.compile((types.int32,))"
$ ls
0          # 1460 bytes of bitcode
```
Same for `false`, `no` and `off`. Every on/off knob in that config block is `int`-typed, including `CUDA_DISABLE_LTO_OPT` on the line above, so `=0` to disable is what a user would reach for.
```suggestion
    if target.lower() in {"0", "false", "no", "off"}:
        return
    if target.lower() in {"1", "true", "yes", "stderr"}:
```

VijayKandiah (changes requested):

src/numba_cuda_mlir/mlir_optimization.py
atmnp: Fixed!

VijayKandiah (approved):
LGTM

PR #255, opened 2026-08-07
Accepts individual fastmath flags for LLVM flags + ftz

### LINKED ISSUE #220: fastmath has almost no effect on generated code, and selective flag sets are rejected

ccam80:
Supersedes #221, which missed the LTO pathway, ftz flag, and fma uplift.
Fixes #220.

Fastmath flags provided as bool | set | dict | FastMathOptions in a @cuda.jit() decorator are respected by the compiler. This brings the MLIR backend past parity with numba-cuda, which accepts the flags but treats any truthy flag as a blanket fastmath=True.

fastmath=True or {"fast"} applies all flags but nnan/ninf, as numba-cuda and nvcc --use_fast_math do. nnan/ninf can still be applied if named explicitly. Each device function compiles under the flags included in its own decorator.

Fastmath-capable arith/math ops are identified from the generated bindings, and get stamped with #arith.fastmath<...>. fast division and tanh get rewritten manually (as in numba-cuda) to __nv_fast_fdividef and tanh.approx.f32. Constant-zero powers are folded to 1.0 before lowering. flush-to-zero is not applicable at function level, so is applied module-wide through the libnvvm/ptxas flags.

This PR passes the module-level libnvvm/ptxas flags to libnvvm through a new CAPI parameter. The flags were also dropped at the linker boundary, so they are added to the linker config. nvJitLink expects an integer option form (-ftz=1) and cuda.core provides a word form (-ftz=true), so there's a LinkerOptions subclass added to translate bool-int.

Piping `contract` through allows fma uplift to take place in the compiler, so the `math-uplift-to-fma` pass in the base pipeline is no longer required. Removing this early uplift allows ptxas to perform some cse before contracting, and afforded a ~5% cycle count and ~23% SASS instruction count improvement on a large, inline-heavy kernel in our package (CC 8.9, LTO on).

Tests:
- test_fastmath.py: {"arcp"} gives div.approx, {"nsz"} keeps precise division (no blanket truthiness), {"afn"} gives sin.approx without approximate division, invalid flags rejected, attributes appear in MLIR, standalone {"ftz"} compiles and emits .ftz instructions, LTO links succeed, nvJitLink options made numeric, lineinfo/debug flags with fastmath keep debug directives and div.approx.
- test_array_reductions.py: NaN-aware reductions stay correct under fastmath=True.
- test_math.py: x ** (0.0) and x ** (-0.0) return 1.0 for NaN, +/-0.0, +/-inf bases under fastmath=False, {"afn"}, and True.

Verification (CC 8.9, Windows): 24/26 xfailed fastmath tests go green, 2 stay xfailed (fastmath doesn't propagate into device functions from decorated kernel, no ZeroDivisionError); GPU fast-vs-precise results agree to within float32 roundoff on a float32 test kernel.


ccam80:
For human review: I had a look into the behaviour CodeRabbit raised - it looks to me like a false positive. I can't see any instances in the repo or really imagine why a user would want to `consteval(current_target_options()==True)`, and I'd suggest that a truthiness check instead would make more sense. The suggestion of preserving and returning the raw args instead would require all callers to know the form that the argument was provided in, which seems like a lot of structure to add to preserve a not-very-helpful use case. I'll update the fastmath.rst to document the input formats, and not action CodeRabbit's suggestion until this PR goes through a human review.

PR #256, opened 2026-08-07
[Extending] Fix overload registry cache key lookup for literals

atmnp:
The fix is to normalize literals so the lookup works.


ZzEeKkAa:
```
[gw8] [ 72%] FAILED tests/numba_cuda_tests/cudapy/test_ssa.py::TestReportedSSAIssues::test_issue3976 
tests/numba_cuda_tests/cudapy/test_ssa.py::TestReportedSSAIssues::test_issue3979 
```

ZzEeKkAa (approved):

Issue #249, opened 2026-08-01
[BUG] Compile time grows much faster than numba-cuda when a kernel composes many non-trivial device functions


raphaelchinchilla:
# Summary

A kernel that inlines N device functions compiles in time that grows roughly `N^1.8` under
`numba-cuda-mlir`, against roughly `N^1.3` under `numba-cuda`. With trivial device functions
`numba-cuda-mlir` is actually **faster**, so this only appears once the composed functions carry real
work (loops, transcendentals, branches).

The gap reaches ~3× by 32 composed functions in the reproducer below, and widens both with the
number of composed functions and with the size of each one — so it gets worse in exactly the
direction a fused kernel grows.

# Environment

| | |
|---|---|
| `numba-cuda-mlir` | 0.4.2 (`[cu12]`, `[cu13]`) |
| control | `numba-cuda` 0.30.4 + `numba` 0.66.0 |
| Python | 3.13 |
| measured on | GTX 1650 Ti (cc 7.5, CUDA 12.9); confirmed on Tesla T4, A10G and A100-SXM4-40GB (CUDA 13.3) |

# Reproducer

Builds N leaf device functions, folds them pairwise into a balanced tree (depth `O(log N)`), and
compiles one kernel that calls the root — the shape of any fused multi-block kernel. `WEIGHT` sizes
each leaf body: `0` is three lines, `>0` adds a bracketing loop plus a Newton refine with
transcendentals and branches.

```python
import math
import sys
import time

import numpy as np

BACKEND = sys.argv[1]                                  # "mlir" | "numba"
MAX_LEAVES = int(sys.argv[2]) if len(sys.argv) > 2 else 32
WEIGHT = int(sys.argv[3]) if len(sys.argv) > 3 else 6

if BACKEND == "mlir":
    from numba_cuda_mlir import cuda
else:
    from numba import cuda


def make_leaf(index):
    scale = 1.0 + index * 0.5
    weight = WEIGHT

    if weight == 0:
        @cuda.jit(device=True)
        def leaf(out, i):
            out[i] = out[i] * scale + index
        return leaf

    @cuda.jit(device=True)
    def leaf(out, i):
        lo, hi = -8.0, 8.0
        for _ in range(weight):
            mid = 0.5 * (lo + hi)
            take = math.exp(mid) - scale > 0.0
            hi = mid if take else hi
            lo = lo if take else mid
        t = 0.5 * (lo + hi)
        for _ in range(weight):
[...]

PR #259, opened 2026-08-10
Fix device function symbol collisions when inlining is disabled

atmnp:
This is a partial step towards fixing #249, this resolved a lingering correctness bug with disabled inlining. A better policy for inlining to resolve the compile-time issue is forthcoming.


jiel-nv:
LGTM. The instance id suffix deduplicates device callee symbols, and the updated cache condition avoids recompiling those with same signature and ABI name. Thanks for fixing this correctness issue!

jiel-nv (approved):

PR #261, opened 2026-08-11
Add overload variadic support

### LINKED ISSUE #257: [BUG] a `*args` `@overload` types but cannot be lowered
# BUG: a `*args` `@overload` types but cannot be lowered

**Component:** `numba_cuda_mlir.mlir_lowering` (`lower_overload_call`)
**Version:** numba-cuda-mlir 0.4.2, cuda-bindings 13.3.1, Python 3.12.13
**Device:** NVIDIA RTX A6000, driver 595.45.04
**Reproducer:** [`repro.py`](repro.py)

## Summary

An `@overload` whose typing function and implementation both take `*args`
passes type inference and then fails during lowering with an `AttributeError`
from inside the compiler. A variadic overload is the natural way to register
one entry point for a family of arities; without it the caller must register a
separate fixed-arity overload per arity.

## Reproducing

```
$ python repro.py
  *args overload -> AttributeError: 'tuple' object has no attribute 'name'
```

## Cause

Traceback (with `NUMBA_CUDA_MLIR_ICE_FULL_TB=1`):

```
  mlir_lowering.py:2250  in lower_call_expr_assign
  descriptor.py:1197     in builder
  mlir_lowering.py:1910  in lower_overload_call
  mlir_lowering.py:3288  in load_vars
  mlir_lowering.py:3292  in load_var
    numba_type = self.get_numba_type(var.name)
AttributeError: 'tuple' object has no attribute 'name'
```

`lower_overload_call` passes `call_vars` to `load_vars` (`mlir_lowering.py:1909`),
which assumes every element is a `numba_ir.Var`:

```python
call_args = [
    convert(val, ty) for val, ty in zip(self.load_vars(call_vars), callee_type.inputs)
]
```

For a variadic call the arguments arrive bundled: one element of `call_vars` is
a *tuple* of `Var`s rather than a `Var`, so `load_var`'s `var.name` raises. The
failure is a plain `AttributeError` escaping the compiler rather than a
diagnosed unsupported-feature error, which is what made this expensive to
identify.

## Suggested fix

Either flatten the variadic bundle before `load_vars` in `lower_overload_call`,
or — if variadic overloads are not intended to be supported — reject them at
registration or typing time with an explicit message, so the failure does not
surface as an internal `AttributeError` after type inference has succeeded.

## Workaround

Register one fixed-arity overload per arity up to some maximum. This works and
is what we do, but the arity ceiling is arbitrary and every additional arity
costs another registration.

ZzEeKkAa:
# Support `*args` in the overload family

Closes #257.

Closes #258.

Blocked by #274.

## Summary

An `@overload` whose typing function and implementation both take `*args` passed
type inference and then failed during lowering with `AttributeError: 'tuple'
object has no attribute 'name'`. Variadic implementations now work across
`@overload`, `@overload_method`, and `@register_jitable`, so one registration
serves a whole family of arities instead of requiring one fixed-arity
registration per arity up to some arbitrary ceiling.

Two independent defects were in the way.

### 1. Lowering did not flatten tuple arguments

`_fold_dispatcher_call_args` folds a call through `typing.fold_arguments`, whose
`stararg_handler` bundles the variadic arguments into a plain tuple of `Var`s.
`lower_overload_call` passed that list straight to `load_vars`, which assumed
every element was a `Var` — hence the `AttributeError` reported in the issue.

The callee is compiled with a `StarArgTuple` in that slot and flattens it to one
MLIR operand per leaf element, so the call site has to flatten to match. This
also fixes plain tuple arguments to overloads, which were broken the same way.

### 2. Dispatcher lookup could not match a folded signature

`get_overload_builder` compares the typed signature against `_impl_cache` keys.
The cache is keyed on the argument types seen at the call site, while the typed
signature has the `*args` arguments folded into a single tuple type, so the two
never matched for a variadic `@overload_method` and lowering fell through to
`NotImplementedError`.

The implementation dispatcher was compiled from the folded types, so its own
`nopython_signatures` is exactly what to compare against — no reconstruction of
the folded signature is needed.

## Changes

| File | Change |
| --- | --- |
| `src/numba_cuda_mlir/mlir_lowering.py` | Flatten loaded call arguments before pairing them with the callee's input types, and load the `*args` bundle element-wise |
| `src/numba_cuda_mlir/descriptor.py` | `_overload_args_match` — match an `_impl_cache` entry via the dispatcher's own compiled signature, falling back to the previous key comparisons |
| `tests/test_overload_varargs.py` | New, 15 tests |
| `docs/source/extending/high-level.rst` | New "Variadic implementations" section |

Net: **9 lines in `mlir_lowering.py`, 20 in `descriptor.py`.**

## Tests

`tests/test_overload_varargs.py` covers the three decorators that accept a
variadic implementation, and the shapes a bundle can take:

- `test_overload_varargs_only` — the reproducer from the issue
- `test_overload_varargs_arities` — arities 1, 2, 5 from one registration
- `test_overload_varargs_empty_bundle` — an empty bundle contributes no operands
- `test_overload_varargs_multiple_arities_in_one_kernel` — distinct arities must
  compile to distinct callees, not collide
- `test_overload_varargs_selects_impl_by_type` — two implementations of one
  variadic overload, differing in element type and arity at once, must not
  cross-talk in the `_impl_cache` lookup
- `test_overload_fixed_plus_varargs` — fixed parameters ahead of `*args`
- `test_overload_varargs_heterogeneous_types` — mixed leaf types in the bundle
- `test_overload_varargs_literal_argument` — a literal inside the bundle
- `test_overload_varargs_len_is_compile_time` — `len(args)` as a constant
- `test_overload_varargs_array_arguments` — array (memref) values in the bundle
- `test_overload_varargs_no_return_value` — void implementation
- `test_overload_method_varargs`, `test_register_jitable_varargs`

### Unxfailed

Two previously `xfail`ing tests now pass and their markers are removed:

- `test_ssa.py::test_issue5219` — passes a tuple to an overload, fixed by the
  argument flattening above.
- `test_extending.py::test_typing_vs_impl_signature_mismatch_handling_var_keyword`
  — matching on the dispatcher's compiled signature is more accurate than the
  previous key comparison, so the non-strict registration in that test now
  resolves and the `**kwargs` strictness assertions are reached.

### Results

Full suite on `2709fc2f`: **4547 passed, 82 skipped, 317 xfailed, 0 failed.**

`tests/numba_cuda_tests/cudapy/test_multigpu.py` is excluded — it fails
identically on a clean tree in a mixed-GPU environment and is unrelated to these
changes.

## Notes for reviewers

**The one judgement call.** Matching on `disp.nopython_signatures` is looser than
the key comparison it extends. That looseness is what unxfailed the `**kwargs`
test, and `test_overload_varargs_selects_impl_by_type` was added specifically to
guard the failure mode it could introduce. The exact key comparisons are still
tried, so this only adds matches that previously fell through to
`NotImplementedError`.

**Iterating a bundle.** A plain `for a in args` types the loop body once and so
requires the elements to share a type; `literal_unroll(args)` types it once per
element and handles heterogeneous bundles, including bundles of arrays. Both
spellings are exercised by the tests and documented in the new docs section.

## Out of scope

Two adjacent defects found while working on this, neither fixed here:

- `literal_unroll(range(n))` does not unroll — it is a silent no-op, despite
  being the only form shown in the public docstring. Written up separately in
  `internal/bug_literal_unroll_range.md`.
- Iterating a tuple of tuples raises `ValueError: Operand 0 of operation
  "func.return" must be a Value`, with or without `literal_unroll`, which blocks
  `for idx, val in literal_unroll(pairs)` destructuring. Not yet filed.


docs/source/extending/high-level.rst
```diff
@@ -77,6 +77,119 @@ finds a matching implementation. If no matching overload is found, then a
...
+   @cuda.jit
+   def kernel(out, a, b, c):
+       i = cuda.grid(1)
+       if i < out.size:
+           out[i] = my_sum(a[i], b[i], c[i])
+
+Each arity is compiled as a separate overload, so calling ``my_sum`` with two
+and with three arguments in the same kernel produces two distinct device
+functions. The same applies to
+:py:func:`~numba_cuda_mlir.extending.overload_method` and the other decorators
```
atmnp, line 117: `and the other decorators below` appears to include `overload_attribute`, where variable call arity does not apply. Could we enumerate the supported cases instead? For example: "Variadic implementations are also supported by `overload_method`; variadic `register_jitable` functions are supported as well."

tests/numba_cuda_tests/cudapy/test_extending.py
```diff
@@ -566,7 +566,6 @@ def foo(a, b):
         self.assertIn("VAR_POSITIONAL (e.g. *args) argument kind", msg)
         self.assertIn("offending argument name is '*star_args_token'", msg)
 
-    @pytest.mark.xfail(True, reason="Typing error")
```
atmnp, line 569: Could we keep this xfail removal out of this PR, or document how the overload lookup/lowering changes make this `**kwargs` typing-validation test pass? The production diff addresses post-typing overload selection and ABI lowering, while this test expects a `TypingError` during signature validation, so the connection is not clear.

tests/test_overload_varargs.py
```diff
@@ -0,0 +1,402 @@
...
+        acc += a
+    return acc
+
+
+def _run(kernel, out, *args):
+    kernel[1, out.size](out, *args)
+    return out
+
+
+def test_overload_varargs_only():
```
atmnp, line 178: Optional cleanup: this three-argument homogeneous case overlaps `test_overload_varargs_arities` and `test_overload_varargs_multiple_arities_in_one_kernel`. Could we drop it to keep the regression suite focused? No correctness concern if you prefer the explicit smoke test.

atmnp (changes requested):

atmnp:
@coderabbit review

ZzEeKkAa:
@CodeRabbit review

tests/test_overload_varargs.py
ZzEeKkAa: Dropped, thx

docs/source/extending/high-level.rst
ZzEeKkAa: Enumerated explicitly, thx

ZzEeKkAa:
@CodeRabbit review

PR #263, opened 2026-08-12
Normalize slices and allow dynamic shared mem views

### LINKED ISSUE #262: Negative-step and negative-bound slices return wrong data or fail to compile
**Describe the bug**

Slice lowering uses python bounds directly as view offsets and sizes. Negative or out-of-range bounds, omitted bounds with negative steps, and negative steps alone all produce wrong data or a compile error. Slices of dynamic shared memory also break because their extent reads as zero and the optimizer deletes their stores as out of bounds. Three vendored numba-cuda tests in test_sm.py are xfailed due to this.

```python
import numpy as np

@cuda.jit
def k(arr, out):
    view = arr[::-1]
    out[0] = view[0]

arr = np.arange(4, dtype=np.float32)
out = np.zeros(1, dtype=np.float32)
k[1, 1](arr, out)
print(out[0])
```

```python
import numpy as np

@cuda.jit
def k(idx, out):
    shared = cuda.shared.array(0, dtype=np.float32)
    shared[0] = 11.0
    shared[1] = 22.0
    shared[2] = 33.0
    view = shared[idx[0]:]
    view[0] = 99.0
    for i in range(3):
        out[i] = shared[i]

idx = np.array([1], dtype=np.int32)
out = cuda.to_device(np.zeros(3, dtype=np.float32))
k[1, 1, 0, 16](cuda.to_device(idx), out)
print(out.copy_to_host())
```

```python
import numpy as np

frozen = slice(-3, 3)

@cuda.jit
def k(arr, out):
    view = arr[frozen]
    out[0] = view[0]

arr = np.arange(4, dtype=np.float32)
out = np.zeros(1, dtype=np.float32)
k[1, 1](arr, out)
print(out[0])
```

**Expected behavior**

3.0 from the first kernel, [11.0, 99.0, 33.0] from the second, 1.0 from the third.

**Observed behavior** (sm_89)

0.0 from the first kernel, [11.0, 22.0, 33.0] from the second, a compiler error from the third.

**Root cause**

Slice bounds don't get wrapped, clamped to extent, step-sign defaulted, or python length computed. The numba slicing helpers are stubbed out, indicating that this is handled elsewhere, but I couldn't see that they were handled on this backend. The dynamic shared memory region is a zero-sized memref, so memref.dim makes its extent zero and clamping a slice to it collapses to an empty view. `__dynamic_shmem__` globals have internal linkage and zero length, so the optimizer sees any runtime-offset accesses as out of bounds and deletes the stores. This store-deletion was previously reported in #178, but a cleaner fix solved the specific repro in that case so it looked like it was fixable without externalising the `__dynamic_shmem__` linkage.

ccam80:
Fixes #262.

Slice bounds are passed through a single helper which wraps negative bounds, clamps to extent, deals with omitted bounds according to the step sign, and gets an exact length. A constant zero step raises at compile time, runtime zero steps set the kernel error code. Sliced and integer-indexed views get built with memref.extract_strided_metadata and memref.reinterpret_cast on the parent array's offset and strides instead of memref.subview, which can't express negative strides. strides reads the parent's real strides instead of assuming that the parent is C-contiguous.

Dynamic shared memory sizes its region from the %dynamic_smem_size register as numba-cuda does, so slices can get clamped to the real extent. Every shared.array(0) view aliases the whole region. `__dynamic_shmem__` globals get external linkage before codegen to stop stores getting optimised out.

Three dynamic-shared xfails in test_sm.py change to passing, 47 test_slicing.py tests added and passing, 38 of which fail on main. 2 test_dynamic_shared_memory.py tests added, both of which fail on main.

### DIFF
--- a/src/numba_cuda_mlir/lowering/cuda.py
+++ b/src/numba_cuda_mlir/lowering/cuda.py
@@ -608,37 +608,31 @@ def lower_strides(
     target: numba_ir.Var,
     array: numba_ir.Var,
 ):
-    from numba_cuda_mlir.lowering_utilities import int_of, index_of
+    from numba_cuda_mlir.lowering_utilities import index_of
 
+    array_numba_type = mlir_lower.get_numba_type(array.name)
     array = mlir_lower.load_var(array)
     array_type = array.type
     rank = array_type.rank
-    array_numba_type = mlir_lower.get_numba_type(array.name)
-
-    # Get element size in bytes
     element_size = storage_itemsize_bytes(array_numba_type)
 
-    # Get dimensions - handle both memref and tensor types
     if isinstance(array_type, ir.MemRefType):
-        dims = [
-            memref.dim(source=array, index=arith.constant(result=T.index(), value=i))
-            for i in range(rank)
+        metadata = memref.extract_strided_metadata(array)
+        element_strides = metadata[2 + rank : 2 + 2 * rank]
+        strides = [
+            arith.muli(index_of(stride), index_of(element_size)) for stride in element_strides
         ]
     elif isinstance(array_type, ir.RankedTensorType):
         dims = [
             tensor.dim(source=array, index=arith.constant(result=T.index(), value=i))
             for i in range(rank)
         ]
+        strides = [None] * rank
+        strides[-1] = index_of(element_size)
+        for i in range(rank - 2, -1, -1):
+            strides[i] = arith.muli(strides[i + 1], dims[i + 1])
     else:
         raise NotImplementedError(f"strides not implemented for {array_type}")
-
-    # Compute strides (C-contiguous order)
-    # stride[rank-1] = element_size
-    # stride[i] = stride[i+1] * dim[i+1]
-    strides = [None] * rank
-    strides[-1] = index_of(element_size)
-    for i in range(rank - 2, -1, -1):
-        strides[i] = arith.muli(strides[i + 1], dims[i + 1])
 
     mlir_lower.store_var(target, tuple(strides))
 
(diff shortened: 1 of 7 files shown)


atmnp:
Could we clarify expected behavior when cuda.shared.array(0) is declared before a runtime-shaped shared array? With this cursor update removed, a later cuda.shared.array(n) starts at the same current offset and overlaps the earlier dynamic view. The tests cover runtime-shaped-before-dynamic and dynamic/dynamic aliasing, but not this ordering. Is overlap intentional? If so, a test or brief contract note would make that explicit; otherwise allocation needs to be order-independent.

ccam80:
@atmnp good point as usual - the overlap is intentional, but on a closer look I'm not confident it's what you'd intend. I'm used to the numba-cuda dynamic shared memory idiom, where we take disjoint views of a `cuda.shared.array(0)` to create runtime-shaped arrays. This implementation swaps that view for the declaration of an array. if `runtime_2` and `runtime_4` are runtime variables with values 2 and 4:

```python
shared = cuda.shared.array(0, int64)
shared_one = shared[:runtime_2]
shared_two = shared[runtime_2:runtime_4].view(int32)
```

behaves the same way as

```python
shared = cuda.shared.array(0, int64)
shared_one = cuda.shared.array(runtime_2, int64)
shared_two = cuda.shared.array(runtime_4, int32)
```

That works, but it makes the meaning of `cuda.shared.array(0)` ambiguous. If I declare `cuda.shared.array(0)` in a device function, but I've already declared the snippet above in the kernel which calls it, then the zero-declaration in the device function now means:

- The whole dynamic memory region for the kernel, if inline="always"
- Only the region after the previous runtime-shaped allocations, if inline="never"
 
The second one might be surprising for a user. The existing test that a zero-declaration after a runtime-shaped array suggests that this is expected inside a kernel, and that makes sense to me, it just gets muddy when crossing into device functions written separately. 

I think that runtime-shaped arrays are a solid improvement, but I can't reconcile them cleanly with the zero-sized allocation approach across function boundaries. My gut feel is that using the zero allocation and views should work to support numba-cuda written code, the runtime-shaped pattern should work as the easier pattern for new users, and combining the two can work a bit weirdly for seasoned users.

If that fits with your intent for how dynamic shared arrays should work, I'll add some documentation and a test - if not, I'll rework the implementation.

PR #264, opened 2026-08-12
Support destructuring targets in consteval loop unrolling

atmnp:
It covers tuple/list targets, recursive unpacking, stored non-literal values, and explicit invalid-unpacking errors.

### DIFF
--- a/src/numba_cuda_mlir/ast_transforms/consteval.py
+++ b/src/numba_cuda_mlir/ast_transforms/consteval.py
@@ -43,15 +43,15 @@ class TargetOptionsReplacer(ast.NodeTransformer):
 
 
 class VariableReplacer(ast.NodeTransformer):
-    """Replace a variable name with a constant value throughout an AST."""
+    """Replace a variable name with an expression throughout an AST."""
 
-    def __init__(self, var_name: str, value):
+    def __init__(self, var_name: str, replacement: ast.expr):
         self.var_name = var_name
-        self.value = value
+        self.replacement = replacement
 
     def visit_Name(self, node: ast.Name) -> ast.AST:
         if node.id == self.var_name:
-            return ast.copy_location(ast.Constant(value=self.value), node)
+            return ast.copy_location(copy.deepcopy(self.replacement), node)
         return node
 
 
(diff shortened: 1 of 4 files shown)


tests/ast_transforms/test_loop_unroll.py
```diff
@@ -93,6 +93,83 @@ def kernel(arr):
...
+def test_unroll_tuple_target_invalid_unpacking_raises():
+    """Test that tuple targets require an exact number of values."""
+    ITEMS = ((0,),)
+
+    @numba_cuda_mlir.cuda.jit
+    def kernel(arr):
+        for idx, val in consteval(ITEMS):
+            arr[idx] = float(val)
+
+    with pytest.raises(ConstevalError, match="Cannot unpack 1 values into"):
```
brandon-b-miller: It should probably say `1 value` or better yet just `not enough values to unpack (expected N, got M)` like the standard python error

tests/ast_transforms/test_loop_unroll.py
```diff
@@ -93,6 +93,83 @@ def kernel(arr):
     assert all(result == 15.0)
 
 
+def test_unroll_tuple_target():
```
brandon-b-miller, line 96: This could be out of scope for this PR but do we have handling for `break` and `continue` under `consteval`?

src/numba_cuda_mlir/ast_transforms/consteval.py
```diff
@@ -244,15 +237,25 @@ def _unroll_for_loop(self, node: ast.For) -> list[ast.stmt]:
         for value in items:
             # Save current local_consts state
             saved_consts = self.local_consts.copy()
-            # Add loop variable to local_consts for this iteration
-            self.local_consts[var_name] = value
+            bindings = self._bind_loop_target(node.target, value)
+            # Add loop variables to local_consts for this iteration
+            self.local_consts.update(bindings)
+            replacements = {}
+            for var_name, bound_value in bindings.items():
```
brandon-b-miller, line 246: Similar to below, we do we do if we hit an `ast.Break`?

tests/ast_transforms/test_loop_unroll.py
atmnp: We don't support it, it used to fail, now it'll fail with an explicit error.

src/numba_cuda_mlir/ast_transforms/consteval.py
atmnp: Same as above.

tests/ast_transforms/test_loop_unroll.py
atmnp: Updated!

brandon-b-miller (approved):
LGTM.

PR #268, opened 2026-08-13
[Lowering] Booleans should be zero-extended not sign extended

### LINKED ISSUE #266: [BUG] Explicit bool->int cast sign-extends: int(True) gives -1 (silently wrong results)
## Description

An explicit scalar cast of a boolean in device code produces `-1` instead of `1`. There is no error — the kernel compiles and runs, and the results are silently wrong.

This looks like the `i1` being **sign**-extended rather than **zero**-extended (sign-extending `i1(1)` yields all-ones, i.e. `-1`).

## Reproducer

```python
import cupy as cp
import numpy as np

src = cp.asarray(np.array([True, False, True, True], dtype=np.bool_))
dst = cp.zeros(4, dtype=cp.int64)

@cuda.jit
def k(src, dst):
    i = cuda.grid(1)
    if i < src.size:
        dst[i] = int(src[i])

k[1, 32](src, dst)
print(dst.get().tolist())
```

Expected `[1, 0, 1, 1]` (and this is what `numba-cuda` gives).
Actual with `numba-cuda-mlir`: **`[-1, 0, -1, -1]`**.

## Only explicit casts are affected

Same kernel shape, varying only how the `bool` is converted to `int64`:

| conversion | numba-cuda | numba-cuda-mlir |
| --- | --- | --- |
| implicit store `dst[i] = src[i]` | `[1, 0, 1, 1]` | `[1, 0, 1, 1]` ✅ |
| **`int(x)`** | `[1, 0, 1, 1]` | **`[-1, 0, -1, -1]`** ❌ |
| **`np.int64(x)`** | `[1, 0, 1, 1]` | **`[-1, 0, -1, -1]`** ❌ |
| `x + 0` | `[1, 0, 1, 1]` | `[1, 0, 1, 1]` ✅ |
| `x * 1` | `[1, 0, 1, 1]` | `[1, 0, 1, 1]` ✅ |
| `1 if x else 0` | `[1, 0, 1, 1]` | `[1, 0, 1, 1]` ✅ |

So arithmetic, branching and implicit conversion all take a correct path; only the explicit scalar-constructor cast is mis-lowered. That narrowness is presumably why it has gone unnoticed.

Only `bool` is affected — the same explicit cast from `int8`, `uint8`, `int16` and `float32` all give correct results.

## Impact

This was found while evaluating `numba-cuda-mlir` for [scikit-hep/awkward](https://github.com/scikit-hep/awkward). `ak.sum` over a boolean array returns the **negation** of the true count:

```
CPU : [1, 0, 1, 0]
GPU : [-1, 0, -1, 0]
```

Awkward reaches the cast by fusing a widening `astype` into a reduction (an op whose body is `int(x)`), but nothing about that is special — any `int(bool)` in device code is affected.

Because it is silent, this seems worth prioritising over louder failures: a caller gets plausible-looking negative counts rather than an exception.

## Possibly related prior work

- #80 — introduced the value/storage data model to fix `bool`/`float16` memory representation. That addressed storage/ABI; the explicit-cast path appears not to be covered.
- #189 — registered Boolean comparison lowerings with unsigned `i1` semantics.

## Environment

```
numba-cuda-mlir  0.4.2
numba-cuda       0.30.4
numba            0.65.1
cupy-cuda13x     14.1.1
Python           3.13.15
CUDA             13.2, driver 595.91.07
GPU              NVIDIA L40S (sm_89), x86_64
```

Reproduces with both released `cuda-cccl` 1.1.1 and a source build of NVIDIA/cccl#9421, and with no `cuda.compute` in the picture at all.

atmnp:
This closes #266.

### DIFF
--- a/src/numba_cuda_mlir/lowering/builtins.py
+++ b/src/numba_cuda_mlir/lowering/builtins.py
@@ -450,6 +450,7 @@ def type_convert(builder, target, args, kwargs):
     target_numba_ty = builder.get_numba_type(target)
     to_type = builder.get_mlir_type(target)
     to_signed = isinstance(target_numba_ty, types.Integer) and target_numba_ty.signed
+    source_numba_ty = builder.get_numba_type(args[0])
 
     if isinstance(target_numba_ty, (types.IntegerLiteral, types.Literal)):
         result = constant(target_numba_ty.literal_value, to_type)
@@ -470,7 +471,11 @@ def type_convert(builder, target, args, kwargs):
             else arith.fptoui(out=to_type, in_=value)
         )
     else:
-        value = convert(value, to_type, signed=to_signed)
+        value = convert(
+            value,
+            to_type,
+            signed=to_signed and not isinstance(source_numba_ty, types.Boolean),
+        )
     builder.store_var(target, value)
 
 
--- a/tests/test_value_storage_runtime.py
+++ b/tests/test_value_storage_runtime.py
@@ -61,6 +61,20 @@ def test_bool_reductions_read_value_elements():
     np.testing.assert_array_equal(flags_out, np.array([1, 0], dtype=np.int32))
 
 
+def test_bool_to_integer_constructor_conversion_uses_zero_extension():
+    @cuda.jit
+    def kernel(out, values):
+        out[0] = int(values[0])
+        out[1] = int(values[1])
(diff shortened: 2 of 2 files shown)


VijayKandiah (approved):
LGTM!

Issue #267, opened 2026-08-12
[BUG] getattr on a NumPy dtype inside device code raises internal compiler error


shwina:
## Description

A kernel (or `cuda.compute` operator) that closes over a NumPy `dtype` object and accesses an attribute on it *inside* device code fails to compile with an internal compiler error:

```
numba_cuda_mlir.errors.UserFacingInternalCompilerError:
    This line in your program has triggered a compiler bug
```

With `NUMBA_CUDA_MLIR_ICE_FULL_TB=1` the underlying cause is:

```
NotImplementedError: getattr(dtype(int64), attr='type')
```

raised at `numba_cuda_mlir/mlir_lowering.py:2652`.

The same code compiles and runs correctly under `numba-cuda`.

## Reproducer

```python
import cupy as cp
import numpy as np
from numba_cuda_mlir import cuda

idx_dtype = np.dtype(np.int64)

@cuda.jit
def k(dst):
    i = cuda.grid(1)
    if i < dst.size:
        dst[i] = idx_dtype.type(0) if i == 0 else i

dst = cp.zeros(4, dtype=cp.int64)
k[1, 32](dst)          # <-- UserFacingInternalCompilerError
print(dst.get().tolist())
```

Expected `[0, 1, 2, 3]` (what `numba-cuda` produces).

## Workaround

Hoisting the attribute lookup to the host, so the closure captures the scalar constructor rather than the `dtype` object, compiles fine on both backends:

```python
idx_ctor = idx_dtype.type      # host-side

@cuda.jit
def k(dst):
    i = cuda.grid(1)
    if i < dst.size:
        dst[i] = idx_ctor(0) if i == 0 else i
```

So the value is perfectly representable — it is the `getattr` on the dtype token inside the compiled body that is unsupported.

## Impact

Found while evaluating `numba-cuda-mlir` for [scikit-hep/awkward](https://github.com/scikit-hep/awkward), via NVIDIA/cccl#9421 (which routes `cuda.compute` operator compilation through `numba-cuda-mlir`).

This was by a wide margin the **largest single cause of failures** in that evaluation: roughly 331 occurrences, accounting for the bulk of 387 failing tests across `tests-cuda`. Awkward captures dtypes this way in several reducer/index helpers, e.g.

```python
def fill_nextoffsets(j):
    stop = offsets[j + 1]
    nextoffsets[j + 1] = idx_dtype.type(0) if stop == 0 else scan[stop - 1]
```

[...]

PR #269, opened 2026-08-13
Support dtype type and kind attributes

atmnp:
Lower captured dtype .type as the existing compile-time type-token representation and .kind as its typed string literal in MLIR lowering.

Cover integer and floating dtype closures in tests/test_dtype_attributes.py.


VijayKandiah (approved):
LGTM for the `.type`/`.kind` gap this is addressing. However, this change does not fully close #267 which also calls out attribute access on a captured array and that still fails with this change:

```python
import numpy as np

ARR = np.arange(4, dtype=np.int64)

@cuda.jit
def k(dst):
    dst[0] = ARR.itemsize      # UserFacingInternalCompilerError

k[1, 1](np.zeros(1, dtype=np.int64))
```

We should leave #267  open and address the above in a follow up change

PR #270, opened 2026-08-17
Support static dtype metadata in device code

atmnp:
Resolve dtype scalar, string, and structural metadata attributes as compile-time literals or type tokens in MLIR lowering.

Cover scalar, record, and subarray dtype metadata access.

This closes @267.

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/typing/arraydecl.py
+++ b/src/numba_cuda_mlir/numba_cuda/typing/arraydecl.py
@@ -7,6 +7,7 @@ from collections import namedtuple
 
 from numba_cuda_mlir.numba_cuda import types
 from numba_cuda_mlir.numba_cuda import utils
+from numba_cuda_mlir.numba_cuda.np import numpy_support
 from numba_cuda_mlir.numba_cuda.typing.templates import (
     AttributeTemplate,
     AbstractTemplate,
@@ -568,6 +569,14 @@ class ArrayAttribute(AttributeTemplate):
 class DTypeAttr(AttributeTemplate):
     key = types.DType
 
+    @staticmethod
+    def _dtype(ary):
+        return numpy_support.as_dtype(ary.dtype)
+
+    @classmethod
+    def _literal_attr(cls, ary, attr):
+        return types.literal(getattr(cls._dtype(ary), attr))
+
     def resolve_type(self, ary):
         # Wrap the numeric type in NumberClass
         return types.NumberClass(ary.dtype)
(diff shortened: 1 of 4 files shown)


src/numba_cuda_mlir/numba_cuda/typing/arraydecl.py
```diff
@@ -581,6 +590,51 @@ def resolve_kind(self, ary):
...
+        return self._literal_attr(ary, "str")
+
+    def resolve_byteorder(self, ary):
+        return self._literal_attr(ary, "byteorder")
+
+    def resolve_names(self, ary):
+        names = self._dtype(ary).names
+        if names is None:
+            return types.none
+        return types.Tuple(tuple(types.literal(name) for name in names))
```
VijayKandiah, line 630: The new test establishes `dtype.names is None` as the way to feature-detect a record dtype (`out[13] = int_dtype.names is None`), but only the `None` side works. The record case still gives an error:
```python
import numpy as np
RECORD = np.dtype([("x", np.int32), ("y", np.float64)])
@cuda.jit
def k(out):
    out[0] = RECORD.names is None      # UserFacingInternalCompilerError
k[1, 1](np.zeros(1, dtype=np.int64))
```
```
NotImplementedError: NotImplemented lowering binop fn=<built-in function is_>
```

So `if dtype.names is None: ...` works for scalar dtypes but fails for record ones. The `is_` registry just has no tuple case, while the new test makes `is None` look supported. Could we lower `is_` for Tuple-vs-None and cover the record case as well in the test?

atmnp: Done!

src/numba_cuda_mlir/numba_cuda/typing/arraydecl.py
```diff
@@ -581,6 +590,51 @@ def resolve_kind(self, ary):
...
+    def resolve_name(self, ary):
+        return self._literal_attr(ary, "name")
+
+    def resolve_str(self, ary):
+        return self._literal_attr(ary, "str")
+
+    def resolve_byteorder(self, ary):
+        return self._literal_attr(ary, "byteorder")
+
+    def resolve_names(self, ary):
```
brandon-b-miller, line 626: One small question here - is this attribute of a structured dtype technically mutable? Does the behavior being introduced here match what a user would expect out of a normal python usage of the dtype? 

Since this PR freezes the attr literals into the kernel, does adjusting the `names` on the python side trigger a recompilation?

atmnp: iirc this is why numba-cuda supported a limited set of attrs, but #267 points out that maybe we should support more (#269 adds the ones numba-cuda had). Right now, this doesn't allow for a mutation on the device, if you mutate on the host, the *user* in charge of making sure the kernel is recompiled and you're not using the cached one by accident.

brandon-b-miller: Whats the recompilation behavior for host-mutating other kinds of globals that might be frozen in, like a list? If that triggers a recompilation, I think we ought to here as well.

That said even if it is the case, I don't think we should _block_ on it here - spinning off an issue is probably fine, that gives a user something to find if they ever run into this.

brandon-b-miller (approved):

VijayKandiah (approved):
LGTM now

src/numba_cuda_mlir/numba_cuda/typing/arraydecl.py
atmnp: In general, most ways of capturing globals don't trigger recompilation automatically on a mutation atm.

PR #272, opened 2026-08-18
Keep array arguments in DWARF

jiel-nv:
Array arguments were missing from a kernel's DWARF under `debug=True`: no `DW_TAG_formal_parameter` and no descriptor struct.

When emitting full debug info, adding volative store to prevent the aggregate slots of array descriptors from being optimized away, together with the `dbg.declare` even at `opt=0`, by libnvvm.

Also, enhancing the LLVM70 translator with the capability of converting and building parameter di type and composite di type.

A new LLVM IR test is added.

### DIFF
--- a/cext/mlir-llvm70/lib/LLVM70Target.cpp
+++ b/cext/mlir-llvm70/lib/LLVM70Target.cpp
@@ -1211,7 +1211,9 @@ llvm::Error MLIRToLLVM70::translateStoreOp(Operation *op) {
                     .getAddressSpace();
   ptr = b.buildBitCast(ptr, b.ptrTy(elemTy, as), "");
 
-  b.buildStore(val, ptr);
+  LLVMValueRef storeInst = b.buildStore(val, ptr);
+  if (storeOp.getVolatile_())
+    b.setVolatile(storeInst, true);
   return llvm::Error::success();
 }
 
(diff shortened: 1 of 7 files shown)


tests/test_cuda_gdb.py
atmnp, line 115: Can we also add a check for `out`?

cext/mlir-llvm70/lib/LLVM70Target.cpp
```diff
@@ -2509,29 +2511,137 @@ llvm::Error MLIRToLLVM70::translateNVVMOp(Operation *op) {
...
+
+  bool isStructLike = tag == llvm::dwarf::DW_TAG_structure_type ||
+                      tag == llvm::dwarf::DW_TAG_class_type ||
+                      tag == llvm::dwarf::DW_TAG_union_type;
+  if (!isStructLike)
+    return opaqueDIType(size);
+
+  // Create a forward declaration first: translating the members recurses
+  // through getOrCreateDIType and may come back to this same type.
+  LLVMMetadataRef fwdDecl = b.createDIForwardDecl(
```
atmnp: Are we sure this recursion works as expected? Can we add a test for this?

tests/test_cuda_gdb.py
jiel-nv: Yes, updated in https://github.com/NVIDIA/numba-cuda-mlir/pull/272/commits/43fa3ff1cb284245b56064373b94196d85e0c9e9

cext/mlir-llvm70/lib/LLVM70Target.cpp
jiel-nv: Thank you for catching it. The recursion logic is tweaked and a test is added in https://github.com/NVIDIA/numba-cuda-mlir/pull/272/commits/43fa3ff1cb284245b56064373b94196d85e0c9e9

jiel-nv:
/ok to test https://github.com/NVIDIA/numba-cuda-mlir/pull/272/commits/a5bcfa3c28415934e65badacfb2302b40e8bf1e6

atmnp (approved):
LGTM assuming green CI

PR #274, opened 2026-08-20
Fix consteval loop unroll

ZzEeKkAa:
# Unroll `consteval` over a tuple parameter

## Summary

`for x in consteval(t)`, where `t` is a parameter, substituted the parameter's
member **types** into the loop body as if they were values. Unrolling a
parameter now emits element accesses — `t[0]`, `t[1]`, … — leaving the values at
runtime.

## The problem

`consteval` unrolls a loop by evaluating its argument at compile time and
substituting each item into a copy of the body. A parameter name does not
evaluate to a value: it resolves through `param_type_map` to the parameter's
Numba type. Iterating a tuple type yields its member types, and those were then
substituted as though they were the elements:

```python
@cuda.jit
def kernel(out, t):          # t: Tuple(float32, float64)
    for a in consteval(t):
        out[0] += a
```

unrolled to

```python
__consteval_0__ = float32    # a type object, not a value
__consteval_1__ = float64
def kernel(out, t):
    out[0] += __consteval_0__
    out[0] += __consteval_1__
```

Usually that fails during type inference with an error that does not mention
`consteval` or the loop. **It can also compile and produce a wrong result.** When
the loop variable is used in a way that is valid for a type — a cast, a `dtype=`
argument, a `.dtype` access — the substitution type-checks and the tuple's actual
values are silently ignored:

```python
@cuda.jit
def kernel(out, t):
    i = 0
    for a in consteval(t):
        out[i] = a(7)        # `a` is float32/float64, so this is a cast
        i += 1

kernel[1, 1](out, (1.5, 2.5))   # -> [7. 7.]  (1.5 and 2.5 never read)
```

This is not specific to any one kind of parameter — a `*args` bundle and an
explicit tuple argument are both affected, since both are tuple-typed
parameters.

## The fix

`_parameter_element_exprs` recognises `consteval(<name>)` where the name is a
parameter, and returns one `ast.Subscript` per element instead of letting the
name be evaluated. The existing unroll loop substitutes those, so the body
becomes:

```python
def kernel(out, t):
    out[0] += t[0]
    out[0] += t[1]
```

Only the indices are compile-time, which is all the unroll needed. Each element
access is typed independently, so elements may have differing types.

A parameter that is not a tuple has no compile-time length, and now raises
`ConstevalError` rather than silently substituting its type.

The substitution itself needed no new machinery: #264 already changed
`VariableReplacer` to accept an arbitrary `ast.expr`, so the only change inside
the unroll loop is one branch that passes an expression through instead of
wrapping it as a constant.

## Tests

Five added to `tests/ast_transforms/test_loop_unroll.py`, following the file's
existing convention of asserting on `transformed_source` and on results:

- `test_unroll_tuple_parameter` — the emitted source contains `t[0]` / `t[1]`
  and no loop
- `test_unroll_tuple_parameter_runs` — values reach the body, with
  `(int32, float32, float64)` elements to cover independent typing per access
- `test_unroll_varargs_bundle_runs` — a `*args` bundle unrolls the same way
- `test_unroll_non_tuple_parameter_rejected` — `ConstevalError`, raised before
  evaluation, so distinct from the existing `test_unroll_non_iterable_raises`
- `test_unroll_tuple_parameter_destructuring_rejected` — a parameter element is
  a single expression, so it cannot be unpacked into a tuple target; distinct
  from the existing `test_unroll_tuple_target_invalid_unpacking_raises`, which
  covers the length-check path

Full suite: 4614 passed, 82 skipped, 299 xfailed, 0 failed.
`tests/numba_cuda_tests/cudapy/test_multigpu.py` is excluded; it fails
identically on a clean tree in a mixed-GPU environment.

## Notes for reviewers

**Destructuring message.** `for i, v in consteval(t)` over a tuple-of-tuples
parameter is rejected by the existing `_bind_loop_target`, which reports
`Cannot unpack Subscript into (i, v)`. That leaks AST vocabulary. Adding a
dedicated check ahead of it would give a clearer message at the cost of a few
more lines; it seemed not worth a second code path for a rare case, but say so
if you disagree.

**Possible follow-up, not done here.** Destructuring over a tuple-of-tuples
parameter could be supported by binding recursively — `i` to `t[0][0]`, `v` to
`t[0][1]` — rather than rejected.

**Shadowing.** `_parameter_element_exprs` consults `param_type_map` directly,
matching the precedence in the `context` property, where parameter types shadow
locals and globals of the same name.


atmnp:
@coderabbit review

ZzEeKkAa:
@CodeRabbit review

ZzEeKkAa:
@CodeRabbit review

ZzEeKkAa:
@CodeRabbit review

atmnp (approved):

PR #275, opened 2026-08-21
Fix duplicate overload call

ZzEeKkAa:
# Memoize overload type resolution across flag contexts

Ports [NVIDIA/numba-cuda#914](https://github.com/NVIDIA/numba-cuda/pull/914) to
numba-cuda-mlir, with the caching mechanism reworked to fit this codebase and the
scope claims corrected against what actually happens here.

## Summary

`@overload` implementation functions are re-executed once per distinct compiler-`Flags`
context in which a call is resolved. The overload body is where user code does its
(sometimes very expensive) work, so this duplicates that work whenever the same
overloaded call is resolved under more than one flag context.

`_OverloadFunctionTemplate` now memoizes the overload function's result on the
**overload function + argument types** — the inputs it actually depends on — so the body
runs at most once per distinct signature, while the flag-sensitive `_impl_cache` is left
untouched.

## Root cause

`_OverloadFunctionTemplate._get_impl` keys its implementation cache on the active flags:

```python
flags = targetconfig.ConfigStack.top_or_none()
cache_key = self.context, tuple(args), tuple(kws.items()), flags
```

That is correct for what `_impl_cache` stores — a Dispatcher compiled under those flags —
but `_build_impl` calls `self._overload_func(*args, **kws)` behind the same key. The
overload function is passed only the argument *types*, never the flags, so its result is
flag-independent and gets recomputed for no benefit.

## What reproduces here

Worth stating precisely, because it differs from the upstream PR. Upstream describes
double execution **within a single kernel compile** (type inference vs. a force-inlined /
LTO device-function compile). Instrumenting `_build_impl` shows that **does not happen in
numba-cuda-mlir** — a kernel calling an overload through an `inline=True` device function
under `lto=True` reaches `_build_impl` exactly once.

What does reproduce is duplication **across** flag contexts: two kernels differing only in
`lto` that share a device function calling an overloaded function.

| | before | after |
| --- | --- | --- |
| `@overload` | 2 executions | 1 |
| `@overload_method` | 2 executions | 1 |
| `@overload_attribute` | 2 executions | 1 |

All three decorators are covered by the single change. `_OverloadAttributeTemplate` has no
`_build_impl` of its own, but `overload_attribute` / `overload_method` in
`numba_cuda_mlir.extending` additionally register the overload function as an `@overload`
of itself, and `_get_function_type` resolves through that function template — so their
bodies land in the same code path.

## Implementation

A plain per-template dict, mirroring the `_impl_cache` idiom already in the file:

```python
cache = self._overload_result_cache
key = self._overload_func, tuple(args), tuple(kws.items())
if key not in cache:
    cache[key] = self._overload_func(*args, **kws)
return cache[key]
```

with `_overload_result_cache={}` alongside `_impl_cache={}` in `make_overload_template`.

`_overload_func` is part of the key so a subclass overriding it gets its own entries in
the dict it inherits from the parent template, rather than silently receiving the
*parent's* implementation.

This diverges from upstream, which binds a per-subclass `functools.lru_cache` lazily via a
`cls.__dict__` probe. Both are correct; the dict version avoids mutating the class from
inside a method and reads like the cache that is already there.

## Alternative considered and rejected

**Dropping `flags` from the `_impl_cache` key** would be a one-term deletion instead of a
new cache, but it does not work. `_impl_cache` stores a Dispatcher that `_build_impl`
compiles under the active flags, so sharing one across contexts returns an artifact built
for the wrong flags. Measured: immediate ICE on the basic repro
(`NotImplementedError: NotImplemented lowering call to ...`) and **102 test failures**
against a 2-failure baseline. The two-level split is inherent — the overload *result* is
flag-independent, the *compiled artifact* is not. Noted in the docstring so it is not
retried.

## Tests

New `tests/test_overload_caching.py`, 7 tests. Two are end-to-end (they fail on the
assertion without the fix, not merely on a missing attribute):

- `test_shared_device_function_across_flag_contexts` — the `@overload` repro above.
- `test_overload_method_and_attribute_across_flag_contexts` — same for
  `@overload_method` / `@overload_attribute`. Worth having precisely because that
  coverage is *indirect*: it depends on the self-registration lines in
  `numba_cuda_mlir.extending`, and if either were refactored away both decorators would
  silently regress with nothing else catching it.

The rest pin the cache key and its scope: `test_distinct_arg_types_run_again`,
`test_kwargs_participate_in_key`, `test_cache_is_per_template`,
`test_subclass_does_not_inherit_parent_cache`, `test_cache_lives_on_template_class`.

Each was mutation-tested — every way of weakening the key or its scope (dropping
`_overload_func`, `args`, or `kws`; moving the cache to the instance) is caught by at
least one test.

## Verification

- Full suite: **4599 passed, 2 failed, 83 skipped, 301 xfailed, 0 xpassed**.
- The 2 failures are pre-existing and unrelated — `test_multigpu.py::TestMultiGPUContext::{test_multigpu_context,test_with_context}`,
  which fail identically with the change reverted (`no kernel image is available for
  execution on the device`, a second GPU of a different arch on the test host).
- No xfails were accidentally fixed.
- `ruff format` / `ruff check` clean.

## Notes for reviewers

- The memoization assumes the overload body is a pure function of the argument types. It
  is passed only types, and no overload body in this repo reads
  `ConfigStack.top_or_none()` itself — but a user overload could, which would make the
  assumption unsound. The docstring states this as an assumption rather than a fact.
- The `(signature, pyfunc)` return path is now memoized too, even though `_build_impl`
  deliberately keeps it out of `_impl_cache` (`cache_key = None  # don't cache`). That
  comment refers to not caching the *Dispatcher*; the body no longer re-runs there.
- `test_overload_method_and_attribute_across_flag_contexts` registers attributes on
  `types.Array` in the global typing registry, which persists for the session. This
  matches what `tests/test_numba_cuda_mlir_extending.py` already does, but it is global
  state the test leaves behind.


tests/test_overload_caching.py
```diff
@@ -0,0 +1,284 @@
...
+    assert len(calls_a) == 1
+    assert len(calls_b) == 1
+    assert ra1 is ra2
+    assert ra1 is not rb1
+
+
+def test_subclass_does_not_inherit_parent_cache():
+    """A subclass overriding ``_overload_func`` must get its own cache.
+
+    This is what the ``cls.__dict__`` lookup in ``_call_overload_func`` buys: a plain
```
atmnp: Based on the PR description, this is referring to the `numba-cuda` impl but it could read like we use the `cls.__dic__` lookup. Can this be clarified upon?

ZzEeKkAa: Thank you! That comment left from the previous implementation iteration.

atmnp:
@coderabbit review

atmnp (approved):
LGTM!

atmnp (changes requested):
Actually, let's not merge this until we figure out how to handle cases where the overload is specialized on `lto` or other `Flags` via `ConfigStack`.

atmnp:
@coderabbit review

atmnp:
`fix_overload_dispatcher` still needs to be changed to use this new path (for attributes) +  we need to scan all matching arg entries for a flag match, it doesn't seem like we do that today.

ZzEeKkAa:
@CodeRabbit review

ZzEeKkAa:
@CodeRabbit review

src/numba_cuda_mlir/numba_cuda/typing/templates.py
```diff
@@ -756,6 +875,54 @@ def _get_jit_decorator(self):
...
+        observed = set()
+        try:
+            result = _run_recording_flag_reads(self._overload_func, args, kws, observed)
+        finally:
+            if not observed <= known:
+                # The body read options we had not seen before (a first run, or a
+                # branch that consults more of them).  Widen the key set from here on
+                # -- including when the body raised, since the read still happened.
+                known = frozenset(known | observed)
+                reads[base] = known
```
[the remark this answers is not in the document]
ZzEeKkAa: It describes the union design that no longer exists. No longer in the code.

ZzEeKkAa:
@CodeRabbit review

atmnp (approved):

PR #278, opened 2026-08-24
Use already resolved type promotion for integer binary ops

### LINKED ISSUE #277: 0.5.0 regression: `uint64 * <integer literal>` lowers to `arith.mulf` and fails MLIR verification
Multiplying a `uint64` value by an **integer literal** in device code lowers to
`arith.mulf` (a *floating-point* multiply) on the raw `i64` operands, so the
module fails verification:

```
error: 'arith.mulf' op operand #0 must be floating-point-like, but got 'i64'
```

This compiled fine on 0.4.2, so it is a 0.5.0 regression — it looks like a side
effect of the #121 signedness rework. It's narrow: only `uint64 * literal` is
affected.

## Environment
- numba-cuda-mlir 0.5.0
- CUDA Toolkit 13.3
- Python 3.12, Linux

## Reproducer
```python

def mul2(x):
    return x * 2

cuda.compile(mul2, T.uint64(T.uint64), device=True, abi="c",
             abi_info={"abi_name": "m"}, output="ltoir")
```
→ `InternalCompilerError: Invalid MLIR module: 'arith.mulf' op operand #0 must
be floating-point-like, but got 'i64'`. The emitted op is:
```mlir
%0 = arith.constant 2 : i64
%1 = arith.mulf %arg0, %0 : (i64, i64) -> i64   // should be arith.muli
```

## Scope (what compiles vs. what doesn't)
| expression | result |
|---|---|
| `uint64 * 2`  (literal) | ❌ `arith.mulf` on i64 |
| `uint32 * 2`            | ✅ |
| `int64 * 2`             | ✅ |
| `uint64 * x` (var × var)| ✅ |
| `uint64 + 2`            | ✅ |
| `float64 * 2`           | ✅ |

Only `uint64` × an integer literal, and only multiply, is affected.

## Likely cause
numba unifies `uint64` with a signed integer literal to `float64` (no common
*integer* type covers both ranges). The lowering then picks the float multiply
`arith.mulf` for the promoted type but applies it to the un-converted `i64`
operands, so the operand types don't match the op. Either the operands should
be converted to `f64` first, or (better) `uint64 * literal` should stay integer
(`arith.muli`) with the literal taken as `uint64`.

## Impact
Any device operator that multiplies a `uint64` by a constant breaks — e.g. a
`lambda x: x * 2` transform over uint64 data, or segmented-reduction offset
arithmetic on uint64 offsets.

brandon-b-miller:
Closes https://github.com/NVIDIA/numba-cuda-mlir/issues/277

### DIFF
--- a/tests/numba_cuda_tests/cudapy/test_operator.py
+++ b/tests/numba_cuda_tests/cudapy/test_operator.py
@@ -98,6 +98,18 @@ def simple_int_lt(ary, a, b):
     ary[0] = a < b
 
 
+def simple_int_mul_literal(ary, a):
+    ary[0] = a * 2
+
+
+def simple_int_sub_literal(ary, a):
+    ary[0] = a - 2
+
+
+def simple_int_add_literal(ary, a):
+    ary[0] = a + 2
+
+
 @cuda.jit("b1(f2, f2)", device=True)
 def hlt_func_1(x, y):
     return x < y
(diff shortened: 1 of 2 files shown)


rparolin (approved):
lgtm

tests/numba_cuda_tests/cudapy/test_operator.py
```diff
@@ -98,6 +98,18 @@ def simple_int_lt(ary, a, b):
     ary[0] = a < b
 
 
+def simple_int_mul_literal(ary, a):
```
rparolin, line 101: Non-blocking: But I'd also consider testing having the literal on the lhs.

```
def simple_literal_mul_int(ary, a):
    ary[0] = 2 * a

def simple_literal_sub_int(ary, a):
    ary[0] = 2 - a

def simple_literal_add_int(ary, a):
    ary[0] = 2 + a
```

rparolin (approved):
lgtm

tests/numba_cuda_tests/cudapy/test_operator.py
brandon-b-miller: https://github.com/NVIDIA/numba-cuda-mlir/pull/278/changes/ee5d6c9fbb433dd7da463bfe20bf2838d6be5bbb

PR #279, opened 2026-08-26
Add typed pointer support for llvm70 lowering

ZzEeKkAa:
# Recover struct pointee types for the llvm70 backend

Below `sm_100` a module is translated to the LLVM 7 dialect, whose pointers are
typed. MLIR's are opaque, so every pointer in an identified struct becomes `i8*`.
That is self-consistent until the module is linked against IR from a typed
producer — an NVRTC-compiled device function, say — which spells the same member
`float*`. Two identified structs of the same shape with different pointee types
are **different LLVM types**, so the call between them links through a bitcast
and stops being inlinable.

This recovers the pointee and passes it to the translation, so the struct built
here matches the one coming from the typed producer.

## No new user-facing API

Numba-CUDA never had this problem: its `CPointerModel` builds
`dmm.lookup(fe_type.dtype).get_data_type().as_pointer()`, so the typed pointer
falls out of the datamodel. Here `to_mlir_type` maps `CPointer` to a bare
`!llvm.ptr` and the pointee is lost from the **IR** — but not from the **type**.

`AggregateType.__init__` already calls `record_named_type`, so
`_map_be_type_names_to_fe_types` maps every identified struct's name to the Numba
type it was built from. `fe_type.fields` gives each member's type and
`CPointer.dtype` is the pointee. Nothing needs to be declared; it is all
derivable, and an earlier draft of this change that added a
`register_struct_pointees` extension API was dropped for exactly that reason —
it duplicated an existing registry and asked users to re-supply what the
compiler already knew.

Only the *transport* to the C++ translator is new: a `llvm70.struct_pointees`
attribute on the `gpu.module`.

## Implementation

**Python** (`struct_pointees.py`, new). `struct_pointees_attr(op)` walks the
module's types, and for each identified struct it finds, looks up the recorded
Numba type and emits `{struct name: [pointee per member]}`. Members with no
recoverable pointee become `UnitAttr` — read as "no hint" — so a struct can have
some members described and the rest left at `i8*`.

The walk is restricted to structs the module actually uses. The recorded-type
registry is process-wide and accumulates every struct seen so far; emitting all
of it would describe structs the module never mentions. The translation ignores
those, but a dumped module would not.

Struct types are reached through block arguments, op results, and type
attributes — `llvm.alloca`'s `elem_type`, function signatures, and `llvm.byval`
argument attributes — then transitively through struct members, array elements
and function signatures. Operands are not visited: each is a block argument or a
result, so it has been seen already.

`llvm.byval` is worth calling out. It carries a struct type that appears nowhere
else in the IR — the argument itself is just `!llvm.ptr` — and it is nested two
deep in `arg_attrs` (ArrayAttr → DictAttr → TypeAttr). This repo uses it for
`CUTensorMap`. Missing it would silently drop that struct's refinement.

**C++** (`LLVM70Target.cpp`). `loadStructPointees` reads the attribute;
`convertStructMemberType` applies it when building a named struct body.

## Translator fixes this required

The refinement breaks the translator's invariant that a value of MLIR
`!llvm.ptr` type is represented as `i8*` (the invariant `translateAddressOfOp`
already maintains). Two defects, both reproduced by running `llvm70-translate`,
not by inspection:

1. **`llvm.insertvalue` emitted invalid IR** — the value went in as
   `i8 addrspace(1)*` while the member was `float addrspace(1)*`. libnvvm:
   *"Inserted value type doesn't match aggregate type"*. This is the primary
   construction path, so the feature was broken for its own use case.

2. **`llvm.extractvalue` leaked the refined type** into the value map. Storing an
   extracted member failed with *"Explicit load/store type does not match pointee
   type of pointer operand"*.

The fix keeps the refinement inside the struct layout: insertvalue casts in,
extractvalue casts back out. Both casts fold away when nothing was refined, so
undeclared structs are bit-for-bit unchanged. Pointer members stay 64-bit
regardless of pointee, so struct layout and GEP offsets do not move — a wrong
pointee costs a bitcast at the link, never a miscompile.

### One pre-existing bug, separable

While probing (2) I found the same class of defect already on `main`:
`llvm.store` of an alloca or GEP result into a pointer slot emitted
`store float* %2, i8** %3`. It is independent of this feature — the refinement
just makes it much easier to hit — and it is a two-line coercion in
`translateStoreOp`. **Happy to split it into its own commit if preferred.**

## Tests

- `cext/mlir-llvm70/test/struct_pointees.mlir` — refinement in the struct
  layout; a declared non-pointer member ignored; partial declarations; unknown
  struct names ignored; an undeclared struct keeping `i8*`; an extracted member
  flowing into a store and a call; nested position lists.
- `tests/mlir/test_struct_pointees.py` — 12 tests over the derivation, the type
  walk (including the `byval` path and literal structs), and the exact attribute
  shape `MLIRToLLVM70::loadStructPointees` parses.

Every branch of the walk was mutation-tested — removing struct-member recursion,
ArrayAttr descent, DictAttr descent, function-signature recursion, the
module-scope filter, or the bitfield skip each breaks at least one test. Bitfield
structs must be skipped because their members collapse to one storage member, so
an index would refine the wrong one.

The lit test was mutation-checked against the C++ change: with it stashed and
rebuilt, `struct_pointees.mlir` is the only file in the llvm70 suite whose status
changes.

## Verification

- Full Python suite: **PLACEHOLDER_SUITE**.
- llvm70 lit suite: no change in pass/fail set versus a stashed rebuild.
- `ruff format` / `ruff check` clean; C++ within 80 columns.

## Notes for reviewers

- `_pointee_types` calls `to_mlir_type` on the pointee, which for an
  `AggregateType` pointee will `new_identified` the struct if it does not exist
  yet. Harmless — `get_identified` is idempotent and in practice it always
  exists — but it is a type-creating side effect on a path that reads like a pure
  query.
- The `to_mlir_type` call is wrapped in `try`/`except`. Nothing else converts the
  pointee — the member itself is a bare `!llvm.ptr` — so an exotic one must not
  fail a compile that previously worked. Omitting it just keeps `i8*`.
- Testing caveat: this dev environment has no `lit` or `FileCheck`, so the llvm70
  suite was run through a minimal RUN-line runner with the system `FileCheck-18`.
  The `--dump-ptx` RUN lines fail here regardless of this branch (libnvvm 13.x
  rejects the emitted debug metadata version), so CI exercises a half of that
  suite I could not.


src/numba_cuda_mlir/struct_pointees.py
```diff
@@ -0,0 +1,146 @@
...
+
+    if fe_type.is_bitfield_struct:
+        return {}
+
+    pointees = {}
+    for i, (_, field_type, *_) in enumerate(fe_type.fields):
+        if not isinstance(field_type, types.CPointer):
+            continue
+        try:
+            pointees[i] = to_mlir_type(field_type.dtype)
```
atmnp: Shouldn't we use the storage type `to_mlir_storage_type()`? Using the value type here will still lead us to miss some inlining opportunities. + test cases.

atmnp (changes requested):

src/numba_cuda_mlir/struct_pointees.py
ZzEeKkAa: Good catch! There was a bug in `to_mlir_storage_type` in name identifier duplication that prevented it's usage. Fixed now

atmnp:
@coderabbit review

ZzEeKkAa:
@CodeRabbit review

atmnp (approved):
LGTM! We still miss some cases, but you don't have to address that here if you prefer. 

`struct_pointees_attr()` starts from identified structs visible in the Python-generated MLIR. Because `CPointer(Inner)` lowers to opaque `!llvm.ptr`, `Inner` is hidden from that initial traversal and its pointer-member metadata is never emitted. LLVM70 consequently emits members such as `Inner::data` as `i8*`, while linked NVRTC code uses the original typed pointer, such as `float*`.

This remains ABI-compatible, but the differing identified struct types prevent LTO from inlining the linked device function. We could process the discovered struct names using a worklist, recursively enqueueing identified structs found in recovered pointee types, reach a fixed point, then emit metadata for the complete set.

ZzEeKkAa:
> LGTM! We still miss some cases, but you don't have to address that here if you prefer.
> 
> `struct_pointees_attr()` starts from identified structs visible in the Python-generated MLIR. Because `CPointer(Inner)` lowers to opaque `!llvm.ptr`, `Inner` is hidden from that initial traversal and its pointer-member metadata is never emitted. LLVM70 consequently emits members such as `Inner::data` as `i8*`, while linked NVRTC code uses the original typed pointer, such as `float*`.
> 
> This remains ABI-compatible, but the differing identified struct types prevent LTO from inlining the linked device function. We could process the discovered struct names using a worklist, recursively enqueueing identified structs found in recovered pointee types, reach a fixed point, then emit metadata for the complete set.

That's another corner case that is related to structs  I was not aware of. Will target separately (probably with struct refactoring).

PR #283, opened 2026-08-27
Fix literally

### LINKED ISSUE #282: [BUG] an `@overload` that *reads* a `Literal` argument is not found
# BUG: an `@overload` that *reads* a `Literal` argument is not found

**Component:** `numba_cuda_mlir.descriptor` (MLIR overload lowering)
**Version:** numba-cuda-mlir 0.5.0, cuda-bindings 13.3.1, Python 3.11.12
**Device:** NVIDIA RTX 6000 Ada Generation, driver 595.45.04
**Related:** https://github.com/NVIDIA/numba-cuda-mlir/pull/256

## Summary

https://github.com/NVIDIA/numba-cuda-mlir/pull/256 added one line to `get_overload_builder`:

```diff
  match_args = (sig.recvr, *sig.args) if sig.recvr else sig.args
+ match_args = tuple(types.unliteral(arg) for arg in match_args)
```

That normalises the *signature* side, which is enough when the template was
typed with unliteral'd arguments.

It is not enough when the template is *literal-aware*. An overload registered
`prefer_literal=True`, or one that requests the constant with `literally()`, is
typed with the literal, so its `_impl_cache` is keyed on the literal — and now
it is the cache side that fails to match.

The two cases are mirror images, which is why fixing one did not fix the other.

The test PR #256 added, `test_extending_overload_with_literal_argument`,
registers its overload without `prefer_literal` and never reads the argument's
value, so it exercises that direction and only that one.

## Reproducing

```
$ python repro.py
no literal (control)     lowered ok -> 1.0
prefer_literal=True      FAILED -> NotImplementedError: signature: (Literal[int](32),) -> float64
literally()              FAILED -> NotImplementedError: signature: (Literal[int](32),) -> float64
```

All three call sites pass the same constant to the same shape of overload. Only
the two whose typing hook *reads* the value fail.

## Cause

`MLIRTypingContext.get_overload_builder` (`descriptor.py:1228`) unliterals the
signature, and only the signature:

```python
match_args = (sig.recvr, *sig.args) if sig.recvr else sig.args
match_args = tuple(types.unliteral(arg) for arg in match_args)
```

For a literal-aware template the cache retains the literal:

```
sig.args   = (Literal[int](32),)   ->  match_args (int64,)
cache args = (Literal[int](32),)
```

Neither branch of the `or` matches, `get_overload_builder` returns `None`, and
`lower_call_expr_assign` falls through to its final `else`.

## `literally()` is not a workaround

The retry itself works: the hook is called twice, first with `int64` and then
with `Literal[int](32)`. But the resulting signature carries the literal either
way, so it meets the same comparison. Any route that gets a constant to a typing
hook produces a literal signature, so `prefer_literal` and `literally` cannot
substitute for each other here.

Unliteralling inside the hook — #256  workaround — costs the constant, which
is the only reason a literal-aware overload exists.

## Suggested fix

Compare against the literal form as well, rather than unliteralling one side:

```diff
-        match_args = (sig.recvr, *sig.args) if sig.recvr else sig.args
-        match_args = tuple(types.unliteral(arg) for arg in match_args)
+        literal_args = tuple((sig.recvr, *sig.args) if sig.recvr else sig.args)
+        match_args = tuple(types.unliteral(arg) for arg in literal_args)
```

```diff
-                if cache_args == match_args or non_omitted_cache_args == non_omitted_match_args:
+                if (
+                    cache_args == match_args
+                    or cache_args == literal_args
+                    or non_omitted_cache_args == non_omitted_match_args
+                ):
```

Keeping both forms covers #256 direction and this one -- normalising only
one side cannot, because the two cases disagree about which side holds the
literal. With it applied:

```
no literal (control)     lowered ok -> 1.0
prefer_literal=True      lowered ok -> 32.0
literally()              lowered ok -> 32.0
```

A test alongside PR #256's would pin it: the same
`test_extending_overload_with_literal_argument`, registered
`prefer_literal=True` with a hook that reads `x.literal_value`.

## Why it matters

Reading a compile-time constant in a typing hook is how an extension
specialises a call: choosing a tile size or an unroll factor, or emitting a
different device function per constant. Today such an overload type-checks and
then fails at lowering, and the error names only the lowering step, so it reads
as "this call is unsupported" rather than "the builder lookup missed".

`repro.py`:

```python
"""An `@overload` reading a literal argument types, then fails to lower.

    python literal_arguments.py

Needs numba-cuda-mlir and a CUDA device. No other dependencies.
"""

import os

# Set before importing numba_cuda_mlir: without it the compiler reports an
# internal error and hides the cause, which here is the whole point.
os.environ.setdefault("NUMBA_CUDA_MLIR_ICE_FULL_TB", "1")

import numpy as np
from numba_cuda_mlir import types as nbt
from numba_cuda_mlir.extending import overload, typing_registry
from numba_cuda_mlir.numba_cuda.misc.special import literally

THREADS = 32


def plain(n):
    """Control: the hook never looks at the value."""
    raise NotImplementedError


@overload(plain, typing_registry=typing_registry)
def _plain(n):
    return lambda n: 1.0


def preferred(n):
    """The hook needs the constant, asked for with `prefer_literal`."""
    raise NotImplementedError


@overload(preferred, typing_registry=typing_registry, prefer_literal=True)
def _preferred(n):
    value = float(n.literal_value)
    return lambda n: value


def forced(n):
    """The same constant, demanded with `literally` instead."""
    raise NotImplementedError


@overload(forced, typing_registry=typing_registry)
def _forced(n):
    if not isinstance(n, nbt.IntegerLiteral):
        return lambda n: literally(n)
    value = float(n.literal_value)
    return lambda n: value


@cuda.jit
def k_plain(out):
    out[0] = plain(THREADS)


@cuda.jit
def k_preferred(out):
    out[0] = preferred(THREADS)


@cuda.jit
def k_forced(out):
    out[0] = forced(THREADS)


def run(name, kernel):
    try:
        kernel[1, 1](out)
        print(f"{name:24} lowered ok -> {out.copy_to_host()[0]}")
    except Exception as exc:
        lines = [ln.strip() for ln in str(exc).splitlines() if ln.strip()]
        why = next(
            (ln for ln in lines if ln.startswith("signature:")), lines[-1]
        )
        print(f"{name:24} FAILED -> {type(exc).__name__}: {why}")


run("no literal (control)", k_plain)
run("prefer_literal=True", k_preferred)
run("literally()", k_forced)
```

ZzEeKkAa:
Fixes #282


atmnp:
@coderabbit review

atmnp (approved):
LGTM assuming merge conflicts are resolved.

ZzEeKkAa:
@CodeRabbit review

PR #284, opened 2026-08-27
Serialize MLIR target initialization

atmnp:
Prevents cold-start races where concurrent threads observe partial registry
state during MLIR target initialization.

**Changes:**
- `descriptor.py` — serialize `ensure_initialized` and `refresh_registries`
  behind an `RLock`; concurrent callers wait rather than racing on registry
  population
- `numba_cuda/descriptor.py` — serialize typing-context refresh under the
  target context lock; add test covering concurrent initialization and
  declaration application
- `_threading.py` (new) — `_LockedCounter` wrapper for explicit thread-safe
  counting without relying on the GIL
- `numba_cuda/core/base.py`, `typing/context.py`, `typing/templates.py`,
  `numba_cuda/descriptor.py` — add `_registry_lock` guards around registry
  reads/writes; use `dict.setdefault` for `RegistryLoader` creation so
  concurrent loaders are deduplicated atomically rather than via try/except
- `numba_cuda/core/event.py` — snapshot listener list before broadcasting so
  mid-broadcast unregistration is safe
- `numba_cuda/types/abstract.py` — make type interning atomic
- `test_free_threading.py` (new) — unit tests for call-stack thread-locality,
  event broadcast safety, and type-interning atomicity
- `test_shared_iterators.py` (new) — tests for concurrent target
  initialization, declaration-application serialization, and registry refresh

### DIFF
--- a/src/numba_cuda_mlir/lowering_utilities/llvm_utils.py
+++ b/src/numba_cuda_mlir/lowering_utilities/llvm_utils.py
@@ -1,9 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 import ctypes
-import itertools
 import os
 import platform
+
+from numba_cuda_mlir._threading import _LockedCounter
 
 if platform.machine() in ("ARM64", "AMD64"):
     NVPTX64_DATALAYOUT = "e-p:64:64:64-p6:32:32:32-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-i128:128:128-f32:32:32-f64:64:64-f128:128:128-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64-a:8:8"
@@ -47,7 +48,7 @@ MODERN_TO_NVVM_BRIDGE_LIB_PATH = _find_modern_to_nvvm_bridge_lib()
 
 _capi = None
 _modern_to_nvvm_bridge = None
-_modern_bridge_dump_counter = itertools.count()
+_modern_bridge_dump_counter = _LockedCounter()
 
 
 def _load_capi_library(path: str, label: str):
--- a/src/numba_cuda_mlir/numba_cuda/core/base.py
+++ b/src/numba_cuda_mlir/numba_cuda/core/base.py
@@ -403,11 +403,7 @@ class BaseContext:
             ):
                 return
 
-            try:
-                loader = self._registries[registry]
-            except KeyError:
-                loader = RegistryLoader(registry)
-                self._registries[registry] = loader
+            loader = self._registries.setdefault(registry, RegistryLoader(registry))
             self.insert_func_defn(loader.new_registrations("functions"))
             self._insert_getattr_defn(loader.new_registrations("getattrs"))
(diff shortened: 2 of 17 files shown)


brandon-b-miller (approved):

src/numba_cuda_mlir/numba_cuda/core/base.py
```diff
@@ -426,54 +431,58 @@ def install_external_registry(self, registry):
...
+                try:
+                    return sig and is_external(sig[0])
+                except (AttributeError, IndexError):
+                    return True
+
+            try:
+                loader = self._registries[registry]
+            except KeyError:
+                loader = RegistryLoader(registry)
+                self._registries[registry] = loader
```
seberg: I am not 100% sure, but this looks a bit like you are _not_ strictly ensuring that this function doesn't get called multiple times, just that `loader` doesn't get replaced.

In that case, you don't need a lock, rather use `.setdefault`:

```suggestion
                loader = self._registries.setdefault(registry, loader)
```
the result may be an already set `loader` from a differnt thread, but if the `new_registrations` calls are no-ops for that (and to me it seems like that is already the case also here) then the whole rest of the diff (and the explicit lock) just goes away.

src/numba_cuda_mlir/lowering_utilities/llvm_utils.py
```diff
@@ -47,7 +48,7 @@ def _find_modern_to_nvvm_bridge_lib():
 
 _capi = None
 _modern_to_nvvm_bridge = None
-_modern_bridge_dump_counter = itertools.count()
+_modern_bridge_dump_counter = _LockedCounter()
```
seberg, line 51: Hmmm, did you actually observe this failing?  `itertools.count()` has code to be thread-safe, so it _should_ be fine and if not that would be a CPython bug.

VijayKandiah (approved):

src/numba_cuda_mlir/lowering_utilities/llvm_utils.py
atmnp: This was defensive for Python 3.13 where I am unsure if it's guaranteed, will leave it up to you and @leofang to make a call on whether this should be undone right now (as people who would know more).

src/numba_cuda_mlir/numba_cuda/core/base.py
atmnp: Good catch! I believe I addressed this

src/numba_cuda_mlir/lowering_utilities/llvm_utils.py
seberg: I just checked and the fix was included in 3.13.  I think you should assume `itertools.count()` is fine if that keeps the code a bit nicer.

leofang: Would there be a problem for 3.10-3.12?

leofang (approved):

src/numba_cuda_mlir/lowering_utilities/llvm_utils.py
seberg: There is just no problem here, I think.  I can't tell immediately from the code if for counts `>MAX_SSIZE_T` it might not be as safe as the code looks, but you wouldn't reach that anyway (and the code author seems to think it is)...

PR #291, opened 2026-09-02
Repair statement bodies emptied by AST transforms

ccam80:
# Repair statement bodies emptied by AST transforms

## Summary

A transform that removed the only statement in a body left that body empty, and `compile()` rejected the function. The pipeline now ends with a pass that fills each emptied body with `pass`.

## The problem

Four transforms delete statements outright: unrolling a loop over an empty iterable, folding a constant `if` with no branch to keep, removing a `with consteval():` block, and removing a bare `None` statement. Each returns an empty list in place of the statement. When nothing else shares the body, the enclosing function, loop, `if`, or `with` is left with no statements, and Python's compiler refuses the tree:

```python
@cuda.jit
def kernel(arr):
    for j in consteval(range(0)):
        arr[j] = 1.0
```

unrolled to

```python
def kernel(arr):
```

and compilation failed with

```
ValueError: empty body on FunctionDef
```

The same happens one level down: `if consteval(False):` as the only statement of a loop body gives `ValueError: empty body on For`. A zero-trip loop already compiled when another statement shared its body, so the failure depended on what surrounded the removed statement, not on which transform removed it.

## The fix

`EmptyBodyRepairPass` runs last in `create_default_pipeline`. It walks the tree and inserts a single `pass` into any statement body that is an empty list, so the example becomes

```python
def kernel(arr):
    pass
```

A `body` list is the only child `compile()` requires to be non-empty; an empty `orelse` is legal, so the pass leaves it alone. It reports no modification when it finds nothing to fill, so it never triggers a recompile by itself.

## Tests

Seven in `tests/ast_transforms/test_empty_body_repair.py`, asserting on `transformed_source` and on results:

- `test_repair_inserts_pass_into_empty_bodies` and `test_repair_leaves_populated_bodies_alone`: the repair function on its own
- `test_zero_trip_loop_as_only_statement`: a kernel body emptied by unrolling compiles to `pass`
- `test_false_if_as_only_statement_in_loop`: a loop body emptied by constant-if folding
- `test_consteval_block_as_only_statement_in_if`: an `if` body emptied by block removal
- `test_none_statement_as_only_statement_in_if`: an `if` body emptied by `None` removal
- `test_repair_runs_after_all_passes`: unrolling empties an `if`, folding then empties the kernel, one `pass` remains

Full suite: 4521 passed, 177 skipped, 300 xfailed, 0 failed.

### DIFF
--- a/src/numba_cuda_mlir/ast_transforms/__init__.py
+++ b/src/numba_cuda_mlir/ast_transforms/__init__.py
@@ -9,6 +9,7 @@ from numba_cuda_mlir.ast_transforms.common import get_function_ast, recompile_fu
 from numba_cuda_mlir.ast_transforms.comprehension import ComprehensionPass
 from numba_cuda_mlir.ast_transforms.consteval import ConstevalError, ConstevalPass
 from numba_cuda_mlir.ast_transforms.constant_if import ConstantIfPass
+from numba_cuda_mlir.ast_transforms.empty_body import EmptyBodyRepairPass
 from numba_cuda_mlir.ast_transforms.pipeline import (
     ASTTransformPass,
     ASTTransformPipeline,
@@ -63,6 +64,7 @@ def create_default_pipeline() -> ASTTransformPipeline:
     pipeline.add_pass(ConstantIfPass())
     pipeline.add_pass(ComprehensionPass())
     pipeline.add_pass(NoneStatementRemovalPass())
+    pipeline.add_pass(EmptyBodyRepairPass())
     return pipeline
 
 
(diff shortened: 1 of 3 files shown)


atmnp:
@coderabbitai review

atmnp (changes requested):
Could you address the CodeRabbit comments? Otherwise, looks great!

ccam80:
I've added a visit_Module method that calls `NodeTransformer.generic_visit` so statements inside the module get repaired but empty modules don't get filled. The path seems unreachable from any user-facing code, you'd need to manually invoke the pass on `ast.parse("")` to trigger it, but it might pop up in dev work. 

CodeRabbit's "at minimum" suggestion of adding an Interactive case isn't reachable - an empty `Interactive` only exists if you build it by hand, like `repair_empty_bodies(ast.Interactive(body=[]))`, or if you parse a statement in "single" mode then delete that statement in a separate pass and manually feed it into this pass. I can add a one line `visit_Interactive = visit_Module` below `visit_Module` if you want to support it, but it looked like unnecessary clutter to me so I left it out of this commit.

PR #292, opened 2026-09-03
Apply AST transforms to inlined device functions

ccam80:
# Apply AST transforms to inlined device functions

## Summary

A device function declared `inline=True` was spliced into its caller from its original bytecode, so `consteval` inside it was never transformed. Inlined callees now go through the AST transform pipeline before the inliner builds their Numba IR, the same way a kernel does.

numba/numba#5673 introduced `InlineWorker` so that every inlinee runs the full untyped pipeline and is transformed as though it were compiled directly. This repository runs AST transforms ahead of that pipeline, and this change extends the same rule to them.

## The problem

Kernels and non-inlined device functions are transformed in `compile_mlir`, before the compiler front end sees them. `InlineInlinables` takes a different route: it reads the callee's `py_func` and runs the untyped pipeline on its bytecode directly. `apply_ast_transforms` never ran on that function:

```python
n = 4

@cuda.jit(device=True, inline=True)
def fill(out, base):
    for i in consteval(range(n)):
        out[i] = base + i

@cuda.jit
def kernel(out):
    fill(out, 10)
```

failed in type inference:

```
TypingError: Untyped global name 'consteval': Cannot determine Numba type of <class 'function'>
    for i in consteval(range(n)):
```

Every `consteval` form inside an inlined callee failed this way: loops, expressions, `with consteval():` blocks and `current_target_options()`.

## The fix

`InlineWorker` gains two optional arguments: `targetoptions`, the caller's jit options, and `inlinee_transform`, a callback `(function, callee_targetoptions, targetoptions) -> function`. `InlineInlinables` reads both from `state.metadata` and calls the callback through `inline_worker.transform_inlinee` inside `if do_inline`, immediately before `inline_function`, so a callable cost model still evaluates the original function and a rejected inline is never transformed. The worker writes both values into the metadata of every callee state it builds, so a callee inlined inside another inlined callee is transformed the same way.

The MLIR compiler registers the callback, `ast_transforms.transform_inline_callee`, in `state.metadata` next to `targetoptions`; the generic pass carries no AST policy. The callback runs `apply_ast_transforms` with the caller's options, taking only `experimental_ast_transforms`, `dump_ast` and `dump_ast_after_all` from the callee's own decorator. Inside a transformed callee `current_target_options()` is therefore the calling kernel's option set. Every parameter name resolves to a placeholder object rather than a type, so a `consteval` expression that uses one raises `ConstevalError` instead of reading a same-named global.

`recompile_function` now shifts the dedented AST to the function's `co_firstlineno` before compiling it, so any transformed function's code object reports the lines of its defining file. This applies to kernels as well as inlined callees.

## Tests

Seven added in `tests/ast_transforms/test_inlined_callees.py`:

- `test_inlined_callee_consteval_loop`: a `consteval(range(n))` loop in an inlined callee unrolls and runs
- `test_nested_inlined_callees_transform`: a callee inlined inside another inlined callee is transformed
- `test_inlined_callee_sees_caller_target_options`: `current_target_options()` in nested callees reads the kernel's `chip`
- `test_inlined_callee_parameter_types_unavailable`: a parameter attribute in `consteval` raises `ConstevalError`
- `test_inlined_callee_parameter_shadows_global`: a parameter named like a module global raises `ConstevalError` rather than reading the global
- `test_callee_option_controls_transform`: a callee with `experimental_ast_transforms=False` is left alone and fails with the `TypingError` above
- `test_rejected_cost_model_does_not_transform`: a cost model returning `False` never invokes the transform

One added in `tests/ast_transforms/test_consteval.py`:

- `test_recompiled_function_keeps_source_lines`: a recompiled function keeps `co_firstlineno` and the original line of its transformed statement

Full suite: 4522 passed, 177 skipped, 300 xfailed, 0 failed. Six of the seven callee tests and the line-number test fail on `main`; `test_callee_option_controls_transform` passes there by construction.

## Summary by CodeRabbit

* **New Features**
  * Added support for applying AST transformations to inlined device functions, including nested inlining.
  * Inlined functions now inherit the calling kernel’s target options while respecting their own transformation settings.
  * Added support for compile-time evaluation within inlined callees.

* **Bug Fixes**
  * Preserved original source line information for transformed and recompiled functions.

* **Documentation**
  * Clarified compile-time evaluation behavior for inline device functions.

### DIFF
--- a/src/numba_cuda_mlir/ast_transforms/__init__.py
+++ b/src/numba_cuda_mlir/ast_transforms/__init__.py
@@ -3,6 +3,7 @@
 # AST transformation passes for numba_cuda_mlir
 # These run before Numba's IR conversion
 import ast
+import inspect
 from typing import Callable
 
 from numba_cuda_mlir.ast_transforms.common import get_function_ast, recompile_function
@@ -17,6 +18,7 @@ from numba_cuda_mlir.ast_transforms.pipeline import (
 
 __all__ = [
     "apply_ast_transforms",
+    "transform_inline_callee",
     "ConstevalError",
     "ASTTransformPass",
     "ASTTransformPipeline",
@@ -127,3 +129,26 @@ def apply_ast_transforms(
         func = recompile_function(func, tree, context.stored_values)
 
     return func, transformed_source
+
+
+# Options taken from the callee's decorator; all others come from the caller.
+_CALLEE_AST_OPTIONS = ("experimental_ast_transforms", "dump_ast", "dump_ast_after_all")
+# Stands in for each parameter of an inlined callee, whose types are unknown.
+_INLINEE_PARAMETER = object()
+
+
+def transform_inline_callee(
+    pyfunc: Callable, callee_targetoptions: dict, caller_targetoptions: dict
+) -> Callable:
+    """Apply the callee's AST transforms under the caller's options; parameters resolve to a placeholder."""
+    if not callee_targetoptions.get("experimental_ast_transforms", False):
+        return pyfunc
+
+    targetoptions = dict(caller_targetoptions)
+    for name in _CALLEE_AST_OPTIONS:
+        if name in callee_targetoptions:
+            targetoptions[name] = callee_targetoptions[name]
+
+    argtypes = (_INLINEE_PARAMETER,) * len(inspect.signature(pyfunc).parameters)
+    transformed, _ = apply_ast_transforms(pyfunc, targetoptions, argtypes)
+    return transformed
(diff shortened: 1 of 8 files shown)


atmnp:
@coderabbitai review

atmnp (changes requested):
It looks like transformed inlinees lose keyword-only defaults. Can we set `new_func.__kwdefaults__ = func.__kwdefaults__` and add a inline-callee regression with omitted keyword-only default.

Otherwise LGTM!

ccam80:
@atmnp keyword-only default preservation added in 87df3b7, and a regression added alongside it, but the test is `xfail` as `fold_arguments` rejects keyword-only args. I'll submit a followup PR that fixes removes the guard and un-xfails the test for device functions - it looks like there's a bit more design work required to support it for kernels through the launcher so I'll keep it confined to device functions.

PR #293, opened 2026-09-03
Compute CFGraph topological order without recursion

### LINKED ISSUE #5611: Control flow graph algorithm error when using many ternary operations
## Reporting a bug

- [x] I am using the latest released version of Numba (most recent is visible in
 the change log (https://github.com/numba/numba/blob/master/CHANGE_LOG).
- [x] I have included below a minimal working reproducer (if you are unsure how
 to write one see http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports).

```python
from numba import njit
import numpy as np


def generate_code(count):
    base_code = """
    dependency_{index} = 1 if True else 0
    value_{index} = x if dependendency_{index} else 0"""

    code = """
@njit
def template(x):"""
    for i in range(count):
        code += base_code.format(**{'index': i})
    return code

if __name__ == '__main__':
    import sys
    import time

    count = int(sys.argv[1])

    exec(generate_code(count), globals(), locals())
    start = time.time()
    template(32)
    end = time.time()
    print(end - start, 'elapsed')
```
When running the above code with count > 200, the growth in compilation time appears to grow exponentially based on the control flow graph search.  @stuartarchibald suggested that the recursive algorithm would likely need to be replaced with an iterative one as "the amount of time it takes to analyse the CFG would be to some degree proportional to the number of [if] blocks."  When run with sufficiently large count > 10000, the following error is produced:

```python
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 420, in _compile_for_args
    raise e
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 353, in _compile_for_args
    return self.compile(tuple(argtypes))
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_lock.py", line 32, in _acquire_compile_lock
    return func(*args, **kwargs)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 794, in compile
    cres = self._compiler.compile(args, return_type)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 77, in compile
    status, retval = self._compile_cached(args, return_type)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 91, in _compile_cached
    retval = self._compile_core(args, return_type)
  File "/***/.local/lib/python3.8/site-packages/numba/core/dispatcher.py", line 104, in _compile_core
    cres = compiler.compile_extra(self.targetdescr.typing_context,
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler.py", line 568, in compile_extra
    return pipeline.compile_extra(func)
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler.py", line 339, in compile_extra
    return self._compile_bytecode()
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler.py", line 401, in _compile_bytecode
    return self._compile_core()
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler.py", line 381, in _compile_core
    raise e
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler.py", line 372, in _compile_core
    pm.run(self.state)
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_machinery.py", line 341, in run
    raise patched_exception
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_machinery.py", line 332, in run
    self._runPass(idx, pass_inst, state)
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_lock.py", line 32, in _acquire_compile_lock
    return func(*args, **kwargs)
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_machinery.py", line 291, in _runPass
    mutated |= check(pss.run_pass, internal_state)
  File "/***/.local/lib/python3.8/site-packages/numba/core/compiler_machinery.py", line 264, in check
    mangled = func(compiler_state)
  File "/***/.local/lib/python3.8/site-packages/numba/core/untyped_passes.py", line 86, in run_pass
    func_ir = interp.interpret(bc)
  File "/***/.local/lib/python3.8/site-packages/numba/core/interpreter.py", line 116, in interpret
    flow.run()
  File "/***/.local/lib/python3.8/site-packages/numba/core/byteflow.py", line 141, in run
    self._build_cfg(runner.finished)
  File "/***/.local/lib/python3.8/site-packages/numba/core/byteflow.py", line 158, in _build_cfg
    graph.process()
  File "/***/.local/lib/python3.8/site-packages/numba/core/controlflow.py", line 125, in process
    self._find_topo_order()
  File "/***/.local/lib/python3.8/site-packages/numba/core/controlflow.py", line 537, in _find_topo_order
    _dfs_rec(self._entry_point)
  File "/***/.local/lib/python3.8/site-packages/numba/core/controlflow.py", line 534, in _dfs_rec
    _dfs_rec(dest)
  File "/***/.local/lib/python3.8/site-packages/numba/core/controlflow.py", line 534, in _dfs_rec
    _dfs_rec(dest)
  File "/***/.local/lib/python3.8/site-packages/numba/core/controlflow.py", line 534, in _dfs_rec
    _dfs_rec(dest)
  [Previous line repeated 970 more times]
  File "/***/.local/lib/python3.8/site-packages/numba/core/controlflow.py", line 531, in _dfs_rec
    seen.add(node)
RecursionError: Failed in nopython mode pipeline (step: analyzing bytecode)
maximum recursion depth exceeded while calling a Python object
```

ccam80:
# Compute `CFGraph` topological order without recursion

## Summary

`CFGraph._find_topo_order` recursed once per basic block along the longest path through the control-flow graph. Kernels with more than about 1000 sequential blocks failed to compile with `RecursionError`. The traversal now runs on an explicit stack and produces the same order as before.

## The problem

Every kernel's CFG is topologically ordered during with-lifting. The order was computed by a recursive depth-first search, so its Python stack depth equalled the length of the longest forward path through the graph. A straight-line kernel of `if` blocks is one long path:

```python
lines = ["def kernel(arr, out):", "    i = cuda.grid(1)"]
for k in range(1500):
    lines.append(f"    if arr[i] > {k}.0:")
    lines.append("        out[i] += 1.0")
exec("\n".join(lines))
cuda.jit("void(float32[:], float32[:])")(kernel)
```

```
  File ".../numba_cuda/core/controlflow.py", line 649, in _dfs_rec
    _dfs_rec(dest)
  [Previous line repeated 975 more times]
RecursionError: maximum recursion depth exceeded
```

Kernels produced by `consteval` unrolling reach this size routinely.

## The fix

`_find_topo_order` runs the same depth-first search on an explicit stack of `(callback, node)` pairs, the shape `_find_postorder` already uses. Successors are pushed in reverse so nodes are visited in the same order as the recursive search, and `topo_order()` returns the same list as before for every graph that compiled.

numba/numba#10482, which fixes numba/numba#5611, removes the same recursion. Its traversal marks a node seen when it is pushed rather than when it is visited, which changes the emitted order and can place a node before one of its predecessors: on the graph `0 -> 1, 0 -> 2, 2 -> 1` it yields `[0, 1, 2]`. This change marks at visit time, so the order is identical to the recursive search and the existing `test_topo_order` assertions are unchanged.

## Tests

One added to `tests/numba_cuda_tests/cudapy/test_flow_control.py`:

- `test_topo_order_deep_graph`: a chain of 2000 diamonds (4001 nodes) is ordered under a recursion limit 200 frames above the test's own depth, with every node present and every edge pointing forward

`test_flow_control.py`: 61 passed, 4 xfailed. The generated 1500-block kernel above compiles and runs. On 500 random CFGs with loops the new order matches the recursive order exactly.

Full suite: 4515 passed, 177 skipped, 300 xfailed, 0 failed.

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/core/controlflow.py
+++ b/src/numba_cuda_mlir/numba_cuda/core/controlflow.py
@@ -641,15 +641,19 @@ class CFGraph:
         post_order = []
         seen = set()
 
-        def _dfs_rec(node):
+        # Successors pushed in reverse so the stack visits them in iteration order.
+        def visit(node):
             if node not in seen:
                 seen.add(node)
-                for dest in succs[node]:
-                    if (node, dest) not in back_edges:
-                        _dfs_rec(dest)
-                post_order.append(node)
+                stack.append((post_order.append, node))
+                forward = [dest for dest in succs[node] if (node, dest) not in back_edges]
+                stack.extend((visit, dest) for dest in reversed(forward))
 
-        _dfs_rec(self._entry_point)
+        stack = [(visit, self._entry_point)]
+        while stack:
+            cb, node = stack.pop()
+            cb(node)
+
         post_order.reverse()
         return post_order
 
--- a/tests/numba_cuda_tests/cudapy/test_flow_control.py
+++ b/tests/numba_cuda_tests/cudapy/test_flow_control.py
@@ -2,7 +2,9 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import pytest
+import inspect
 import itertools
+import sys
 
 import unittest
 from numba_cuda_mlir import cuda
@@ -630,6 +632,33 @@ class TestCFGraph(NumbaCUDATestCase):
                 [0, 10, 13, 26, 19, 6],
             ),
         )
+
+    def test_topo_order_deep_graph(self):
+        # Diamond chain longer than the recursion limit.
+        n_diamonds = 2000
+        adj = {}
+        for i in range(n_diamonds):
+            head, body, join = 2 * i, 2 * i + 1, 2 * i + 2
+            adj[head] = [body, join]
+            adj[body] = [join]
+        adj[2 * n_diamonds] = []
+        g = self.from_adj_list(adj)
+        g.set_entry_point(0)
+        g.process()
+
+        current_depth = len(inspect.stack())
+        old_limit = sys.getrecursionlimit()
+        sys.setrecursionlimit(current_depth + 200)
+        try:
+            order = g.topo_order()
+        finally:
+            sys.setrecursionlimit(old_limit)
+
+        self.assertEqual(sorted(order), sorted(adj))
+        position = {node: i for i, node in enumerate(order)}
+        for src, dests in adj.items():
+            for dest in dests:
+                self.assertLess(position[src], position[dest])
 
     def test_topo_sort(self):
         def check_topo_sort(nodes, expected):


atmnp:
@coderabbitai review

brandon-b-miller (changes requested):
I'm +1 on this change. While we're touching this, I think up in topo_sort we should:

```python
nodes = set(nodes)
if not nodes:
    return
it = self._topo_order
```

As soon as this generator is iterated it computes `self._topo_order`. Theres no need to topo sort the whole CFG for kernels without a with. This might actually fix the case in the OP by itself (I haven't tested) but we should still make this change otherwise, plenty of real world cases that need the topo sort.

ccam80:
I've added the empty-set skip in 64b7c75. As @brandon-b-miller pointed out, the skip alone fixes the compile-time blowup in the repro and in my repo, so it makes the rest of this PR nice-to-have for future users of the CFG topo sort rather than a current-pipeline improvement.

PR #294, opened 2026-09-03
Preserve signedness in integer-float conversions

### LINKED ISSUE #290: [BUG] Unsigned integer/float conversions use signed instructions (silently wrong values)
**Describe the bug**

Converting an unsigned integer to a float lowers to `arith.sitofp`, and converting a float to an unsigned integer lowers to `arith.fptosi`. Values outside the signed range come out wrong, with no error.

```python
import numpy as np

@cuda.jit
def widen(src, dst):
    i = cuda.grid(1)
    if i < src.size:
        dst[i] = np.float64(src[i])

src = np.array([200], dtype=np.uint8)
dst = cuda.device_array(1, dtype=np.float64)
widen[1, 32](cuda.to_device(src), dst)

print(dst.copy_to_host()[0])   # -56.0, expected 200.0
```

The emitted IR for that kernel:

```
%3 = load i8, ptr %0, align 1
%4 = sitofp i8 %3 to double
```

Same results for a plain `dst[i] = src[i]` assignment as for the explicit cast:

| case | expected | numba-cuda | numba-cuda-mlir |
|---|--:|--:|--:|
| `uint8 -> float64`, 200 | 200.0 | 200.0 | -56.0 |
| `uint16 -> float64`, 60000 | 60000.0 | 60000.0 | -5536.0 |
| `uint32 -> float64`, 3e9 | 3e9 | 3e9 | -1.29e9 |
| `uint32 -> float32`, 3e9 | 3e9 | 3e9 | -1.29e9 |
| `float64 -> uint32`, 3e9 | 3000000000 | 3000000000 | 2147483647 |
| `int8 -> float64`, -56 | -56.0 | -56.0 | -56.0 |

**Expected behavior**

The values NumPy and numba-cuda produce, i.e. `uitofp` for an unsigned source and `fptoui` for an unsigned target.

**Possible source**

`lowering_utilities.convert` takes a `signed=` keyword and passes it to `_convert_integer_to_integer`, but the integer/float arms ignore it and branch on width instead, which only distinguishes `i1`:

```python
case ir.IntegerType(), ir.FloatType():
    return (
        arith.sitofp(out=target_type, in_=value)
        if value_type.width > 1
        else arith.uitofp(out=target_type, in_=value)
    )

case ir.FloatType(), ir.IntegerType():
    return (
        arith.fptosi(out=target_type, in_=value)
        if target_type.width > 1
        else arith.fptoui(out=target_type, in_=value)
    )
```

The vector arms in the same function use the same width test.

**Environment details**

- Environment location: Bare-metal, Linux
- Method of numba-cuda-mlir install: pip, `numba-cuda-mlir==0.5.1`
- numba-cuda 0.30.4, numba 0.65.1, CUDA 13.3, sm_89

atmnp:
Closes #290.

Select signed or unsigned floating-point conversion operations from frontend types during casts and array stores, while preserving the legacy fallback for conversions without frontend signedness metadata. Boolean-to-integer stores remain zero-extended.\n\nAdds a table-driven regression matrix for unsigned-to-float, float-to-unsigned, and signed control conversions.

### DIFF
--- a/src/numba_cuda_mlir/lowering/builtins.py
+++ b/src/numba_cuda_mlir/lowering/builtins.py
@@ -449,8 +449,14 @@ def lower_broadcasted_div(builder, target, args, kwargs):
 def type_convert(builder, target, args, kwargs):
     target_numba_ty = builder.get_numba_type(target)
     to_type = builder.get_mlir_type(target)
-    to_signed = isinstance(target_numba_ty, types.Integer) and target_numba_ty.signed
     source_numba_ty = builder.get_numba_type(args[0])
+    signed = (
+        source_numba_ty.signed
+        if isinstance(source_numba_ty, types.Integer)
+        else target_numba_ty.signed
+        if isinstance(target_numba_ty, types.Integer)
+        else False
+    )
 
     if isinstance(target_numba_ty, (types.IntegerLiteral, types.Literal)):
         result = constant(target_numba_ty.literal_value, to_type)
@@ -474,7 +480,7 @@ def type_convert(builder, target, args, kwargs):
         value = convert(
             value,
             to_type,
-            signed=to_signed and not isinstance(source_numba_ty, types.Boolean),
+            signed=signed and not isinstance(source_numba_ty, types.Boolean),
         )
     builder.store_var(target, value)
 
--- a/src/numba_cuda_mlir/lowering/builtins.py
+++ b/src/numba_cuda_mlir/lowering/builtins.py
@@ -473,7 +473,7 @@ def type_convert(builder, target, args, kwargs):
     if isinstance(value.type, ir.BF16Type) and isinstance(to_type, ir.IntegerType):
         value = (
             arith.fptosi(out=to_type, in_=value)
-            if to_type.width > 1
+            if signed
             else arith.fptoui(out=to_type, in_=value)
         )
     else:
--- a/src/numba_cuda_mlir/lowering/builtins.py
+++ b/src/numba_cuda_mlir/lowering/builtins.py
@@ -18,6 +18,7 @@ from numba_cuda_mlir.lowering_utilities import (
     tensor_to_memref,
     simple_scalar_conversion_op,
     expensive_coerce_tensor_type,
+    get_conversion_signedness,
     try_extract_constant,
 )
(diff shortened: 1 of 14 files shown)


atmnp:
@coderabbit review

atmnp:
@coderabbit review

atmnp:
@coderabbit review

src/numba_cuda_mlir/lowering/builtins.py
```diff
@@ -449,8 +449,14 @@ def casted_div(lhs: lhs_etype, rhs: rhs_etype, init: result_value_type):
...
     to_type = builder.get_mlir_type(target)
-    to_signed = isinstance(target_numba_ty, types.Integer) and target_numba_ty.signed
     source_numba_ty = builder.get_numba_type(args[0])
+    signed = (
+        source_numba_ty.signed
+        if isinstance(source_numba_ty, types.Integer)
+        else target_numba_ty.signed
+        if isinstance(target_numba_ty, types.Integer)
+        else False
+    )
```
brandon-b-miller: Switching things to work this way also appears to repair a bunch of other cases that are wrong on main and unreported in the issue. We should add tests that explicitly cover these cases. 

```
# left column is what main produces today; right is this branch, which matches numpy
np.int64(np.uint8(200))        # -56          ->  200
np.uint32(np.int8(-56))        #  200         ->  4294967240
np.int64(np.uint32(3e9))       # -1294967296  ->  3000000000
dst_i64[0] = np.int8(-56)      #  200         -> -56
dst_i64[0] = np.int32(-5)      #  4294967291  -> -5
```

brandon-b-miller (changes requested):
Couple of comments here.

1. Basically fuzzing this locally I think I managed to find a place where arithmetic lowering codepath is vulnerable to this same bug: https://github.com/NVIDIA/numba-cuda-mlir/blob/main/src/numba_cuda_mlir/lowering/math.py#L391-L393. 
2. The signedness calculation lives in a few different places. As part of addressing (1) I think we should extract this policy into a helper function.

atmnp:
@coderabbitai review

brandon-b-miller (approved):

PR #296, opened 2026-09-03
Compute CFG dominators from immediate dominators

ccam80:
# Compute CFG dominators from immediate dominators

## Summary

Dominator and post-dominator sets were built by a fixed point whose cost grew with the square of the block count, and the compiler asked for the full sets twice per kernel. They are now built from the immediate dominators the graph already computes, and the two callers that only needed one chain no longer build the sets at all. A kernel of 400 sequential conditionals compiles in 5.2 s instead of 12.4 s.

numba/numba#10494 (the compile-time half of numba/numba#5611) builds dominator sets from immediate dominators in the same way. This change also finds post-dominators without editing the graph, walks the immediate post-dominator chain in `backbone()`, and tests dominance in SSA from the dominator tree.

## The problem

Every block started with the full node set and the sets shrank until nothing changed. Each step took the common elements of the predecessors' sets and copied the result, so a block deep in a long chain was revisited once per block above it, copying a chain-length set each time. Kernels with many sequential conditionals have that shape:

```python
@cuda.jit
def kernel(out, x):
    a = x
    a = a + 1 if a > 0 else a - 1
    a = a + 2 if a > 1 else a - 2
    ...                                 # N steps, one diamond of blocks each
    out[0] = a
```

`backbone()` and `ssa._find_defs_violators` then requested the full sets, each for a question that needs one chain. Post-dominators also required adding a placeholder exit block, linking every block of an infinite loop to it, and removing both afterwards.

## The fix

`_immediate_dominators(roots, preds, succs)` runs Cooper-Harvey-Kennedy from a set of starting blocks under a shared placeholder root, so one routine serves the forward graph from the entry and the reversed graph from the exits. A starting block maps to itself; a block dominated only by the placeholder maps to `None`.

`_find_immediate_post_dominators` starts from the exit blocks and the blocks of loops that never exit, so the graph is never edited. With no exit and no such loop there is nothing to start from, and every block keeps the all-blocks set it had before.

`_dominator_sets(idom)` follows each block's chain of immediate dominators, reusing chains already built. Only `dominators()` and `post_dominators()` call it.

`backbone()` follows the immediate post-dominator chain from the entry. `ssa._find_defs_violators` numbers the dominator tree in visiting order and compares two numbers per use. `_find_back_edges` keeps the current path in a set beside the stack.

## Tests

`test_dominators_deep_graph` checks `backbone`, immediate dominators, dominators and post-dominators on a chain of 2000 diamonds. `TestSSAViolators` flags a single-definition variable exactly when a use is not dominated by the definition.

Every analysis was compared with the previous implementation on all 66,066 directed graphs of one to four blocks and on 3,000 random graphs of two to twelve blocks; all results are identical.

| | before | after |
|---|---|---|
| 400-step kernel, first compile | 12.4 s | 5.2 s |
| `backbone()`, 4001-block diamond chain | 0.608 s | 0.009 s |
| `dominators()`, 4001-block diamond chain | 0.792 s | 0.133 s |
| back edges, 9001 blocks | 0.068 s | 0.004 s |

Full suite: 4518 passed, 177 skipped, 300 xfailed, 0 failed.

## Notes for reviewers

Starting blocks map to themselves, as `_find_dominator_tree` and `_find_dominance_frontier` already expect of the entry, so `_dominator_sets` and `backbone()` stop on the same test in both directions. Each block's dominator set is still its own `set`; only the two loop-exit fix-ups request the post-dominator sets.

### DIFF
--- a/src/numba_cuda_mlir/numba_cuda/core/controlflow.py
+++ b/src/numba_cuda_mlir/numba_cuda/core/controlflow.py
@@ -241,6 +241,10 @@ class CFGraph:
         return self._find_immediate_dominators()
 
     @functools.cached_property
+    def _ipdom(self):
+        return self._find_immediate_post_dominators()
+
+    @functools.cached_property
     def _df(self):
         return self._find_dominance_frontier()
 
@@ -275,7 +279,16 @@ class CFGraph:
          must go through).  By construction, it is non-empty: it contains
          at least the entry point.
         """
-        return self._post_doms[self._entry_point]
+        ipdom = self._ipdom
+        node = self._entry_point
+        if node not in ipdom:
+            return set(self._nodes)
+        backbone = set()
+        while node is not None:
+            backbone.add(node)
+            parent = ipdom[node]
+            node = None if parent == node else parent
+        return backbone
 
     def loops(self):
         """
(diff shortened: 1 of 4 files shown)


atmnp:
@coderabbitai review

atmnp:
@coderabbitai review

atmnp:
@coderabbitai full review

atmnp (approved):

PR #298, opened 2026-09-03
Add cuda.unroll and cuda.nounroll loop unroll hints

ccam80:
# Add `cuda.unroll` and `cuda.nounroll` loop unroll hints

## Summary

`cuda.unroll(it)`, `cuda.unroll(it, n)` and `cuda.nounroll(it)` wrap the iterable of a `for` loop and ask libnvvm to unroll that loop fully, `n` times, or not at all. They do the same job as `#pragma unroll`, `#pragma unroll n` and `#pragma unroll 1` in CUDA C++. Any iterable a `for` loop lowers takes the hint.

## The problem

The only way to unroll a loop today is `consteval`, which expands it in the AST. The kernel grows by the body times the trip count before the front end sees it, so compile time grows with it (linearly once #296 and #297 land, faster before). Besides compile cost, `consteval` does not support partial unrolling or preventing libnvvm unrolling a rolled loop.

## The fix

`cuda.unroll` and `cuda.nounroll` pass their iterable through directly, typed as identity. `count` (positional or keyword) is how many copies of the body to unroll, and must be a compile-time integer. The lowering step attaches the hint to the iterable, which travels with it through assignments and into the loop's iterator. The hint is stored in a table by the loop's header, and every jump back to that header gets emitted as an `llvm.br` with the hint as its `loop_annotation`. The LLVM70 bridge turns the annotation into `!llvm.loop` metadata on the branch, which the modern bridge does already.

`docs/source/reference/loop_unroll.rst` documents `consteval` unrolling and the two hints.

## Tests

Nineteen tests in `tests/test_loop_unroll.py`, 22 cases:

- `test_unroll_full_annotates_latch`, `test_unroll_count_annotates_latch` (positional and keyword), `test_nounroll_annotates_latch`: each hint form lands as a `loop_annotation` on the loop's back edge
- `test_array_loop_is_annotated`, `test_tuple_loop_is_annotated`, `test_nditer_loop_is_annotated`: the same for the other iterables
- `test_hinted_iterable_held_in_a_variable`: `it = cuda.nounroll(x)` followed by `for v in it` keeps the hint
- `test_nounroll_keeps_loop_in_ptx`, `test_unroll_count_shapes_ptx`, `test_nounroll_survives_lto`: libnvvm honours the hint: one load in the PTX for `nounroll`, four for `count=4`, one after an LTO link
- `test_break_as_last_statement_keeps_hint`, `test_return_as_last_statement_keeps_hint`: a body whose last statement leaves the loop is unrolled four times and fully
- `test_debug_build_annotates_latch`: a `debug=True` build keeps the annotation and its LLVM IR carries the count node
- `test_break_and_continue_loop_is_annotated`, `test_nested_loops_carry_their_own_hints`, `test_latches_of_one_loop_share_its_llvm_loop_id`: `break`/`continue` in the body, nested loops with different hints, and one `!llvm.loop` node shared by a loop's back edges
- `test_runtime_count_is_rejected`, `test_out_of_range_count_is_rejected` (0 and 2**31), `test_non_iterable_is_rejected`: typing errors for a runtime count, an out-of-range count, and a scalar iterable
- `test_markers_return_the_iterable_outside_compiled_code`: the functions return their argument in plain Python

Full suite: 4536 passed, 177 skipped, 300 xfailed, 0 failed (sm_89, CUDA 13, Windows).

### DIFF
--- a/cext/mlir-llvm70/include/llvm70/LLVM70IRBuilder.h
+++ b/cext/mlir-llvm70/include/llvm70/LLVM70IRBuilder.h
@@ -211,6 +211,11 @@ public:
   LLVMValueRef mdString(const char *str, unsigned len);
   LLVMValueRef mdNode(LLVMValueRef *vals, unsigned count);
   void addNamedMetadataOperand(const char *name, LLVMValueRef val);
+  /// Build a loop-ID node: `!N = distinct !{!N, vals...}`.
+  LLVMValueRef selfReferentialMDNode(LLVMValueRef *vals, unsigned count);
+  /// Attach `node` to `inst` under metadata kind `kindName`.
+  void setInstructionMetadata(LLVMValueRef inst, const char *kindName,
+                              LLVMValueRef node);
 
   // --- Debug info ---
   void initDebugInfo();
@@ -478,6 +483,12 @@ private:
   LLVM_FN(LLVMValueRef, fnMDNode, LLVMContextRef, LLVMValueRef *, unsigned)
   LLVM_FN(void, fnAddNamedMetadataOperand, LLVMModuleRef, const char *,
            LLVMValueRef)
+  LLVM_FN(unsigned, fnGetMDKindIDInContext, LLVMContextRef, const char *,
+           unsigned)
+  LLVM_FN(void, fnSetMetadata, LLVMValueRef, unsigned, LLVMValueRef)
+  LLVM_FN(LLVMMetadataRef, fnValueAsMetadata, LLVMValueRef)
+  LLVM_FN(LLVMMetadataRef, fnTemporaryMDNode, LLVMContextRef,
+           LLVMMetadataRef *, size_t)
 
   // Debug info
   LLVMDIBuilderRef diBuilder = nullptr;
--- a/cext/mlir-llvm70/include/llvm70/LLVM70Target.h
+++ b/cext/mlir-llvm70/include/llvm70/LLVM70Target.h
@@ -90,6 +90,10 @@ private:
 
   llvm::StringMap<LLVMTypeRef> namedStructCache;
 
+  // `!llvm.loop` ID node by loop annotation.
+  llvm::DenseMap<mlir::Attribute, LLVMValueRef> loopMetadata;
+  LLVMValueRef loopMetadataFor(mlir::LLVM::LoopAnnotationAttr attr);
+
   // Debug info state
   LLVMMetadataRef diCompileUnit = nullptr;
   LLVMMetadataRef diSubroutineType = nullptr;
--- a/cext/mlir-llvm70/lib/LLVM70IRBuilder.cpp
+++ b/cext/mlir-llvm70/lib/LLVM70IRBuilder.cpp
@@ -7,6 +7,9 @@
 //===----------------------------------------------------------------------===//
 
 #include "llvm70/LLVM70IRBuilder.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include <cstring>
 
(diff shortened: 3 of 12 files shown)


atmnp:
Love the intention of this change! I can definitely see the advantage of having this escape hatch. Can you give a reproducer kernel from cubie in this discussion where you see this compile-time blow-up being a concern? I can construct a pathological example myself but I'd like to see what the problem looks like in-vivo before we commit to this API or choose another.

ccam80:
@atmnp Here's a repro. Took me a while to figure out how to present it without dragging you through cubie. Unroll hints cut compile time from 18s to 7s in the provided kernel, but we apply all of my open compile-time PRs on import - without #197, #297, #199 it blows up to 256s/27s. Unroll hints also add the ability for us to unroll warp-voted loops that feature a `break`, and force-roll a loop that libnvvm would otherwise unroll. Both of these have shown runtime improvements in some kernels.

Cubie kernels run huge straight-line math blocks inside nested, warp-voted early-exit loops. To make them configurable we break them down into many still-too-big device functions, and for runtime performance we set them `inline=True`. This is a hard job for the compiler, hence the flood of optimization PRs.

I'm currently shipping a wheel with the C-side changes from my open PRs on your repo, `cubie-numba-cuda-mlir`. `pip install cubie[mlir-cuda13]` in a fresh environment should drag it along for the repro. If you want to use a numba-cuda-mlir build it will need the branches from #255, #225, #298, and two from my fork:  ccam80/numba-cuda-mlir#6 and ccam80/numba-cuda-mlir#21. Of those two fork PRs, 21 is a draft of a launcher memory-leak fix that I'll submit after a bit more review, and 6 adds a hook for a typed whole-function planner that I use for an operation-scheduling pass. The scheduler hook seems domain-specific enough that I was going to keep it fork-only. Let me know if you've got any appetite for it and I can post it too, otherwise I'll just keep publishing a cubie-optimised wheel.

loop_shape.py runs on this branch.


https://gist.github.com/ccam80/0cc031844619a4c84823fa1ed732b4e3

- **cubie\_unroll.py**
  - adds a monkeypatch to convert cubie's unroll-hints patch to consteval
  - compiles a moderately complex cubie kernel under both settings
  - this 7s/18s compile time is mid-range for cubie. setting `linear_correction_type='bicgstab'` in the `Solver()` init args adds an inner nested early-exit loop which blows compile time to 20s/90s.
  
- **loop\_shape.py**
  - compiles a numba-cuda-mlir-only kernel based on cubie's loop shape under consteval and unroll hints

- **model\_fabbri\_linder.cellml**
  - model file used by cubie_unroll.py
  
 Numbers on my machine (RTX4070 Super, Windows):

cubie\_unroll.py:


|                     | consteval | cuda.unroll |
| ------------------- | --------- | ----------- |
| compile             | 18.2 s    | 7.7 s       |


loop\_shape.py (compile time), on this branch:


| N   | consteval | cuda.unroll |
| --- | --------- | ----------- |
| 8   | 0.2 s     | 0.1 s       |
| 16  | 0.3 s     | 0.1 s       |
| 32  | 0.7 s     | 0.1 s       |
| 64  | 1.6 s     | 0.3 s       |

ccam80:
CodeRabbit's point is true but I can't find  way to reach it. `np.zeros` inside a kernel is a stack allocation with no ref count. With NRT enabled, the refcounted values are runtime strings and extension types with an NRT model, iterating either would fail to compile or at launch.  I won't action it for now, let me know in review whether the it's worthwhile to extension-proof `cuda.unroll`.

PR #310, opened 2026-09-11
Fix ill typed insertvalue for alloca in LLVM7 Translator

brandon-b-miller:
Fixes an issue in the LLVM 7 translator where  `convertType` renders every opaque `!llvm.ptr` as `i8*`, but `translateAllocaOp` maps its result to a real `elemTy*. so any op writing a pointer into an i8* gets the wrong pointer type.

### DIFF
--- a/cext/mlir-llvm70/include/llvm70/LLVM70IRBuilder.h
+++ b/cext/mlir-llvm70/include/llvm70/LLVM70IRBuilder.h
@@ -74,6 +74,8 @@ public:
   // --- Basic blocks ---
   LLVMBasicBlockRef appendBB(LLVMValueRef fn, const char *name);
   void positionAtEnd(LLVMBasicBlockRef bb);
+  void positionBefore(LLVMValueRef inst);
+  LLVMValueRef getTerminator(LLVMBasicBlockRef bb);
   LLVMBasicBlockRef getInsertBlock();
 
   // --- Constants ---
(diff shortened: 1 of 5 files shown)


kkraus14 (commented):
Confirmed locally that this fixes the failure reported in #319 and the standalone reproducer in #321: both the minimal kernel and the original STUMPY kernel compile successfully with this change.

There is one remaining instance of the same typed-pointer mismatch in translatePhiOps. Direct br/cond_br incoming pointer values use lookupValueAsDeclared, but entries recorded in switchForwarders still use the raw lookupValue result. If an llvm.switch forwards an llvm.alloca result to a !llvm.ptr block argument, the PHI is declared as i8* while its incoming value remains (for example) i64*, producing invalid LLVM 7 IR.

The conversion needs to be emitted in each switch trampoline before its terminator so that it dominates the PHI, analogous to the existing predecessor handling. A focused regression should pass an alloca through llvm.switch to a pointer block argument and verify an i8* PHI with typed-pointer-to-i8* casts in the trampoline blocks.

kkraus14:
The switch/`cond_br` gap from my prior review is fixed, but two other pointer consumers still use raw `lookupValue` and can emit ill-typed LLVM 7 IR:

- `translateReturnOp`: returning an `llvm.alloca` from a function declared `-> !llvm.ptr` produces `define i8* ... { ret i64* ... }`.
- `translateICmpOp`: comparing allocas of different element types under `!llvm.ptr` produces `icmp eq i64* ..., i32* ...`.

Both reproduce with small MLIR functions on the current head. These operands should be reconciled via `lookupValueAsDeclared`; regressions for pointer return and pointer `icmp` would close the remaining gap.

brandon-b-miller:
> The switch/`cond_br` gap from my prior review is fixed, but two other pointer consumers still use raw `lookupValue` and can emit ill-typed LLVM 7 IR:
> 
> * `translateReturnOp`: returning an `llvm.alloca` from a function declared `-> !llvm.ptr` produces `define i8* ... { ret i64* ... }`.
> * `translateICmpOp`: comparing allocas of different element types under `!llvm.ptr` produces `icmp eq i64* ..., i32* ...`.
> 
> Both reproduce with small MLIR functions on the current head. These operands should be reconciled via `lookupValueAsDeclared`; regressions for pointer return and pointer `icmp` would close the remaining gap.

Fixed these and two more I found while auditing 

https://github.com/NVIDIA/numba-cuda-mlir/pull/310/changes/01e85f879887d943c8cea6f6c8e87b77f110a908

atmnp (approved):

PR #311, opened 2026-09-13
Fix Array.reshape() returning a wrong view when order doesn't match layout

M-A-Nagyd:
## Summary

`Array.reshape()` in `src/numba_cuda_mlir/numba_cuda/cudadrv/dummyarray.py` took the
no-copy fast path whenever the array was C- *or* F-contiguous, without checking that
the requested `order` argument actually matched which layout it was contiguous in.

Reshaping an F-contiguous array with `order="C"` (or a C-contiguous array with
`order="F"`) then produced a view with strides for the wrong traversal order over the
existing memory. This silently returns incorrect data instead of raising or copying,
since NumPy would copy in this situation and no such copy (or exception) ever happens
here.

`DeviceNDArray.reshape()` delegates directly to this function
(`devicearray.py:617`), so the bug affects device array reshapes as well.

### Reproduction

```python
import numpy as np
from numba_cuda_mlir.numba_cuda.cudadrv import dummyarray

base = np.arange(6, dtype=np.int64).reshape(2, 3)
x = np.asfortranarray(base)                     # F-contiguous, not C-contiguous
arr = dummyarray.Array.from_desc(0, x.shape, x.strides, x.itemsize)

newarr, extents = arr.reshape(6, order="C")      # reports a no-copy reshape
view = np.lib.stride_tricks.as_strided(x, shape=(6,), strides=newarr.strides)

print(view.tolist())                 # [0, 3, 1, 4, 2, 5]  <- wrong
print(x.reshape(6, order="C").tolist())  # [0, 1, 2, 3, 4, 5]  <- correct (NumPy copies)
```

## Fix

Restrict the fast path to the case where the requested order actually matches the
array's contiguity, and fall through to the existing order-aware
`attempt_nocopy_reshape()` algorithm otherwise, which correctly raises
`NotImplementedError` when a copy is genuinely required.

## Test plan

- Added `tests/numba_cuda_tests/cudadrv/test_dummyarray_reshape.py`, which builds
  real NumPy-backed memory, applies the strides `Array.reshape()` produces, and
  checks the result against NumPy's own order-aware reshape. Covers 1D/2D/3D
  reshapes in both mismatched-order (must copy/raise) and matching-order
  (must stay zero-copy) cases.
- Verified the new test fails on unpatched `main` (5/6 failures) and passes with
  this fix (6/6).
- `ruff check` and `ruff format --check` pass on both changed files.

Note: this environment has no NVIDIA GPU/CUDA toolkit, so I could not build the
native MLIR extension the package normally requires, or run the full existing
test suite (its `conftest.py` imports `cuda.core`). `dummyarray.py` has no
dependency on the native extension, so I verified the bug and fix by loading it
directly and running the exact assertions in the committed test file. CI should
have the final word on the full suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


M-A-Nagyd:
Good catch — the combined loop treated a raised `NotImplementedError` as acceptable for both the matching- and mismatched-order cases, so a regression that made the matching-order fast path wrongly reject a valid no-copy reshape would have passed silently.

Fixed in 069d303: split into `_check_matching_order_reshape` (must succeed with a correct no-copy view) and `_check_mismatched_order_reshape` (must raise `NotImplementedError`). Verified both directions still hold: the tightened tests pass against the fix and fail (5/5) against unpatched `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

atmnp (approved):
LGTM, thanks for your contribution!

PR #312, opened 2026-09-14
Separate dynamic shared memory from static allocations

tpn:
Dynamic shared arrays can overlap static shared arrays in the same kernel, corrupting either array. Their reported runtime size also becomes zero because the shared base lowers to an internal zero-length global.

This change uses an external shared declaration and reads the launch-time byte count from `%dynamic_smem_size`. Static shared arrays then remain separate, and runtime-sized allocations advance through aligned windows. The register is supported by both translation paths through inline PTX. [PTX register specification](https://docs.nvidia.com/cuda/parallel-thread-execution/#special-registers-dynamic-smem-size)

For example, `fixed = cuda.shared.array(16, np.int32)` can safely coexist with `dynamic = cuda.shared.array(0, np.int32)`. Launching with 128 dynamic bytes leaves `dynamic.size == 32` regardless of declaration order.

Zero-sized declarations view the remaining window without consuming it, preserving Numba-CUDA's repeated-view and mixed-dtype slicing idioms. Actual runtime-sized allocations advance the cursor; a later view starts at the aligned cursor and has zero elements if the window is exhausted. Explicit alignment assumptions retain their MLIR result. The memory guide describes these allocation and aliasing rules.

Validation:

- All 26 allocation tests passed with CUDA 12.9 and CUDA 13.3 on an RTX PRO 6000 (sm_120). The CUDA 13.3 run also passed Compute Sanitizer memcheck with zero errors.
- The documented float32/int32 mixed-view slicing example reproduces an invalid shared-memory write before the zero-view fix and passes afterward, including after a runtime-sized prefix.
- Six GPU-hidden PTX and optimized-IR tests passed across sm_90 and sm_120 under both CUDA versions, including explicit alignment. The original four PTX regressions were previously shown to fail against the old lowering.
- Changed-file pre-commit passed. Local validation used current Python source with native LLVM/MLIR libraries from the 0.5.1 wheel. The [full CI run](https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/34995899313) passed, including rebuilt wheels, Linux x86-64/ARM64 and Windows. One Windows lane passed on retry after an unrelated existing record-copy fixture produced an uninitialized NaN; all 26 allocation tests passed on both attempts.


src/numba_cuda_mlir/mlir_lowering.py
```diff
@@ -2392,33 +2426,45 @@ def _store_total_shared_memory_bytes(self, value: ir.Value):
...
         bytes = get_type_size_bytes(mr_type.element_type)
         assert self.mlir_funcOp
         # Emit at the current insertion point: the entry block may
         # already have a terminator once the request appears after
         # control flow. The shared-memory base itself is still created
         # at the entry block's start by _get_shared_memory_base.
         bytes_op = arith.constant(result=T.index(), value=bytes)
-        shm_base = self._get_shared_memory_base()
-        total_shared_memory_bytes = self._load_total_shared_memory_bytes()
+        alignment = max(alignment, bytes)
```
atmnp:
```
@cuda.jit
def kernel(n, out):
    shared = cuda.shared.array(n[0], dtype=cuda.float32x3)
    out[0] = shared.size


signature = types.void(types.int64[::1], types.int64[::1])
compiler.compile_ptx(kernel, signature, cc=(9, 0))
```
will fail here because the alignment won't be powers of 2 for the `x3` family of vector types.

atmnp (changes requested):

src/numba_cuda_mlir/mlir_lowering.py
tpn: Fixed in [550f884b](https://github.com/NVIDIA/numba-cuda-mlir/commit/550f884bc32913cadf681bca914a2b86c64ba823). Both runtime-sized allocations and zero-sized views now use power-of-two vector alignment and LLVM's padded element stride. `float32x3` therefore reserves 16 bytes per element; rounding only the alignment could still make adjacent allocations overlap. Explicit alignment changes the window start while preserving the element stride.

Added x3 regression coverage across 1-, 2-, 4-, and 8-byte components for the compile failure, remaining view sizes, adjacent allocations, and explicit alignment. All 58 shared-memory tests passed locally on CUDA 12.9 and 13.3, with zero Compute Sanitizer errors under CUDA 13.3. [Full CI](https://github.com/NVIDIA/numba-cuda-mlir/actions/runs/35645453303) also passed, including all 58 cases on rebuilt Windows and ARM64 wheels.

PR #315, opened 2026-09-14
[Py315] Basic Support for Python 3.15

atmnp:


brandon-b-miller:
https://github.com/NVIDIA/numba-cuda-mlir/pull/316 will be needed to avoid requiring python 3.15 `numba-cuda` wheels which are (I think erroneously) implicitly required in CI by our doc examples currently.

PR #323, opened 2026-09-18
Normalize negative array indices

### LINKED ISSUE #322: Negative array indices are not normalized in CUDA kernels
## Summary

Negative integer indices used to access NumPy arrays inside CUDA kernels are passed directly to MLIR `memref.load` instead of being normalized relative to the dimension size. A literal `-1` therefore reads before the requested row/array rather than reading its last element.

This was isolated while testing the STUMPY migration in #319, but the bug is independent of STUMPY.

## Minimal reproducer

```python
import numpy as np

@cuda.jit("(f8[:, :], f8[:])")
def read_last_column(matrix, out):
    i = cuda.grid(1)
    if i < matrix.shape[0]:
        out[i] = matrix[i, -1]


matrix = np.array([[10.0], [20.0], [30.0], [40.0]])
out = np.zeros(matrix.shape[0])
read_last_column[1, 32](matrix, out)
cuda.synchronize()

expected = matrix[:, -1]
print("expected:", expected)
print("actual:  ", out)
np.testing.assert_array_equal(out, expected)
```

## Actual result

```text
expected: [10. 20. 30. 40.]
actual:   [ 0. 10. 20. 30.]
```

The `-1` column index is used as a literal negative offset. With a one-column C-contiguous array, each row therefore reads the preceding row, and the first row reads before the allocation.

## Expected result

Negative indices should have NumPy/Numba semantics. In this example, `matrix[i, -1]` should read the last column and produce `[10, 20, 30, 40]`.

## Environment

- Current `main` plus the LLVM70 pointer fix from #310
- numba-cuda-mlir 0.5.2
- Python 3.14.7
- NumPy 2.4.6
- CUDA 13.4 / RTX 3500 Ada (compute capability 8.9)

## Analysis

`lower_array_tuple_getitem` in `src/numba_cuda_mlir/lowering/numpy.py` converts integer tuple indices directly to MLIR index constants/values and passes them to `array_element_value_load`. It does not normalize negative indices by adding the corresponding dimension extent. The one-dimensional `lower_array_getitem` and array setitem paths appear to follow the same raw-index pattern and should be audited as part of the fix.

This accounts for the largest post-#310 STUMPY failure cluster. Its GPU kernels use `profile[i, -1]` when maintaining nearest-neighbor profiles. The out-of-bounds read causes incorrect profile values and indices rather than a compiler error.

## STUMPY validation

After applying #310 plus its `llvm.switch` follow-up locally, the complete STUMPY test run reported:

```text
88 failed, 2399 passed, 16 skipped
```

All 88 failures were GPU numerical-correctness assertions. As a diagnostic workaround, I changed only the following two accesses, one in `gpu_stump.py` and one in `gpu_aamp.py`:

```diff
- if p_norm < profile[i, -1]:
+ if p_norm < profile[i, k - 1]:
```

Rerunning the complete previously-failing set then reported:

```text
88 passed
```

This indicates that the additional post-#310 STUMPY failures are one deduplicated negative-index issue rather than multiple algorithm-specific defects.

I searched the repository issue tracker for negative indexing, negative indices, multidimensional indexing, and incorrect array indexing and did not find an existing report for this behavior.

kkraus14:
## Summary

- normalize negative integer array indices against the corresponding dimension extent before lowering loads, stores, and rank-reducing subviews
- apply the normalization to scalar indexing and every integer component of tuple indexing
- add GPU coverage for literal and runtime negative indices across 1-D reads, 2-D reads, row views, and writes

Fixes #322.

## Validation

- standalone reproducer from #322 now returns the expected last column
- `tests/test_array.py`, `tests/test_array_view.py`, and `tests/test_numpy.py`: 181 passed, 7 skipped, 4 subtests passed
- all ten STUMPY modules that previously contained failures, with the diagnostic workaround removed: 188 passed
- pre-commit hooks: passed


kkraus14:
Addressed in `00d43fe`.

`test_getitem_2d` now includes a dedicated single-thread kernel covering both `arr[-1, 0]` and `arr[row, 0]` with `row=-2`. This exercises literal and runtime negative tuple-index normalization in dimension 0 through `lower_array_tuple_getitem`, independently of the rank-reducing scalar-index path.

Focused validation: `4 passed` for `TestNegativeArrayIndices`.

tests/test_array.py
```diff
@@ -372,6 +372,75 @@ def kernel(out):
...
+        out = np.zeros(2, dtype=np.int64)
+        kernel[1, 1](arr, out, -2)
+        np.testing.assert_array_equal(out, [8, 4])
+
+    def test_setitem_1d_and_2d(self):
+        @cuda.jit
+        def kernel(arr1d, arr2d, index, row, column):
+            arr1d[-1] = 40
+            arr1d[index] = 30
+            arr2d[-1, -1] = 120
```
atmnp, line 428: Can we also test  `arr2d[-1] = value` and verify the array?

kkraus14: Addressed in c8a1c7a. The test now executes `arr2d[-1] = 90` before the tuple-index writes and verifies the complete expected 2-D array, directly covering the rank-reducing scalar setitem path with a negative row index. The focused negative-index tests pass (4 passed).

atmnp (approved):

